PR #70, opened 2026-03-24
feat(vault): implement core vault contract with deposit/withdraw functionality

### LINKED ISSUE #16: Implement base Vault smart contract with deposit and withdraw logic
## Description

The Vault contract is the core primitive of Nester's savings layer. It accepts stablecoin deposits from users, tracks their balances, and allows withdrawals. Every other smart contract in the protocol interacts with or depends on the Vault contract—yield allocation, rebalancing, and offramp withdrawals all operate on vault state.

The current `packages/contracts/contracts/nester/src/lib.rs` contains a hello-world example. This issue replaces it with a production-oriented Vault contract that implements deposit, withdraw, and balance tracking using Soroban's storage and token interfaces.

At this early stage, the vault is a single-strategy vault—it holds deposits in a single token (e.g., USDC) and tracks per-user balances. Multi-strategy allocation (splitting across yield sources) comes in later issues. The focus here is correctness: ensuring that deposits, withdrawals, and balance accounting are mathematically precise and safe.

## Tasks

- [ ] Replace the hello-world contract with a Vault contract in `packages/contracts/contracts/nester/src/`
- [ ] Implement core storage:
  - Per-user balance tracking using Soroban persistent storage
  - Total vault balance (sum of all deposits)
  - Vault configuration (admin, accepted token, vault status)
- [ ] Implement public functions:
  - `initialize(admin, token_address)` — Set up the vault with an accepted token
  - `deposit(user, amount) -> Balance` — Accept token deposit, update user balance
  - `withdraw(user, amount) -> Balance` — Return tokens to user, update balance
  - `get_balance(user) -> i128` — Read user's vault balance
  - `get_total_deposits() -> i128` — Read total vault deposits
- [ ] Implement safety checks:
  - Prevent withdrawal exceeding balance
  - Prevent deposits when vault is paused
  - Validate token address matches configured token
  - Only admin can initialize and pause/unpause
- [ ] Emit Soroban events for deposits and withdrawals:
  - `(DEPOSIT, user, amount, new_balance)`
  - `(WITHDRAW, user, amount, new_balance)`
- [ ] Write unit tests in `test.rs` covering all functions and edge cases

## End Goal

A functional Vault contract deployed on Stellar testnet that accepts deposits, processes withdrawals, and tracks balances with mathematical precision. The contract is the foundation that all other protocol components build upon.

## Acceptance Criteria

- [ ] `deposit()` transfers tokens from user to vault and updates storage correctly
- [ ] `withdraw()` transfers tokens from vault to user and prevents over-withdrawal
- [ ] Balance accounting is exact: sum of all user balances equals total vault deposits at all times
- [ ] Admin-only functions (initialize, pause) reject non-admin callers
- [ ] Events are emitted for all state-changing operations
- [ ] Unit tests achieve >90% code coverage with edge cases (zero amounts, max values, unauthorized calls)

martinvibes:
- Add VaultStatus enum to track Active/Paused states
- Implement DataKey enum for persistent and instance storage management
- Add VaultError enum with 6 error types for validation and authorization
- Implement initialize() to set up admin, token address, and vault state
- Implement deposit() with balance tracking and event emission
- Implement withdraw() with balance validation and event emission
- Add get_balance() and get_total_deposits() query functions
- Implement pause() and unpause() admin functions for vault control
- Add get_status() and get_token() accessor functions
- Create comprehensive test suite with 283 lines covering all contract functions
- Add helper functions for authorization checks (require_admin, require_active, require_initialized)
- Add storage helper functions for balance and total deposit management

closes #16

### DIFF
--- a/packages/contracts/contracts/vault/src/test.rs
+++ b/packages/contracts/contracts/vault/src/test.rs
@@ -310,13 +310,13 @@ fn test_multiple_large_deposits_overflow_protection() {
     let user_a = Address::generate(&env);
     let user_b = Address::generate(&env);
 
-    let large_amount: i128 = i128::MAX / 2;
+    let large_amount: i128 = (i128::MAX / 2) + 1;
     mint_tokens(&env, &token_address, &user_a, large_amount);
     mint_tokens(&env, &token_address, &user_b, large_amount);
 
     client.deposit(&user_a, &large_amount);
 
-    // Second large deposit should panic due to overflow in total tracking
+    // Second deposit would make total exceed i128::MAX, causing overflow
     let result = std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
         client.deposit(&user_b, &large_amount);
     }));


0xDeon (dismissed):
## PR #70 - Review

**Score: 7.5/10**

### Summary
Implements the core vault contract on Soroban with deposit/withdraw/pause/unpause, real token transfers, per-user balance tracking, and events. This is a meaningful implementation over the prior stub. The logic is mostly sound but there is a critical Soroban-specific correctness issue and weak event test assertions.

---

### What's Good
- Proper guard chain: require_initialized -> require_active -> require_auth -> business logic on all mutating functions
- Correct storage tier usage: Persistent for per-user balances, Instance for contract-level state (admin, token, totals)
- Real token transfer via token::Client - not just internal accounting, actual asset movement
- VaultStatus enum instead of a bare bool - semantically cleaner
- is_paused() defaults to true when uninitialized (fail safe)
- Withdrawal correctly allowed during pause (standard safety pattern for yield protocols)
- Comprehensive 283-line test suite with real Stellar mock tokens, multi-user scenarios, and edge cases
- Tests cover zero/negative amounts, double-initialize prevention, per-user isolation, full withdrawal

---

### Issues Found

**1. CRITICAL: No TTL extension for Persistent storage**
Soroban Persistent storage entries expire after the ledger max TTL. The contract never calls env.storage().persistent().extend_ttl() on DataKey::Balance entries. This means user balances will silently expire from the ledger after ~2.5 months on mainnet, making deposited funds unrecoverable without a ledger state restore. This is not theoretical - it is how Soroban storage works. Fix: call extend_ttl on every read/write of persistent entries.

**2. Weak event test assertions**
Both test_deposit_emits_event and test_withdraw_emits_event only assert that the events list is non-empty. This does not verify the event topic, key, or payload. Assert the actual event structure using env.events().all() with topic matching.

**3. require_active missing from withdraw - undocumented intentional design**
Withdrawal during pause is the correct pattern for yield protocols (users must be able to exit). However, this deliberate design decision is not documented anywhere in the contract code. Add a comment on the withdraw function explicitly stating that withdrawals are permitted during pause.

**4. Old inline tests removed, new test.rs module added correctly**
The old tests that were removed called client.deposit() and client.withdraw() with no arguments - those signatures are now gone. The new tests properly use the updated signatures. Good cleanup.

---

### Verdict: REQUEST CHANGES

**Required Fixes:**
1. Add env.storage().persistent().extend_ttl() on every read/write of DataKey::Balance - this is non-negotiable for Soroban contracts
2. Strengthen event tests to assert topic and payload content, not just existence
3. Add a comment on withdraw documenting the intentional pause-allows-withdrawal design decision

packages/contracts/contracts/vault/src/lib.rs
```diff
@@ -78,43 +151,102 @@ impl VaultContract {
...
+        require_initialized(&env);
+        require_active(&env);
+
+        if amount <= 0 {
+            panic_with_error!(&env, ContractError::InvalidAmount);
+        }
+
+        user.require_auth();
+
+        let token_address: Address = env.storage().instance().get(&DataKey::Token).unwrap();
```
0xDeon: Calling `.unwrap()` on a storage read will cause a runtime panic if the `Token` key was somehow not set during initialization. Use `unwrap_or_else` with a proper `ContractError` variant instead, so failures produce a meaningful error code rather than an opaque contract trap.

PR #82, opened 2026-03-25
feat: Introduce intelligence API service with a Prometheus client, in…

### LINKED ISSUE #15: Implement Prometheus AI service relay endpoint
## Description

Nester's Prometheus AI layer runs as a separate Python/FastAPI service that analyzes vault performance, market sentiment, and user portfolios to generate advisory recommendations. The Go backend needs a relay layer that fetches recommendations from Prometheus and serves them to the frontend through the main API.

This relay pattern keeps the frontend simple—it calls one API (the Go backend) rather than juggling multiple service URLs. The Go backend handles authentication, caching, and error translation for the AI service. If Prometheus is down or slow, the backend returns graceful degradation responses rather than breaking the frontend.

At this stage, we implement the HTTP client, basic caching, and the API endpoints. The actual Prometheus service API contract will evolve, so the integration should be loosely coupled with clear interface boundaries.

## Tasks

- [ ] Create `internal/service/prometheus_client.go`:
  - HTTP client with configurable base URL, timeout, and retry logic
  - `GetVaultRecommendations(ctx, vaultID) ([]Recommendation, error)`
  - `GetMarketSentiment(ctx) (*SentimentReport, error)`
  - `GetPortfolioInsights(ctx, userID) (*PortfolioInsights, error)`
- [ ] Define response models in `internal/domain/intelligence/`:
  ```go
  type Recommendation struct {
      Type        string  // "rebalance", "yield_alert", "risk_warning"
      Title       string
      Description string
      Confidence  float64
      CreatedAt   time.Time
  }
  ```
- [ ] Add in-memory cache (TTL-based) for AI responses:
  - Market sentiment: cache for 5 minutes
  - Vault recommendations: cache for 2 minutes
  - Cache is per-key, not global
- [ ] Create handler endpoints:
  - `GET /api/v1/vaults/:id/recommendations` — Vault-specific recommendations
  - `GET /api/v1/intelligence/market` — Market sentiment summary
  - `GET /api/v1/users/:userId/insights` — Portfolio insights
- [ ] Implement circuit breaker pattern: if Prometheus fails 5 times in 30 seconds, stop calling for 60 seconds and return cached/empty responses
- [ ] Add `PrometheusConfig` to the config struct (base URL, API key, timeout)

## End Goal

The frontend can fetch AI-driven recommendations through the main Go API. The relay handles caching, retries, and circuit breaking so that Prometheus failures don't cascade to the user experience. The integration is loosely coupled—swapping or upgrading the AI service requires only changes to the client layer.

## Acceptance Criteria

- [ ] Recommendation endpoints return data when Prometheus is available
- [ ] Endpoints return empty/cached results (not errors) when Prometheus is unavailable
- [ ] Cached responses are served within 1ms (no upstream call)
- [ ] Circuit breaker opens after consecutive failures and auto-recovers
- [ ] HTTP client has configurable timeout (default 5s) and does not block indefinitely
- [ ] Unit tests mock the Prometheus HTTP client and verify caching, circuit breaking, and error handling

githoboman:
i have implemented the Prometheus AI Relay Layer in the Nester Go backend. This layer acts as a proxy for the Prometheus AI service, providing caching, retries, and circuit breaking to ensure a reliable and performant user experience.
Closes #15 

Changes Made
Intelligence Domain
Created 
models.go
 to define the AI-driven data structures:
Recommendation
: For vault-specific advisory suggestions.
SentimentReport
: For aggregate market sentiment.
PortfolioInsights
: For user-specific portfolio analysis.
Configuration
Modified 
config.go
 to include 
PrometheusConfig
, supporting:
PROMETHEUS_BASE_URL (default: http://localhost:8000)
PROMETHEUS_API_KEY
PROMETHEUS_TIMEOUT (default: 5s)
Prometheus Service Client
Implemented 
prometheus_client.go
 with:
In-memory Caching: TTL-based (2m for vaults, 5m for market/users) to minimize upstream calls.
Circuit Breaker: Trips after 5 failures in 30s, holding for 60s to prevent cascading failures.
Automatic Retries: Up to 3 attempts with exponential backoff for transient errors.
Graceful Degradation: Returns empty/cached data when the upstream service is unavailable or the circuit is open.
API Handlers and Routing
Created 
intelligence.go
 with handlers for the new endpoints.
Updated 
router.go
 and 
main.go
 to wire up the new routes:
GET /api/v1/vaults/:id/recommendations
GET /api/v1/intelligence/market
GET /api/v1/users/:userId/insights
Verification Results
Automated Tests
Service Tests: 
prometheus_client_test.go
TestPrometheusClient_Caching
: Verified that consecutive calls within the TTL return cached data without hitting the server.
TestPrometheusClient_CircuitBreaker
: Confirmed the circuit breaker opens after 5 failures and prevents further calls for the cooldown period.
TestPrometheusClient_Retry
: Validated that the client successfully retries on transient errors.
Handler Tests: 
intelligence_test.go
Verified that API endpoints return the expected JSON structure and status codes.
Test Execution
bash
# internal/service tests
ok  	github.com/Suncrest-Labs/nester/internal/service	3.147s
# internal/handler tests
ok  	github.com/Suncrest-Labs/nester/internal/handler	1.089s
Summary
The implementation follows the requested relay pattern, ensuring that the Go backend remains the single point of entry for the frontend while providing a resilient integration with the Prometheus AI service.


0xDeon (changes requested):
## PR #82 - Review

**Score: 6/10**

### Summary
Implements a Go backend relay layer for the Prometheus AI service with in-memory caching, a basic circuit breaker, and retry logic. The architecture is sound and the patterns are real production patterns. However, there is a critical resource leak bug that must be fixed before this merges.

---

### What's Good
- Circuit breaker uses a failure window (30s) to track recent failures, not just a raw count - this is the correct approach
- sync.RWMutex for cache reads/writes, sync.Mutex for circuit breaker - correct synchronization granularity
- TTL-based cache with separate TTLs per endpoint type (VaultCacheTTL=2m, others=5m)
- http.NewRequestWithContext correctly propagates cancellation
- Authorization header conditionally set only when API key is non-empty
- Graceful degradation returns empty structs instead of errors when circuit is open
- Test coverage for caching, circuit breaker, and retry behaviors with atomic counters

---

### Issues Found

**1. CRITICAL: HTTP response body leak in doRequest**
In the retry loop, resp.Body.Close() is only called in the success branch (Status 200). On non-200 responses, the body is never closed:

for i := 0; i < 3; i++ {
    resp, err = c.httpClient.Do(req)
    if err == nil && resp.StatusCode == http.StatusOK {
        defer resp.Body.Close()  // only closed on success
        return json.NewDecoder(resp.Body).Decode(target)
    }
    // resp.Body is never closed here on 4xx/5xx
}

This exhausts the connection pool under any load where the upstream returns non-200 responses. Close the body in the non-success path: if resp != nil { resp.Body.Close() } before continuing to the next retry.

**2. Request reused across retries**
The http.Request created before the loop is passed to httpClient.Do() multiple times. A Request with a non-nil Body cannot be safely reused after the first Do() call. Even for GET requests (bodyless), this is a bad pattern that will cause subtle issues if the contract ever changes. Create the request inside the retry loop.

**3. IntelligenceHandler depends on concrete type, not interface**
IntelligenceHandler holds a *service.PrometheusClient concrete type. This couples the handler to the implementation and makes unit testing impossible without a real HTTP server (as seen in the tests). Define a PrometheusService interface in the handler package and inject against that.

**4. Unsafe type assertion from cache**
val.([]intelligence.Recommendation) will panic if a cache collision ever occurs or if the wrong type gets stored. Use the comma-ok form: val, ok := c.getFromCache(key); if ok { if recs, ok := val.([]intelligence.Recommendation); ok { return recs, nil } }.

**5. Typo in test: Rebalance Suggesion**
intelligence_test.go line: Title: Rebalance Suggesion - missing t. Fix the typo.

**6. GetPortfolioInsights handler has no test coverage**
The test file covers GetVaultRecommendations and GetMarketSentiment but not GetPortfolioInsights. Add coverage for the third endpoint.

---

### Verdict: REQUEST CHANGES

**Required Fixes:**
1. Fix the response body leak: close resp.Body in all non-success retry iterations
2. Move http.NewRequestWithContext inside the retry loop
3. Define a PrometheusService interface and inject it into IntelligenceHandler
4. Use comma-ok type assertion when reading from cache
5. Fix the Suggesion typo in the test
6. Add test for GetPortfolioInsights handler

0xDeon (changes requested):

PR #270, opened 2026-04-22
fix(auth): populate Roles in JWT claims to restore admin RBAC (#241)

### LINKED ISSUE #241: [Backend] JWT issuance never populates the Roles field — admin role-based access control is permanently broken
## Summary
The Go API defines a role-based access control (RBAC) system where admin routes are protected by middleware that checks for an `admin` role in the JWT claims. However, the JWT issuance code in `apps/api/internal/service/auth_service.go` creates tokens with only `Subject`, `WalletAddress`, `IssuedAt`, and `ExpiresAt` fields — the `Roles` field is never populated. As a result, every JWT ever issued by this system has an empty `Roles` slice. The admin middleware checks `!contains(claims.Roles, "admin")` which is always `true` (the empty slice never contains "admin"), causing every request to admin endpoints to return HTTP 403 Forbidden — including requests from the actual system administrator. The admin panel is permanently inaccessible to everyone, including the team operating the protocol.

## Evidence

In `apps/api/internal/service/auth_service.go`:
```go
func (s *AuthService) IssueToken(ctx context.Context, walletAddress string) (string, error) {
    claims := auth.Claims{
        Subject:       user.ID.String(),
        WalletAddress: walletAddress,
        IssuedAt:      time.Now().Unix(),
        ExpiresAt:     time.Now().Add(s.tokenTTL).Unix(),
        // ❌ Roles: []string{} — never set, even for admin users
        // ❌ Scopes: []string{} — never set
    }
    return auth.GenerateToken(claims, s.jwtSecret)
}
```

In `apps/api/internal/middleware/auth.go`:
```go
// Admin check — always fails because claims.Roles is always empty
if rule.Role == "admin" && !contains(claims.Roles, "admin") {
    http.Error(w, "forbidden", http.StatusForbidden) // Always returns this
    return
}
```

There is no code path anywhere in the codebase that sets `claims.Roles` to include `"admin"` or any other role before token issuance.

## Cascading Impact
1. **All admin endpoints are permanently locked** — `POST /api/v1/admin/vaults/:id/pause`, `POST /api/v1/admin/vaults/:id/unpause`, user management, fee configuration, and any other admin route return 403 for everyone without exception
2. **Emergency response is impossible** — if the vault needs to be paused during an active exploit, the admin API cannot be used. This removes the primary off-chain emergency control mechanism
3. **False security posture** — the RBAC system gives the appearance that admin routes are protected, but they are actually protected by accident (they block everyone including admins, not by design)
4. **No role assignment mechanism exists** — even if the code were fixed to populate Roles from a database, there is currently no `user_roles` table, no role assignment endpoint, and no CLI tool to bootstrap the first admin user

## Root Cause
The `Claims` struct was designed with `Roles []string` but the implementation of `IssueToken` was never updated to query user roles from the database and include them. This is likely because the `user_roles` table and role management system were planned but not yet implemented.

### Step 1 — Add roles to the users table (or a separate roles table)
```sql
-- Migration 009_add_user_roles.sql
CREATE TABLE user_roles (
    user_id UUID NOT NULL REFERENCES users(id) ON DELETE CASCADE,
    role    VARCHAR(50) NOT NULL,
    granted_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
    granted_by UUID REFERENCES users(id),
    PRIMARY KEY (user_id, role)
);

-- Seed the first admin (replace with actual admin wallet address)
INSERT INTO user_roles (user_id, role)
SELECT id, 'admin' FROM users WHERE wallet_address = '${ADMIN_WALLET_ADDRESS}';
```

### Step 2 — Query roles at token issuance time
```go
func (s *AuthService) IssueToken(ctx context.Context, walletAddress string) (string, error) {
    user, err := s.userRepo.GetByWalletAddress(ctx, walletAddress)
    if err != nil { return "", err }
    
    roles, err := s.userRepo.GetUserRoles(ctx, user.ID)
    if err != nil { return "", err }
    
    claims := auth.Claims{
        Subject:       user.ID.String(),
        WalletAddress: walletAddress,
        IssuedAt:      time.Now().Unix(),
        ExpiresAt:     time.Now().Add(s.tokenTTL).Unix(),
        Roles:         roles, // e.g., ["admin"] or []
    }
    return auth.GenerateToken(claims, s.jwtSecret)
}
```

### Step 3 — Create a bootstrap CLI command for first admin
```go
// cmd/bootstrap-admin/main.go
// A one-time CLI tool to grant admin role to a wallet address
// Run: go run ./cmd/bootstrap-admin --wallet=G...
```

### Step 4 — Write integration tests for the admin middleware
```go
func TestAdminMiddleware_BlocksNonAdmin(t *testing.T) { ... }
func TestAdminMiddleware_AllowsAdmin(t *testing.T) { ... }
func TestAdminMiddleware_BlocksExpiredToken(t *testing.T) { ... }
```

## References
- JWT RFC 7519 — Claims structure
- OWASP API Security — API5: Broken Function Level Authorization
- [Go JWT best practices](https://pkg.go.dev/github.com/golang-jwt/jwt)

Miracle656:
Closes #241 

  Problem

  Every JWT ever issued by this API had an empty Roles slice. The IssueToken path built auth.Claims with Subject,
  WalletAddress, IssuedAt, and ExpiresAt — but never queried or set Roles. The admin middleware's check
  !user.HasRole("admin") was therefore always true, causing every request to /api/v1/admin/… to return 403 Forbidden,
  including requests from actual administrators. The admin panel has been permanently inaccessible since the middleware
  was introduced.

  Root Cause

  Claims.Roles was defined in the struct but AuthService.VerifyAndIssue had no code to fetch roles from the database and
   populate the field. There was also no user_roles table for roles to live in.

  Changes

  Migration (009_add_user_roles)

  - Creates user_roles(user_id, role, granted_at, granted_by) with a composite primary key and a cascade-delete foreign
  key to users
  - Includes a down migration (DROP TABLE user_roles)

  Repository & Service

  - Added GetRoles(ctx, userID) ([]string, error) to the UserRepository interface
  - Implemented it in postgres.UserRepository with a SELECT role FROM user_roles WHERE user_id = $1 query
  - Added GetUserRoles to UserService as a thin wrapper

  Auth fix

  - AuthService.VerifyAndIssue now calls GetUserRoles after resolving the user and populates claims.Roles before signing
   the token

  Bootstrap CLI (cmd/bootstrap-admin)

  - One-shot tool to grant the admin role to a wallet address:
  go run ./cmd/bootstrap-admin --wallet=G... [--dsn=postgres://...]
  - Required to seed the first administrator on any existing or fresh deployment

  Tests

  - TestAuthService_VerifyAndIssue_AdminRolePopulatedInToken — verifies the issued JWT contains ["admin"] when the user
  has the admin role in the repository
  - TestAuthService_VerifyAndIssue_RegularUserHasEmptyRoles — verifies a newly registered user's token has no roles
  - Updated all UserRepository mocks in service and handler test packages to satisfy the extended interface

  Testing

  # Unit tests (no database required)
  go test ./internal/auth/... ./internal/middleware/... ./internal/service/... ./internal/handler/...

  # Bootstrap admin on a running instance
  go run ./cmd/bootstrap-admin --wallet=<ADMIN_STELLAR_ADDRESS>

  Migration path for existing deployments

  # Apply migration 009
  # (the server runs MigrateUp automatically on startup)

  # Grant admin role to the first operator
  go run ./cmd/bootstrap-admin --wallet=G<YOUR_ADMIN_WALLET>

  All previously issued tokens will remain valid but will not carry roles. Users must re-authenticate after the
  migration to receive a token with their roles.

### DIFF
--- a/apps/api/cmd/bootstrap-admin/main.go
+++ b/apps/api/cmd/bootstrap-admin/main.go
@@ -1,0 +1,71 @@
+// bootstrap-admin grants the admin role to a user identified by wallet address.
+// Run once to seed the first administrator after deploying a fresh database.
+//
+// Usage:
+//
+//	go run ./cmd/bootstrap-admin --wallet=G... [--dsn=postgres://...]
+//
+// The DSN defaults to the DATABASE_DSN environment variable.
+package main
+
+import (
+	"database/sql"
+	"flag"
+	"fmt"
+	"os"
+
+	_ "github.com/jackc/pgx/v5/stdlib"
+	"github.com/joho/godotenv"
+)
+
+func main() {
+	if err := run(); err != nil {
+		fmt.Fprintln(os.Stderr, "bootstrap-admin:", err)
+		os.Exit(1)
+	}
+}
+
+func run() error {
+	wallet := flag.String("wallet", "", "Stellar wallet address to grant admin role (required)")
+	dsn := flag.String("dsn", "", "PostgreSQL DSN (defaults to DATABASE_DSN env var)")
+	flag.Parse()
+
+	if *wallet == "" {
+		flag.Usage()
+		return fmt.Errorf("--wallet is required")
+	}
+
+	_ = godotenv.Load()
+	if *dsn == "" {
+		*dsn = os.Getenv("DATABASE_DSN")
+	}
+	if *dsn == "" {
+		return fmt.Errorf("DATABASE_DSN environment variable or --dsn flag is required")
+	}
+
+	db, err := sql.Open("pgx", *dsn)
+	if err != nil {
+		return fmt.Errorf("open db: %w", err)
+	}
+	defer db.Close()
+
+	var userID string
+	err = db.QueryRow(`SELECT id FROM users WHERE wallet_address = $1`, *wallet).Scan(&userID)
+	if err == sql.ErrNoRows {
+		return fmt.Errorf("no user found with wallet address %q", *wallet)
+	}
+	if err != nil {
+		return fmt.Errorf("query user: %w", err)
+	}
+
+	_, err = db.Exec(
+		`INSERT INTO user_roles (user_id, role) VALUES ($1, 'admin') ON CONFLICT DO NOTHING`,
+		userID,
+	)
+	if err != nil {
+		return fmt.Errorf("insert role: %w", err)
+	}
+
+	fmt.Printf("granted admin role to user %s (wallet %s)\n", userID, *wallet)
+	return nil
+}
--- a/apps/api/internal/domain/user/model.go
+++ b/apps/api/internal/domain/user/model.go
@@ -35,4 +35,5 @@ type UserRepository interface {
 	Create(ctx context.Context, user *User) error
 	GetByID(ctx context.Context, id uuid.UUID) (*User, error)
 	GetByWalletAddress(ctx context.Context, addr string) (*User, error)
+	GetRoles(ctx context.Context, id uuid.UUID) ([]string, error)
 }
--- a/apps/api/internal/handler/user_handler_test.go
+++ b/apps/api/internal/handler/user_handler_test.go
@@ -51,6 +51,10 @@ func (m *mockUserRepository) GetByWalletAddress(ctx context.Context, addr string
 	return nil, user.ErrUserNotFound
 }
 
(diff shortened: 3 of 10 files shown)


0xDeon (approved):
Lgtm, good work

PR #271, opened 2026-04-23
fix(settlements): enforce ownership check on PATCH status endpoint (#239)

### LINKED ISSUE #239: [Backend] Settlement status update has no ownership check — any authenticated user can modify any other user's settlement
## Summary
`PATCH /api/v1/settlements/{id}/status` correctly requires a valid JWT (the caller must be authenticated) but entirely fails to verify that the authenticated user owns the settlement being modified. The handler fetches the settlement by its UUID and updates its status without checking whether the `user_id` on the settlement record matches the authenticated caller's identity from the JWT. This is a Broken Object Level Authorization (BOLA/IDOR) vulnerability — the most common and highest-impact category in the OWASP API Security Top 10. Any authenticated wallet holder can update the status of any other user's fiat off-ramp settlement to any value they choose.

## Evidence

In `apps/api/internal/handler/settlement_handler.go`:
```go
func (h *SettlementHandler) UpdateStatus(w http.ResponseWriter, r *http.Request) {
    settlementID := chi.URLParam(r, "id")
    // JWT is validated by middleware — but the authenticated user ID is never extracted here
    
    var req UpdateStatusRequest
    json.NewDecoder(r.Body).Decode(&req)
    
    // Calls service with only the settlement ID — no ownership check
    err := h.service.UpdateStatus(r.Context(), settlementID, req.Status)
    // ...
}
```

In `apps/api/internal/service/settlement_service.go`:
```go
func (s *SettlementService) UpdateStatus(ctx context.Context, settlementID string, status string) error {
    // Fetches by ID only — no WHERE user_id = $authenticated_user clause
    return s.repo.UpdateStatus(ctx, settlementID, status)
}
```

The authenticated user's ID (available from the JWT claims in the request context) is never read in either the handler or service layer during a status update operation.

### Scenario A — Mark settlement as completed to release funds prematurely
A malicious user discovers or guesses another user's settlement UUID (UUIDs are not secret if they appear in API responses, URLs, or logs). The attacker sends:
```
PATCH /api/v1/settlements/victim-settlement-uuid/status
{"status": "completed"}
```
The settlement is marked as complete in the database. If any downstream process (liquidity node payout, balance reconciliation) triggers on "completed" status, the victim's funds may be released to the wrong party or the legitimate release may be blocked.

### Scenario B — Cancel another user's settlement
```
PATCH /api/v1/settlements/victim-settlement-uuid/status
{"status": "cancelled"}
```
The victim's pending off-ramp request is cancelled. They lose their place in the queue and their funds may be locked until they re-initiate the process. For users in emerging markets who urgently need fiat liquidity, this can have real-world financial consequences.

### Scenario C — Status manipulation for double-spend
An attacker who also has a legitimate settlement:
1. Initiates their own settlement → attacker settlement is "processing"
2. Finds another user's settlement UUID that is also "processing"
3. Marks the victim's settlement as "completed" — this satisfies the settlement provider's success condition for the attacker's settlement
4. The attacker receives fiat payout while the victim's settlement is stuck

## Impact
1. **Direct financial harm to users** — settlement cancellation or premature completion can result in users not receiving their fiat off-ramp funds
2. **Settlement integrity** — the entire off-ramp flow relies on settlement status being a trusted record of the actual fiat transfer state; unauthorized modifications break this trust
3. **Regulatory exposure** — incorrect settlement records in a money-transfer context create compliance and audit trail issues
4. **Scalability of attack** — if settlement UUIDs are exposed in any API response (e.g., a list endpoint), an attacker can enumerate and attack all active settlements systematically

## Root Cause
The authentication middleware validates the JWT and may store claims in the request context, but the settlement handler never reads those claims. The ownership check was likely intended but omitted during development.

### Step 1 — Extract authenticated user ID in the handler
```go
func (h *SettlementHandler) UpdateStatus(w http.ResponseWriter, r *http.Request) {
    settlementID := chi.URLParam(r, "id")
    
    // Extract the authenticated user's ID from JWT claims in context
    claims, ok := r.Context().Value(middleware.ClaimsKey).(*auth.Claims)
    if !ok {
        http.Error(w, "unauthorized", http.StatusUnauthorized)
        return
    }
    authenticatedUserID, err := uuid.Parse(claims.Subject)
    if err != nil {
        http.Error(w, "invalid token", http.StatusUnauthorized)
        return
    }
    
    var req UpdateStatusRequest
    json.NewDecoder(r.Body).Decode(&req)
    
    err = h.service.UpdateStatus(r.Context(), settlementID, authenticatedUserID, req.Status)
    if errors.Is(err, service.ErrForbidden) {
        http.Error(w, "forbidden", http.StatusForbidden)
        return
    }
    // ...
}
```

### Step 2 — Enforce ownership in the service layer
```go
func (s *SettlementService) UpdateStatus(
    ctx context.Context,
    settlementID string,
    authenticatedUserID uuid.UUID,
    newStatus string,
) error {
    settlement, err := s.repo.GetByID(ctx, settlementID)
    if err != nil {
        return err
    }
    if settlement.UserID != authenticatedUserID {
        return ErrForbidden // 403, not 404 — don't leak existence
    }
    return s.repo.UpdateStatus(ctx, settlementID, newStatus)
}
```

### Step 3 — Apply the same pattern to all settlement mutations
Audit every settlement endpoint (create, update, delete, status change) and ensure the same ownership verification is applied consistently.

### Step 4 — Consider restricting status transitions
Not all status values should be settable by users. Consider:
- Users can only cancel their own "pending" settlements
- Status transitions to "processing" and "completed" should only come from the liquidity node service or admin, not from users

## References
- [OWASP API Security Top 10 — API1: Broken Object Level Authorization (BOLA)](https://owasp.org/API-Security/editions/2023/en/0xa1-broken-object-level-authorization/)
- IDOR (Insecure Direct Object Reference) — CWE-639

Miracle656:
Closes #239 

Problem

  PATCH /api/v1/settlements/{id}/status verified the caller's JWT but never checked whether the authenticated user owned
  the settlement being modified. Any authenticated wallet holder could update any other user's settlement status to any
  value — a BOLA/IDOR vulnerability (OWASP API Security Top 10, API1).

  Concretely: attacker authenticates → discovers or guesses a victim's settlement UUID → sends PATCH
  /api/v1/settlements/<victim-uuid>/status {"status":"cancelled"} → victim's off-ramp is cancelled. The UUID is often
  exposed in API list responses, making enumeration trivial.

  Root Cause

  UpdateStatus in the service fetched the settlement by ID and immediately applied the transition. The authenticated
  caller's identity was stored in the request context by the auth middleware but was never read in either the handler or
  service layer.

  Changes

  Domain (offramp/model.go)

  - Added ErrForbidden = errors.New("caller does not own this settlement")

  Service (settlement_service.go)

  - Added CallerID uuid.UUID to UpdateStatusInput
  - After fetching the settlement, checks settlement.UserID != input.CallerID and returns ErrForbidden before any
  transition logic runs

  Handler (settlement_handler.go)

  - Calls auth.GetUserFromContext to extract the authenticated user from context; returns 401 if absent (defense-in-depth
  — the auth middleware normally guarantees this is populated, but the check guards against misconfiguration)
  - Parses caller.ID as a UUID and passes it in UpdateStatusInput.CallerID
  - writeDomainError now maps ErrForbidden → HTTP 403

  Tests

  Service layer:
  - TestUpdateStatus_ForbiddenForNonOwner — a different user's ID as CallerID returns ErrForbidden
  - All existing UpdateStatus call sites updated with CallerID: s.UserID

  Handler layer:
  - injectAuthUser test helper — wraps a mux with middleware that plants an auth.User in context, simulating production
  auth middleware
  - TestSettlementHandler_PatchStatus200 — updated to inject the settlement owner's identity; still passes
  - TestSettlementHandler_PatchStatus403NonOwner — a different authenticated user gets 403
  - TestSettlementHandler_PatchStatus401NoAuth — no auth context at all gets 401

  Testing

  go test ./internal/domain/offramp/... ./internal/service/... ./internal/handler/...

### DIFF
--- a/apps/api/internal/domain/offramp/model.go
+++ b/apps/api/internal/domain/offramp/model.go
@@ -39,6 +39,7 @@ var (
 	ErrInvalidStatus         = errors.New("invalid settlement status")
 	ErrInvalidTransition     = errors.New("invalid status transition")
 	ErrInvalidPrecision      = errors.New("decimal precision exceeds supported scale")
+	ErrForbidden             = errors.New("caller does not own this settlement")
 )
 
 const MaxAmountScale = int32(8)
--- a/apps/api/internal/handler/settlement_handler.go
+++ b/apps/api/internal/handler/settlement_handler.go
@@ -7,6 +7,7 @@ import (
 	"github.com/google/uuid"
 	"github.com/shopspring/decimal"
 
+	"github.com/suncrestlabs/nester/apps/api/internal/auth"
 	"github.com/suncrestlabs/nester/apps/api/internal/domain/offramp"
 	"github.com/suncrestlabs/nester/apps/api/internal/service"
 	logpkg "github.com/suncrestlabs/nester/apps/api/pkg/logger"
@@ -169,6 +170,17 @@ func (h *SettlementHandler) updateStatus(w http.ResponseWriter, r *http.Request)
 		return
 	}
 
+	caller, ok := auth.GetUserFromContext(r.Context())
+	if !ok {
+		response.WriteJSON(w, http.StatusUnauthorized, response.Err(http.StatusUnauthorized, "UNAUTHORIZED", "unauthorized"))
+		return
+	}
+	callerID, err := uuid.Parse(caller.ID)
+	if err != nil {
+		response.WriteJSON(w, http.StatusUnauthorized, response.Err(http.StatusUnauthorized, "UNAUTHORIZED", "invalid caller identity"))
+		return
+	}
+
 	var req updateStatusRequest
 	if err := decodeJSON(r, &req); err != nil {
 		response.WriteJSON(w, http.StatusBadRequest, response.ValidationErr(err.Error()))
@@ -177,6 +189,7 @@ func (h *SettlementHandler) updateStatus(w http.ResponseWriter, r *http.Request)
 
 	model, err := h.service.UpdateStatus(r.Context(), service.UpdateStatusInput{
 		SettlementID: id,
+		CallerID:     callerID,
 		NewStatus:    offramp.SettlementStatus(req.Status),
 	})
 	if err != nil {
@@ -191,6 +204,8 @@ func (h *SettlementHandler) updateStatus(w http.ResponseWriter, r *http.Request)
 
 func (h *SettlementHandler) writeDomainError(w http.ResponseWriter, r *http.Request, err error) {
 	switch {
+	case errors.Is(err, offramp.ErrForbidden):
+		response.WriteJSON(w, http.StatusForbidden, response.Err(http.StatusForbidden, "FORBIDDEN", "you do not own this settlement"))
 	case errors.Is(err, offramp.ErrSettlementNotFound):
 		response.WriteJSON(w, http.StatusNotFound, response.NotFound("settlement"))
 	case errors.Is(err, offramp.ErrUserNotFound):
--- a/apps/api/internal/handler/settlement_handler_test.go
+++ b/apps/api/internal/handler/settlement_handler_test.go
@@ -13,10 +13,19 @@ import (
 	"github.com/google/uuid"
 	"github.com/shopspring/decimal"
 
+	"github.com/suncrestlabs/nester/apps/api/internal/auth"
 	"github.com/suncrestlabs/nester/apps/api/internal/domain/offramp"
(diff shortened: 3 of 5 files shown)


0xDeon (approved):
Lgtm

PR #275, opened 2026-04-23
fix: Performance fee calculation compares assets to shares — users charged fees on principal, not yield

### LINKED ISSUE #236: [Vault] Performance fee calculation compares assets to shares — users charged fees on principal, not yield
## Summary
In the vault contract's `withdraw()` function, the performance fee is calculated by subtracting the share count from the asset amount received:

```rust
let yield_part = assets_to_withdraw.saturating_sub(shares);
```

This is a unit mismatch error. `assets_to_withdraw` is denominated in the underlying token (e.g., USDC). `shares` is denominated in vault share tokens. These are the same number only when the exchange rate is exactly 1:1. After any yield accrual — which is the entire point of the vault — the exchange rate drifts above 1.0, and this formula begins charging performance fees on the exchange rate appreciation that was baked in before the user deposited. A user who deposits after yield has already accrued will be charged a performance fee on appreciation they never benefited from.

### Scenario A — User charged fee on other people's yield
1. Alice deposits 1,000 USDC at exchange rate 1.0 → receives 1,000 shares
2. The vault accrues yield → exchange rate moves to 1.1
3. Bob deposits 1,000 USDC at rate 1.1 → receives ~909 shares
4. Bob immediately withdraws his 909 shares → receives ~1,000 USDC (he has zero real yield)
5. Current formula: `yield_part = 1000 - 909 = 91`
6. Bob is charged a performance fee on 91 USDC of "yield" he never earned — this was pure exchange rate effect from Alice's deposit period

### Scenario B — Correct user pays excessive fee
1. Alice deposits 1,000 USDC at rate 1.0 → 1,000 shares
2. Rate moves to 2.0 (Alice has earned 1,000 USDC yield)
3. Alice withdraws 1,000 shares → receives 2,000 USDC
4. Current formula: `yield_part = 2000 - 1000 = 1000` ✓ (this case is accidentally correct)
5. But Alice's actual yield is `2000 - 1000 = 1000` USDC, so the formula works here only because Alice deposited at rate 1.0

The formula is only correct for users who deposited at an exchange rate of exactly 1.0. For every user who deposited after the rate drifted, they are overcharged.

## Impact
1. **Incorrect fee collection** — the protocol collects more fees than entitled, directly reducing user returns below the advertised rate
2. **Violation of stated fee schedule** — if the product spec says "X% performance fee on yield earned," the contract charges fees on a different amount, constituting a misrepresentation
3. **Compounding effect** — in a high-TVL vault where many users deposit at varying exchange rates, the aggregate overcharge can be significant
4. **User trust** — emerging markets users with thin margins will notice unexplained fee deductions and lose trust in the protocol

## Root Cause
The formula was likely written during initial development when the exchange rate was assumed to start at 1.0 and the fact that `assets - shares` equals yield only holds at rate = 1.0 was not caught in review.

### Track deposit cost basis per user
Store the exchange rate at the time of deposit (or the USDC equivalent of shares issued) in contract storage:

```rust
// On deposit
let exchange_rate = calculate_exchange_rate(&env); // total_assets / total_supply
let cost_basis_per_share = exchange_rate; // price paid per share at deposit time
env.storage().persistent().set(
    &(Symbol::new(&env, "basis"), from.clone()),
    &cost_basis_per_share
);

// On withdrawal — calculate actual yield
let cost_basis: i128 = env.storage().persistent()
    .get(&(Symbol::new(&env, "basis"), from.clone()))
    .unwrap_or(FIXED_POINT_SCALE); // default to 1.0 if no basis recorded

let cost_basis_value = shares
    .checked_mul(cost_basis)
    .unwrap_or(i128::MAX) / FIXED_POINT_SCALE;

let actual_yield = assets_to_withdraw.saturating_sub(cost_basis_value);
let yield_fee = if actual_yield > 0 {
    actual_yield
        .checked_mul(performance_fee_bps as i128)
        .unwrap_or(0) / 10_000
} else {
    0 // Never charge a fee on a loss
};
```

### Handle partial withdrawals
For users who make multiple deposits at different exchange rates, use a weighted average cost basis or FIFO lot accounting.

### Write regression tests covering:
- User deposits at rate > 1.0, withdraws immediately → zero performance fee
- User deposits at rate 1.0, rate doubles, withdraws → performance fee on exact yield earned
- User deposits at rate 1.0, rate halves (impairment scenario), withdraws → zero fee (never charge on loss)

## References
- EIP-4626 Tokenized Vault Standard — yield accounting discussion
- Yearn Finance vault share accounting: `pricePerShare` recorded at deposit time
- [Uniswap v3 position cost basis tracking](https://github.com/Uniswap/v3-core)

wheval:
## Summary
- #236: Performance fee calculation compares assets to shares — users charged fees on principal, not yield

## Implementation notes
- Issue #236: validate root cause and apply focused changes only

## Closing
Closes Suncrest-Labs/nester#236

### DIFF
--- a/packages/contracts/contracts/vault/src/lib.rs
+++ b/packages/contracts/contracts/vault/src/lib.rs
@@ -703,6 +703,8 @@ impl VaultContract {
         let total_assets = get_total_assets(&env);
         let accrued_fees = get_accrued_fees(&env);
         let mut assets_to_withdraw = vault_token_client(&env).amount_for_shares(&shares);
+        let current_principal = get_user_principal(&env, &user);
+        let principal_to_remove = current_principal * shares / current_shares;
 
         // Trigger circuit breaker check
         check_circuit_breaker(&env, assets_to_withdraw);
@@ -711,8 +713,8 @@ impl VaultContract {
         let config = get_fee_config(&env);
         let mut total_fee = 0_i128;
 
-        // 1. Performance fee (10% of yield)
-        let yield_part = assets_to_withdraw - shares;
+        // 1. Performance fee applies only to realized gain above user cost basis.
+        let yield_part = assets_to_withdraw - principal_to_remove;
         if yield_part > 0 {
             let perf_fee = nester_common::fees::calculate_performance_fee(
                 yield_part,
@@ -756,12 +758,6 @@ impl VaultContract {
         let new_user_shares = current_shares - shares;
         set_total_assets(&env, total_assets - assets_to_withdraw);
 
-        let current_principal = get_user_principal(&env, &user);
-        let principal_to_remove = if current_shares > 0 {
-            current_principal * shares / current_shares
-        } else {
-            0
-        };
         set_user_principal(&env, &user, current_principal - principal_to_remove);
 
         let current_reserves = get_vault_liquid_reserves(&env);
--- a/packages/contracts/contracts/vault/src/test.rs
+++ b/packages/contracts/contracts/vault/src/test.rs
@@ -7,6 +7,7 @@ use soroban_sdk::{
     testutils::{Address as _, Ledger, LedgerInfo},
     token, Address, Env, String,
 };
+use nester_access_control::Role;
 use vault_token::{VaultTokenContract, VaultTokenContractClient};
 
 use crate::{VaultContract, VaultContractClient, VaultStatus};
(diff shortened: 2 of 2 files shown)


0xDeon (approved):
Lgtm. Thanks for your cotribution

PR #276, opened 2026-04-23
fix: On-chain event indexer is defined but never started — vault state in the database is never updated from chain

### LINKED ISSUE #264: [Backend/Feature] On-chain event indexer is defined but never started — vault state in the database is never updated from chain
## Summary
`internal/stellar/event_indexer.go` defines an `EventIndexer` struct with a `SyncEvents()` method intended to watch the Stellar network for vault contract events (deposits, withdrawals, pauses, fee collection) and update the PostgreSQL database accordingly. This indexer is never started as a background process in `apps/api/cmd/main.go` and `SyncEvents()` appears to be a stub. As a result, the PostgreSQL database only reflects state from direct API calls — it has no awareness of on-chain events that occur independently. This means: vault balance displayed in the UI will be stale or wrong after any on-chain interaction that didn't go through the API; if the vault's own circuit breaker auto-pauses the vault on-chain, the API and UI will continue showing it as active; and emergency withdrawals processed on-chain by an operator will not be reflected in user portfolio views.

## Evidence

In `internal/stellar/event_indexer.go`:
```go
type EventIndexer struct {
    rpcClient   *SorobanRPCClient
    vaultRepo   VaultRepository
    logger      *log.Logger
}

func (e *EventIndexer) SyncEvents(ctx context.Context) error {
    // Stub — implementation is incomplete or missing
    return nil
}
```

In `apps/api/cmd/main.go`:
```go
func main() {
    // ... setup code ...
    // EventIndexer is never instantiated
    // No goroutine starts event polling
    // No event subscription is set up
}
```

## What State Becomes Stale Without the Indexer

| On-chain Event | API/UI Impact Without Indexer |
|----------------|-------------------------------|
| User deposits via DApp | Database balance may already be updated via the API call, but a deposit made directly via the contract (bypassing the API) is invisible |
| Circuit breaker auto-pauses vault | API shows vault as active; users can still attempt deposits that will fail on-chain |
| Admin calls unpause directly on-chain | API shows vault as paused; users see "vault paused" even though it's operational |
| Emergency withdrawal processed on-chain | User's portfolio balance in the API remains unchanged |
| Fee collection reduces liquid reserves | API's `liquid_reserves` value is stale |
| Exchange rate moves due to yield accrual | APY and value displayed in UI are frozen at last API-sync value |

## Impact
1. **Stale financial data** — users make deposit or withdrawal decisions based on incorrect balance and APY data shown in the UI
2. **Emergency response confusion** — if the vault is auto-paused by the circuit breaker, the admin dashboard shows it as active, potentially delaying manual response
3. **Trust erosion** — users who check their portfolio and see incorrect balances will question whether their funds are safe
4. **Reconciliation debt** — the longer the indexer remains unimplemented, the larger the divergence between on-chain truth and the database representation, and the more expensive the eventual reconciliation

### Step 1 — Implement SyncEvents using Soroban's getEvents RPC method

```go
// internal/stellar/event_indexer.go

func (e *EventIndexer) SyncEvents(ctx context.Context) error {
    // Use the Soroban getEvents RPC to subscribe to vault contract events
    filter := EventFilter{
        ContractIDs: e.vaultContractIDs,
        Topics: [][]string{
            {"deposit", "*"},
            {"withdraw", "*"},
            {"pause", "*"},
            {"unpause", "*"},
            {"fees_collected", "*"},
        },
        StartLedger: e.getLastIndexedLedger(),
    }
    
    events, err := e.rpcClient.GetEvents(ctx, filter)
    if err != nil {
        return fmt.Errorf("getEvents RPC failed: %w", err)
    }
    
    for _, event := range events {
        if err := e.processEvent(ctx, event); err != nil {
            e.logger.Errorf("failed to process event %s: %v", event.ID, err)
            // Continue processing other events — don't abort on single failure
        }
    }
    
    e.setLastIndexedLedger(events.LatestLedger)
    return nil
}

func (e *EventIndexer) processEvent(ctx context.Context, event SorobanEvent) error {
    switch event.Topic[0] {
    case "deposit":
        return e.handleDeposit(ctx, event)
    case "withdraw":
        return e.handleWithdraw(ctx, event)
    case "pause":
        return e.vaultRepo.SetPaused(ctx, event.ContractID, true)
    case "unpause":
        return e.vaultRepo.SetPaused(ctx, event.ContractID, false)
    }
    return nil
}
```

### Step 2 — Start the indexer as a background goroutine in main.go
```go
// main.go
indexer := stellar.NewEventIndexer(rpcClient, vaultRepo, logger)

go func() {
    ticker := time.NewTicker(6 * time.Second) // Poll every ~1 ledger
    defer ticker.Stop()
    for {
        select {
        case <-ctx.Done():
            return
        case <-ticker.C:
            if err := indexer.SyncEvents(ctx); err != nil {
                logger.Errorf("event sync failed: %v", err)
            }
        }
    }
}()
```

### Step 3 — Add a manual sync endpoint for recovery
```go
// POST /api/v1/admin/sync-events
// Triggers a full resync from ledger 0 (or a specified ledger)
// Useful for recovering from indexer downtime or catching up after a gap
```

### Step 4 — Track last indexed ledger in the database
Persist the last successfully indexed Stellar ledger sequence number in a `system_state` table so restarts resume from where they left off rather than reprocessing all history.

## References
- [Stellar Soroban — getEvents RPC method](https://developers.stellar.org/docs/data/rpc/api-reference/methods/getEvents)
- [Event-driven architecture for blockchain indexers](https://docs.thegraph.com/en/indexing/)
- The Graph Protocol — decentralized indexing reference architecture

wheval:
## Summary
- #264: On-chain event indexer is defined but never started — vault state in the database is never updated from chain

## Implementation notes
- Issue #264: validate root cause and apply focused changes only

## Closing
Closes Suncrest-Labs/nester#264

### DIFF
--- a/apps/api/cmd/api/main.go
+++ b/apps/api/cmd/api/main.go
@@ -96,6 +96,11 @@ func run() error {
 		cfg.SettlementProviderURL(),
 	)
 	adminHandler := handler.NewAdminHandler(adminService)
+	adminHandler.SetEventSyncer(&mainEventSyncer{
+		db:     db,
+		rpcURL: cfg.Stellar().RPCURL(),
+		logger: baseLogger,
+	})
 
 	var challengeStore service.ChallengeStore
 	if addr := cfg.Redis().Addr(); addr != "" {
@@ -605,4 +610,56 @@ ON CONFLICT (event_id) DO NOTHING`,
 		return false, err
 	}
 	return rowsAffected == 1, nil
+}
+
+// mainEventSyncer implements handler.EventSyncer using the same indexer
+// helpers defined in this package, allowing the admin handler to trigger
+// a manual one-shot sync for recovery purposes.
+type mainEventSyncer struct {
+	db     *sql.DB
+	rpcURL string
+	logger *slog.Logger
+}
+
+func (s *mainEventSyncer) SyncEvents(ctx context.Context) (int, error) {
+	if err := ensureIndexerTables(ctx, s.db); err != nil {
+		return 0, fmt.Errorf("indexer tables: %w", err)
+	}
+
+	startLedger, err := getLastIndexedLedger(ctx, s.db)
+	if err != nil {
+		return 0, fmt.Errorf("load cursor: %w", err)
+	}
+
+	contractIDs, err := loadVaultContractIDs(ctx, s.db)
+	if err != nil {
+		return 0, fmt.Errorf("load contracts: %w", err)
+	}
+	if len(contractIDs) == 0 {
+		return 0, nil
+	}
+
+	client := &http.Client{Timeout: 30 * time.Second}
+	events, latestLedger, err := fetchSorobanEvents(ctx, client, s.rpcURL, contractIDs, startLedger)
+	if err != nil {
+		return 0, fmt.Errorf("fetch events: %w", err)
+	}
+
+	processed := 0
+	for _, event := range events {
+		ok, err := applyIndexedEvent(ctx, s.db, event)
+		if err != nil {
+			s.logger.Error("admin sync: failed to apply event", "event_id", event.ID, "error", err)
+			continue
+		}
+		if ok {
+			processed++
+		}
+	}
+
+	if err := setLastIndexedLedger(ctx, s.db, latestLedger); err != nil {
+		s.logger.Error("admin sync: failed to persist cursor", "ledger", latestLedger, "error", err)
+	}
+
+	return processed, nil
 }
--- a/apps/api/cmd/api/main.go
+++ b/apps/api/cmd/api/main.go
@@ -1,17 +1,24 @@
 package main
 
 import (
+	"bytes"
 	"context"
+	"database/sql"
+	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
+	"log/slog"
 	"net/http"
 	"os"
 	"os/signal"
+	"strings"
 	"syscall"
 	"time"
 
 	"github.com/jackc/pgx/v5/stdlib"
 	"github.com/redis/go-redis/v9"
+	"github.com/shopspring/decimal"
 
 	"github.com/suncrestlabs/nester/apps/api/internal/auth"
 	"github.com/suncrestlabs/nester/apps/api/internal/config"
@@ -168,6 +175,8 @@ func run() error {
 	shutdownCtx, stop := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
 	defer stop()
 
+	startEventIndexer(shutdownCtx, baseLogger, db, cfg.Stellar().RPCURL())
+
 	serverErr := make(chan error, 1)
 	go func() {
 		err := server.ListenAndServe()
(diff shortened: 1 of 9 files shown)


apps/api/cmd/api/main.go
```diff
@@ -119,6 +127,8 @@ func run() error {
 	shutdownCtx, stop := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
 	defer stop()
 
+	startEventIndexer(shutdownCtx, baseLogger, db, cfg.Stellar().RPCURL())
+
 	serverErr := make(chan error, 1)
 	go func() {
```
0xDeon, line 186: `startLedger` is a local variable — it resets to 0 on every API restart. The deposit and withdraw update queries are additive (`total_deposited + amount`), so every restart replays all historical events from ledger 0 and doubles every vault balance. The last indexed ledger must be persisted to the database (e.g. a `system_state` table) and read on startup so the indexer resumes from where it left off rather than replaying from the beginning.

apps/api/cmd/api/main.go
```diff
@@ -167,3 +177,247 @@ func healthHandler(db *repository.PostgresDB, timeout time.Duration) http.Handle
...
+				if len(contractIDs) == 0 {
+					continue
+				}
+
+				events, latestLedger, err := fetchSorobanEvents(ctx, client, rpcURL, contractIDs, startLedger)
+				if err != nil {
+					logger.Error("event indexer fetch failed", "error", err)
+					continue
+				}
+
```
0xDeon, line 285: This update is not idempotent. If events are replayed (restart, catchup), `total_deposited` and `current_balance` accumulate duplicate additions. Either persist processed event IDs and skip duplicates, or use absolute SET values sourced from on-chain state instead of additive increments.

apps/api/cmd/api/main.go
```diff
@@ -167,3 +177,247 @@ func healthHandler(db *repository.PostgresDB, timeout time.Duration) http.Handle
...
+			event.ContractID,
+		)
+		return err
+	case "withdraw", "withdrawal":
+		amount, ok := extractEventAmount(event)
+		if !ok {
+			return fmt.Errorf("withdraw event missing parseable amount")
+		}
+		_, err := db.ExecContext(
+			ctx,
```
0xDeon, line 399: `float64` cannot represent large financial amounts precisely (max 15 significant digits). On-chain Stellar amounts are integers — if the event value arrives as a JSON number, parse it as a string via `fmt.Sprintf("%v", v)` or reject it. Using `decimal.NewFromFloat` on a float64 can silently corrupt amounts for large vaults.

0xDeon (changes requested):
Three blocking issues: (1) startLedger resets to 0 on every API restart — all balance updates are additive so replaying history doubles every vault balance in the database. (2) The existing EventPoller.PollEvents in internal/stellar/events.go remains a stub returning empty events — this PR adds a parallel implementation in main.go instead of fixing the actual infrastructure. (3) No tests on any function that writes to the financial database.

PR #277, opened 2026-04-23
fix(vault): add slippage protection for deposit and withdraw

### LINKED ISSUE #248: [Vault] No slippage protection on deposit and withdrawal — users exposed to unexpected losses without recourse
## Summary
The vault contract's `deposit()` and `withdraw()` functions accept only a raw amount parameter with no minimum output guarantee (slippage protection). In a DeFi vault, the exchange rate between shares and assets can move between when a user constructs a transaction and when it is included in a ledger — particularly during periods of high activity, yield accrual events, or fee collection. Without slippage protection, a user who submits a deposit expecting 100 shares might receive 92 shares by the time their transaction is processed, with no ability to opt out. For Nester's target users in emerging markets who may be making deposits that represent a significant portion of their savings, unexpected slippage without any error or warning is a serious UX and financial issue.

## Affected Function Signatures

```rust
// packages/contracts/contracts/vault/src/lib.rs

pub fn deposit(env: Env, from: Address, amount: i128) -> i128 {
    // No min_shares_out parameter — user has no way to set a minimum
}

pub fn withdraw(env: Env, from: Address, shares: i128) -> i128 {
    // No min_assets_out parameter — user has no way to set a minimum
}
```

Compare to industry standard (EIP-4626):
```
deposit(assets, receiver) returns (shares)  // Optional: add minSharesOut
withdraw(assets, receiver, owner)           // Optional: add minAssetsOut
```

### Scenario A — Fee collection between submit and inclusion
1. User constructs a deposit transaction for 1,000 USDC expecting ~1,000 shares at rate 1.0
2. Before the transaction is included, an admin calls `collect_fees()` which reduces `liquid_reserves`
3. The management fee calculation changes the effective exchange rate
4. User's transaction is included and they receive 987 shares — a 1.3% slippage with no warning

### Scenario B — Large deposit sandwiching
1. User submits a deposit transaction visible in the XDR stream
2. A sophisticated actor submits a large deposit immediately before the user's, diluting shares
3. User's shares are fewer than expected; the sandwicher immediately withdraws at the new rate
4. This is the Stellar equivalent of an MEV sandwich attack

### Scenario C — Yield accrual between submit and inclusion
If the vault accrues yield (via the APY relayer calling into the yield registry) between when the user constructs and when their transaction is included, the exchange rate moves and they receive a different number of shares than their wallet UI displayed at the time of signing.

## Impact
1. **Unexpected losses with no recourse** — users cannot opt out of a transaction that would result in worse-than-expected output; the contract will succeed regardless
2. **Trust erosion** — users who deposit and receive fewer shares than shown in the UI preview will blame the protocol
3. **Economic attack surface** — without slippage limits, more sophisticated MEV-equivalent strategies on Stellar become viable
4. **Compliance risk** — in some regulatory frameworks, executing a financial transaction at materially worse terms than displayed without the user's consent may have legal implications

### Add min_shares_out to deposit()
```rust
pub fn deposit(env: Env, from: Address, amount: i128, min_shares_out: i128) -> i128 {
    from.require_auth();
    
    let shares = calculate_shares(&env, amount);
    
    if shares < min_shares_out {
        panic_with_error!(&env, VaultError::SlippageExceeded);
    }
    
    // Proceed with deposit
    mint_shares(&env, &from, shares);
    shares
}
```

### Add min_assets_out to withdraw()
```rust
pub fn withdraw(env: Env, from: Address, shares: i128, min_assets_out: i128) -> i128 {
    from.require_auth();
    
    let assets = calculate_assets(&env, shares);
    
    if assets < min_assets_out {
        panic_with_error!(&env, VaultError::SlippageExceeded);
    }
    
    // Proceed with withdrawal
    burn_shares(&env, &from, shares);
    transfer_assets(&env, &from, assets);
    assets
}
```

### Frontend integration
The DApp should calculate expected output, apply a configurable slippage tolerance (default 0.5%), and pass the minimum:
```typescript
const expectedShares = await vault.previewDeposit(amount);
const minSharesOut = expectedShares * (1 - slippageTolerance); // e.g., 99.5% of expected
await vault.deposit(amount, Math.floor(minSharesOut));
```

### Add a `preview_deposit` and `preview_withdraw` view function
These pure functions let the frontend display expected output before the user signs:
```rust
pub fn preview_deposit(env: Env, amount: i128) -> i128 {
    calculate_shares(&env, amount)
}

pub fn preview_withdraw(env: Env, shares: i128) -> i128 {
    calculate_assets(&env, shares)
}
```

## References
- [EIP-4626 — Tokenized Vault Standard: slippage protection](https://eips.ethereum.org/EIPS/eip-4626)
- [Uniswap v2: amountOutMin pattern](https://docs.uniswap.org/contracts/v2/reference/smart-contracts/router-02#swapexacttokensfortokens)
- [Yearn Finance vault deposit/withdraw with slippage](https://github.com/yearn/yearn-vaults)

wheval:
## Summary
- add slippage-safe vault entrypoints by introducing `min_shares_out` for `deposit` and `min_assets_out` for `withdraw`
- revert with `ContractError::SlippageExceeded` when output falls below caller-specified minimums
- add `preview_deposit` / `preview_withdraw` views and update vault + integration tests to pass explicit slippage params

## Test plan
- [x] Updated unit/integration call sites to new signatures
- [x] Added slippage regression tests for deposit and withdraw revert paths
- [ ] Run `cargo test -p vault-contract` in CI environment with `vault_token.wasm` artifact available

Closes Suncrest-Labs/nester#248

### DIFF
--- a/packages/contracts/contracts/vault/src/lib.rs
+++ b/packages/contracts/contracts/vault/src/lib.rs
@@ -580,7 +580,7 @@ impl VaultContract {
     }
 
     /// Deposit funds into the vault.
-    pub fn deposit(env: Env, user: Address, amount: i128) -> i128 {
+    pub fn deposit(env: Env, user: Address, amount: i128, min_shares_out: i128) -> i128 {
         require_initialized(&env);
         require_active(&env);
 
@@ -596,6 +596,9 @@ impl VaultContract {
         if amount <= 0 {
             panic_with_error!(&env, ContractError::InvalidAmount);
         }
+        if min_shares_out < 0 {
+            panic_with_error!(&env, ContractError::InvalidAmount);
+        }
 
         user.require_auth();
         accrue_management_fee(&env);
@@ -606,7 +609,11 @@ impl VaultContract {
         token::Client::new(&env, &token_address).transfer(&user, &contract_address, &amount);
 
         let total_assets = get_total_assets(&env);
-        let shares_to_mint = vault_token_client(&env).mint_for_deposit(&user, &amount);
+        let shares_to_mint = vault_token_client(&env).shares_for_deposit(&amount);
+        if shares_to_mint < min_shares_out {
+            panic_with_error!(&env, ContractError::SlippageExceeded);
+        }
+        let _ = vault_token_client(&env).mint_for_deposit(&user, &amount);
         let new_user_shares = get_shares(&env, &user);
         set_total_assets(&env, total_assets + amount);
 
@@ -684,11 +691,14 @@ impl VaultContract {
     }
 
     /// Withdraw funds from the vault.
-    pub fn withdraw(env: Env, user: Address, shares: i128) -> i128 {
+    pub fn withdraw(env: Env, user: Address, shares: i128, min_assets_out: i128) -> i128 {
         require_initialized(&env);
         require_active(&env);
 
         if shares <= 0 {
+            panic_with_error!(&env, ContractError::InvalidAmount);
+        }
+        if min_assets_out < 0 {
             panic_with_error!(&env, ContractError::InvalidAmount);
         }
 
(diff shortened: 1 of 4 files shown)


0xDeon:
@wheval Fix conflicts please

0xDeon (approved):
Lgtm

PR #282, opened 2026-04-23
feat(backend): replace in-memory challenge store with Redis-backed persistence

### LINKED ISSUE #253: [Backend] Auth challenge store is in-memory — lost on restart, unsafe under load balancing, no TTL eviction
## Summary
The wallet-based challenge/response authentication flow in `apps/api/internal/auth/` stores pending challenges in a plain Go `map[string]Challenge` in process memory. This creates three distinct problems: (1) every server restart clears all in-flight challenges, forcing users to restart the sign-in flow with no useful error message; (2) under horizontal scaling or load balancing, a challenge generated by instance A cannot be verified by instance B because they have separate memory spaces; and (3) there is no TTL-based eviction of unverified challenges, so a flood of challenge generation requests (from bots or automated scripts) can grow the map unboundedly and eventually exhaust server memory, creating a denial-of-service condition.

## Evidence

In `apps/api/internal/auth/` challenge generation:
```go
// In-memory store — no persistence, no TTL, no size limit
var challengeStore = struct {
    sync.RWMutex
    challenges map[string]Challenge
}{challenges: make(map[string]Challenge)}

func GenerateChallenge(walletAddress string) (string, error) {
    challenge := generateRandomHex(32)
    
    challengeStore.Lock()
    challengeStore.challenges[walletAddress] = Challenge{
        Value:     challenge,
        CreatedAt: time.Now(),
        // ExpiresAt: never set, or not enforced
    }
    challengeStore.Unlock()
    
    return challenge, nil
}
```

There is no goroutine cleaning up expired challenges. If 100,000 unique wallet addresses generate challenges and never verify them, 100,000 entries accumulate in memory indefinitely.

### Scenario A — Server restart during sign-in
1. User opens the Nester DApp for the first time
2. DApp calls `POST /auth/challenge` → server stores challenge in memory, returns challenge hex
3. User opens Freighter wallet, is about to sign
4. Server restarts (scheduled deployment, OOM kill, etc.)
5. User signs and sends `POST /auth/verify` with the signature
6. Server responds: "challenge not found" — user sees a generic error
7. User has no idea what happened; must retry from scratch

### Scenario B — Load balancer distributes requests across instances
1. Request to `POST /auth/challenge` goes to Instance A → challenge stored in Instance A's memory
2. Request to `POST /auth/verify` goes to Instance B → "challenge not found"
3. Authentication fails with no useful error even though both server and user did everything correctly

### Scenario C — Memory exhaustion via challenge flooding
```bash
# Attacker generates challenges for millions of wallet addresses
for i in $(seq 1 1000000); do
    curl -X POST /auth/challenge -d "{\"walletAddress\": \"G${RANDOM}...\"}" &
done
```
Each request allocates a map entry. No eviction. Server OOMs and crashes.

## Impact
1. **Authentication failures under normal operation** — any multi-instance deployment will have intermittent auth failures, degrading user experience and appearing as random sign-in bugs
2. **DoS via memory exhaustion** — challenge flooding causes OOM, taking down the entire API
3. **Poor user experience at scale** — as deployment scales, the probability of cross-instance auth failures increases linearly with the number of instances
4. **Operations complexity** — rolling deployments (the standard for zero-downtime upgrades) will always cause sign-in failures for users in the middle of the auth flow

### Replace in-memory store with Redis using TTL

```go
// apps/api/internal/auth/challenge_store.go

type RedisChallengeStore struct {
    client *redis.Client
    ttl    time.Duration
}

func NewRedisChallengeStore(client *redis.Client, ttl time.Duration) *RedisChallengeStore {
    return &RedisChallengeStore{client: client, ttl: ttl}
}

func (s *RedisChallengeStore) Set(ctx context.Context, walletAddress, challenge string) error {
    key := fmt.Sprintf("auth:challenge:%s", walletAddress)
    return s.client.Set(ctx, key, challenge, s.ttl).Err()
}

func (s *RedisChallengeStore) GetAndDelete(ctx context.Context, walletAddress string) (string, error) {
    key := fmt.Sprintf("auth:challenge:%s", walletAddress)
    // GETDEL is atomic — prevents replay attacks where the same challenge is verified twice
    challenge, err := s.client.GetDel(ctx, key).Result()
    if err == redis.Nil {
        return "", ErrChallengeNotFound
    }
    return challenge, err
}
```

### Set an appropriate TTL
```go
// Challenges expire after 5 minutes — enough for a user to open their wallet and sign
store := NewRedisChallengeStore(redisClient, 5*time.Minute)
```

### Wire Redis in main.go
```go
redisClient := redis.NewClient(&redis.Options{Addr: cfg.RedisAddr})
challengeStore := auth.NewRedisChallengeStore(redisClient, 5*time.Minute)
authService := service.NewAuthService(challengeStore, jwtSecret, tokenTTL)
```

### Add Redis to infrastructure
Add Redis to `docker-compose.yml` (for local dev) and to the production deployment config. Redis is already a likely requirement for the rate limiter migration as well.

## References
- [Redis GETDEL command](https://redis.io/docs/latest/commands/getdel/) — atomic get-and-delete for nonce patterns
- [WalletConnect v2 — challenge nonce storage in Redis](https://docs.walletconnect.com/advanced/auth)
- [OWASP — Session Management Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/Session_Management_Cheat_Sheet.html)

Salmatcre8:
## Summary

- Introduces a `ChallengeStore` interface with `Set` / `GetAndDelete` that `authService` delegates to — removing the `sync.RWMutex` + `map` it previously owned
- `RedisChallengeStore` uses `GETDEL` (atomic get-and-delete) so a challenge can only be verified once even under concurrent requests; TTL is enforced by Redis natively
- `InMemoryChallengeStore` is retained as a single-instance fallback (and for tests) — active when `REDIS_ADDR` is empty
- `main.go` builds the store at startup and logs which backend is active
- `docker-compose.yml` adds a Redis 7 Alpine service; the API service waits for it to be healthy before starting
- `apps/api/.env.example` documents `REDIS_ADDR`

## Why each problem is now fixed

| Scenario | Before | After |
|---|---|---|
| Server restart mid-auth | Challenges lost; user sees "challenge not found" | Challenges survive in Redis |
| Load-balanced instances | Instance A stores challenge; Instance B can't find it | Both instances share Redis |
| Challenge-flood DoS | Unbounded map growth → OOM | Redis enforces TTL; entries evict automatically |

## Test plan

- [x] `InMemoryChallengeStore`: 6 unit tests (set/get, one-time use, missing key, expiry, overwrite, wallet isolation)
- [x] `authService` tests updated to pass a store; all 6 existing tests still pass logic-unchanged
- [x] `GOOS=linux go build ./...` clean
- [x] `GOOS=linux go test -c ./internal/service/` compiles with no errors

Closes #253

### DIFF
--- a/apps/api/.env.example
+++ b/apps/api/.env.example
@@ -25,6 +25,10 @@ DATABASE_POOL_SIZE=25
 # Timeout used when opening the initial database connection and pinging.
 DATABASE_CONNECTION_TIMEOUT=5s
 
+# Redis address for the challenge store (host:port).
+# When empty, falls back to an in-memory store (single-instance only).
+REDIS_ADDR=localhost:6379
+
 # Stellar network passphrase.
 STELLAR_NETWORK_PASSPHRASE=Test SDF Network ; September 2015
 
--- a/apps/api/cmd/api/main.go
+++ b/apps/api/cmd/api/main.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	"github.com/jackc/pgx/v5/stdlib"
+	"github.com/redis/go-redis/v9"
 
 	"github.com/suncrestlabs/nester/apps/api/internal/config"
 	"github.com/suncrestlabs/nester/apps/api/internal/handler"
@@ -70,7 +71,16 @@ func run() error {
 	)
 	adminHandler := handler.NewAdminHandler(adminService)
 
-	authService := service.NewAuthService(userService, cfg.Auth())
+	var challengeStore service.ChallengeStore
+	if addr := cfg.Redis().Addr(); addr != "" {
+		redisClient := redis.NewClient(&redis.Options{Addr: addr})
+		challengeStore = service.NewRedisChallengeStore(redisClient, cfg.Auth().ChallengeExpiry())
+		baseLogger.Info("challenge store: redis", "addr", addr)
+	} else {
+		challengeStore = service.NewInMemoryChallengeStore(cfg.Auth().ChallengeExpiry())
+		baseLogger.Info("challenge store: in-memory (single-instance only)")
+	}
+	authService := service.NewAuthService(challengeStore, userService, cfg.Auth())
 	authHandler := handler.NewAuthHandler(authService)
 
 	mux := http.NewServeMux()
--- a/apps/api/go.mod
+++ b/apps/api/go.mod
@@ -14,7 +14,9 @@ require (
 )
 
 require (
+	github.com/cespare/xxhash/v2 v2.3.0 // indirect
 	github.com/davecgh/go-spew v1.1.2-0.20180830191138-d8f796af33cc // indirect
+	github.com/dgryski/go-rendezvous v0.0.0-20200823014737-9f7001d12a5f // indirect
 	github.com/gabriel-vasile/mimetype v1.4.12 // indirect
 	github.com/go-playground/locales v0.14.1 // indirect
 	github.com/go-playground/universal-translator v0.18.1 // indirect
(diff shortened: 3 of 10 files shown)


0xDeon (approved):
Lgtm

Issue #251, opened 2026-04-20
[Backend] API CORS policy allows all origins — enables cross-site request forgery against authenticated users


0xDeon:
## Summary
`apps/api/internal/middleware/cors.go` sets `Access-Control-Allow-Origin: *` for all API responses. A wildcard CORS policy instructs browsers to allow any website on the internet to make cross-origin XMLHttpRequest and fetch calls to the Nester API and read the full response bodies. For a DeFi API that handles wallet authentication, settlement requests with bank account details, vault management, and user portfolio data, this is a significant over-permission. While Bearer token authentication (as opposed to cookie-based auth) reduces the direct CSRF risk, a wildcard CORS policy enables sophisticated attacks where a malicious site silently reads authenticated API responses if it can obtain the user's JWT through any means (XSS, local storage theft, or if the token is ever passed as a URL parameter).

## Evidence

In `apps/api/internal/middleware/cors.go`:
```go
func CORSMiddleware(next http.Handler) http.Handler {
    return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
        w.Header().Set("Access-Control-Allow-Origin", "*")
        w.Header().Set("Access-Control-Allow-Methods", "GET, POST, PUT, PATCH, DELETE, OPTIONS")
        w.Header().Set("Access-Control-Allow-Headers", "Content-Type, Authorization")
        // ...
    })
}
```

### Cross-Origin Reading of API Responses
With `Access-Control-Allow-Origin: *`, any website can read the response body of API calls (when the browser sends the JWT in the Authorization header). A malicious site that has obtained a user's JWT (e.g., through an XSS vulnerability in any site the user visits) can make authenticated API calls to Nester and read the results — including portfolio balances, settlement history, bank account details, and user profile information.

### Example Attack
```javascript
// On evil.com, after obtaining user's JWT via XSS on another site:
const response = await fetch('https://api.nester.finance/api/v1/users/me/vaults', {
    headers: { 'Authorization': `Bearer ${stolenJWT}` }
});
const vaults = await response.json(); // Browser allows this because of wildcard CORS
// evil.com now has the user's full portfolio data
```

### Pre-flight Leakage
Even unauthenticated API metadata (endpoint discovery, error messages, version headers) is accessible to any origin, which can aid in targeted attacks against the API.

## Impact
[...]

PR #288, opened 2026-04-24
Fix/wallet rate limiter unused

### LINKED ISSUE #252: [Backend] Per-wallet rate limiter is defined but never wired — only coarse IP-based limits are active
## Summary
`apps/api/internal/middleware/` defines a `WalletRateLimiter` type designed to enforce per-wallet-address rate limits on authenticated API endpoints. This granular rate limiting is essential for a DeFi protocol: a single compromised or malicious wallet address should not be able to spam vault creation, settlement requests, challenge generation, or other state-changing operations regardless of how many IP addresses it uses. The limiter is fully defined in code but is never instantiated or applied in `main.go` or any route registration. As a result, only the coarse global IP-based token bucket is active, which provides no protection against: wallet-level abuse from a single IP, distributed wallet spam from multiple IPs, or API abuse from behind shared NAT (where many legitimate users share one IP).

## Evidence

In `apps/api/internal/middleware/rate_limit.go`:
```go
type WalletRateLimiter struct {
    mu      sync.Mutex
    buckets map[string]*tokenBucket
    rate    float64
    burst   int
}

func NewWalletRateLimiter(rate float64, burst int) *WalletRateLimiter {
    return &WalletRateLimiter{
        buckets: make(map[string]*tokenBucket),
        rate:    rate,
        burst:   burst,
    }
}

func (w *WalletRateLimiter) Middleware(next http.Handler) http.Handler {
    return http.HandlerFunc(func(rw http.ResponseWriter, r *http.Request) {
        // Extracts wallet address from JWT claims and applies per-wallet limit
        // This code is correct and complete — but is never called
    })
}
```

In `apps/api/cmd/main.go`:
```go
// Global IP rate limiter is wired:
ipLimiter := middleware.NewIPRateLimiter(cfg.RateLimitGlobal, cfg.RateLimitWrite)
router.Use(ipLimiter.Middleware)

// WalletRateLimiter is NEVER referenced here — the struct exists but is orphaned
```

## Scenarios That Are Not Rate Limited

| Attack | IP Limiter Catches? | Wallet Limiter Would Catch? |
|--------|--------------------|-----------------------------|
| Single wallet spams 1000 vault creates from one IP | Yes (IP limit) | Yes |
| Single wallet spams 1000 vault creates from 100 IPs (1 per 10 reqs) | No — 10 reqs/IP is under limit | Yes |
| 100 bots each with one wallet, coordinated attack | No — distributed below IP threshold | Yes (each wallet limited individually) |
| Shared NAT: 500 legitimate users from same office IP | False positive rate limiting | No false positives |

The last row highlights why IP-only rate limiting is also problematic: in emerging markets where corporate NAT or carrier-grade NAT is common, a single IP may represent hundreds of legitimate users who will all be rate-limited together.

## Impact
1. **Wallet-level abuse** — a malicious actor using rotating IP addresses (VPN, Tor, proxies) faces no per-wallet rate limit and can spam any authenticated endpoint freely
2. **Settlement flooding** — an attacker can create thousands of settlement requests from a single wallet using multiple IPs, flooding the settlement queue and denying service to legitimate users
3. **Challenge replay flooding** — authentication challenge generation can be flooded at the wallet level to exhaust memory (see related in-memory challenge store issue)
4. **Legitimate user false positives** — IP-based rate limiting may block entire office networks or mobile carrier subnets in regions with scarce IP space

### Step 1 — Instantiate and apply the WalletRateLimiter in main.go
```go
// main.go — after the existing IP limiter setup
walletLimiter := middleware.NewWalletRateLimiter(
    cfg.WalletRateLimitPerMinute,  // e.g., 30 requests per minute per wallet
    cfg.WalletRateLimitBurst,      // e.g., 10 burst
)

// Apply to all authenticated routes
authRouter.Use(walletLimiter.Middleware)
```

### Step 2 — Add wallet rate limit config values
```bash
# .env.example
WALLET_RATE_LIMIT_PER_MINUTE=30
WALLET_RATE_LIMIT_BURST=10
```

### Step 3 — Add endpoint-specific tighter limits for sensitive operations
```go
// Tighter limits for settlement creation
settlementLimiter := middleware.NewWalletRateLimiter(5, 2) // 5 per minute
authRouter.With(settlementLimiter.Middleware).Post("/settlements", handler.CreateSettlement)

// Even tighter for auth challenge generation (prevent memory exhaustion)
challengeLimiter := middleware.NewIPRateLimiter(10, 3) // 10 per minute per IP
publicRouter.With(challengeLimiter.Middleware).Post("/auth/challenge", handler.GenerateChallenge)
```

### Step 4 — Migrate to Redis-backed rate limiting for multi-instance support
The current in-memory bucket implementation does not work correctly behind a load balancer. Use Redis:
```go
// Using go-redis + a sliding window script
func (w *WalletRateLimiter) isAllowed(ctx context.Context, walletAddress string) bool {
    key := fmt.Sprintf("rate:wallet:%s", walletAddress)
    count, _ := redisClient.Incr(ctx, key).Result()
    if count == 1 {
        redisClient.Expire(ctx, key, time.Minute)
    }
    return count <= w.limit
}
```

## References
- [OWASP API Security Top 10 — API4: Unrestricted Resource Consumption](https://owasp.org/API-Security/editions/2023/en/0xa4-unrestricted-resource-consumption/)
- [golang.org/x/time/rate — token bucket implementation](https://pkg.go.dev/golang.org/x/time/rate)
- [go-redis sliding window rate limiter](https://redis.io/docs/manual/patterns/rate-limiting/)

deltron-fr:
Closes #252 and #251 

---

 ## Summary

  - Wires the existing but unused `middleware.WalletRateLimiter` into the production chain in `cmd/api/main.go`, immediately after `Authenticate` so the bucket is keyed by the wallet address in the JWT claims (not the caller's IP). A single wallet can no longer bypass rate limits by rotating IPs across VPNs, proxies, or Tor.
  - Adds `RATELIMIT_WALLET_LIMIT` (default `60`) and `RATELIMIT_WALLET_WINDOW` (default `1m`) to config with the same validation as the existing global and write limits. Documented in `.env.example`.
  - Adds a `walletKeyFromContext` helper that pulls the wallet from the authenticated `auth.User`; unauthenticated requests produce an empty key and pass through the wallet limiter (public routes are still covered by the IP and write limiters that wrap the chain outside of auth).
 
  - Replaces the wildcard Access-Control-Allow-Origin: * in apps/api/internal/middleware/cors.go with an origin allowlist: the request's Origin is echoed back only when it matches ALLOWED_ORIGINS, Vary: Origin is always set, and Access-Control-Allow-Credentials: true plus the preflight method/header hints are emitted only for allowed origins.
  - Adds ALLOWED_ORIGINS to config.go (comma-separated). Validation rejects *, non-http(s) schemes, and any path/query/fragment. A non-empty list is required in production/staging; development/test may leave it empty (CORS disabled).
  - Wires middleware.CORS into the production chain in cmd/api/main.go: it was previously only reachable via the test-only server.New constructor. CORS sits inside rate limiting (preflights count against the bucket) and outside auth (preflights short-circuit with 204 before Authenticate can reject them).

### DIFF
--- a/apps/api/.env.example
+++ b/apps/api/.env.example
@@ -43,3 +43,10 @@ LOG_LEVEL=info
 
 # Log output format: text or json. Defaults to text outside production/staging.
 LOG_FORMAT=text
+
+# Comma-separated list of origins permitted to make cross-origin browser
+# requests to the API. Entries must be full origins (scheme://host[:port])
+# with no path or trailing slash. Wildcards are rejected.
+# Required in production and staging.
+# Example (production): https://app.nester.finance,https://nester.finance
+ALLOWED_ORIGINS=http://localhost:3000
--- a/apps/api/.env.example
+++ b/apps/api/.env.example
@@ -50,3 +50,9 @@ LOG_FORMAT=text
 # Required in production and staging.
 # Example (production): https://app.nester.finance,https://nester.finance
 ALLOWED_ORIGINS=http://localhost:3000
+
+# Per-wallet rate limit for authenticated requests. Buckets are keyed by the
+# wallet address in JWT claims, so a single wallet cannot bypass the limit by
+# rotating IPs. Unauthenticated requests are unaffected.
+RATELIMIT_WALLET_LIMIT=60
+RATELIMIT_WALLET_WINDOW=1m
--- a/apps/api/cmd/api/main.go
+++ b/apps/api/cmd/api/main.go
@@ -13,6 +13,7 @@ import (
 	"github.com/jackc/pgx/v5/stdlib"
 	"github.com/redis/go-redis/v9"
 
+	"github.com/suncrestlabs/nester/apps/api/internal/auth"
 	"github.com/suncrestlabs/nester/apps/api/internal/config"
 	"github.com/suncrestlabs/nester/apps/api/internal/handler"
 	"github.com/suncrestlabs/nester/apps/api/internal/middleware"
@@ -127,6 +128,14 @@ func run() error {
 	globalLimiter := middleware.IPRateLimiter(cfg.RateLimit().GlobalLimit(), cfg.RateLimit().GlobalWindow())
 	// Write rate limit is stricter and applies only to mutating methods (POST/PUT/PATCH/DELETE).
 	writeLimiter := middleware.WriteMethodRateLimiter(cfg.RateLimit().WriteLimit(), cfg.RateLimit().WriteWindow())
+	// Wallet rate limit runs inside Authenticate so the caller's wallet address
+	// is in context. Unauthenticated requests produce an empty key and pass
+	// through — public routes are covered by the IP and write limiters above.
+	walletLimiter := middleware.WalletRateLimiter(
+		cfg.RateLimit().WalletLimit(),
+		cfg.RateLimit().WalletWindow(),
+		walletKeyFromContext,
+	)
 	// CORS sits inside rate limiting (preflights count against the bucket) and
 	// outside auth (preflights don't carry credentials and must short-circuit
 	// before Authenticate rejects them).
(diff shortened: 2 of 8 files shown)


0xDeon (approved):
Lgtm

PR #289, opened 2026-04-25
fix(devops): remove committed binary artifacts and add intelligence .env.example

### LINKED ISSUE #184: fix(devops): remove committed .env file and binary artifacts from repository
## Summary

Several files that should never be in version control are currently committed to the repository: a `.env` file with configuration values and two compiled binary files totaling ~16MB. These should be removed and properly gitignored.

### 1. apps/intelligence/.env

Look at `apps/intelligence/.env` — this file contains environment configuration (NODE_ENV, HOST, PORT, CORS settings, rate limit values). While none of these appear to be sensitive secrets in the current state, committing `.env` files sets a dangerous precedent. Future contributors may add actual API keys to this file not realizing it's tracked by git.

The file should be:
- Removed from git tracking
- Renamed to `.env.example` (or verified that the existing `.env.example` or template covers these values)
- Added to `.gitignore` specifically for the intelligence directory

Note: `apps/intelligence/.env.template` was previously the template file (now deleted in the Node.js → Python migration). The current Python setup uses `apps/intelligence/app/config.py` for configuration, so the `.env` file may not even be used anymore. Verify and clean up.

### 2. services/api/bin/server (8MB)

A compiled Go binary for Linux. Should never be in git — binaries are build artifacts that should be generated by CI/CD or locally.

### 3. services/api/server.exe (8MB)

A compiled Go binary for Windows. Same issue.

### Step 1: Update .gitignore

Verify the root `.gitignore` covers:
- `*.exe` — Windows executables
- `bin/` — compiled binary directories
- `.env` — environment files (should already be there but verify it covers all directories)

Also check that `apps/intelligence/.gitignore` was removed during the Python migration and that the root `.gitignore` is the source of truth.

### Step 2: Remove Files from Git Tracking

Remove the files from the git index without deleting them locally:
- `git rm --cached apps/intelligence/.env`
- `git rm --cached services/api/bin/server`
- `git rm --cached services/api/server.exe`

### Step 3: (Optional) Clean Git History

The two binary files add ~16MB to every `git clone`. Consider using BFG Repo Cleaner or `git filter-repo` to remove them from history entirely, reducing clone size for all future contributors. This is a destructive history rewrite, so it should be coordinated with all contributors.

### Step 4: Verify .env.example Exists

Ensure every service that needs environment variables has a `.env.example` file:
- `apps/api/.env.example` — exists (good)
- `apps/intelligence/` — needs one documenting ANTHROPIC_API_KEY and any FastAPI config
- `apps/dapp/frontend/` — needs one (covered by separate issue)
- `services/api/` — may need one (or may be removed per the API consolidation issue)

## Files to Remove

- `apps/intelligence/.env`
- `services/api/bin/server`
- `services/api/server.exe`

## Files to Modify

- `.gitignore` — verify coverage
- `apps/intelligence/` — create `.env.example` if missing

Depo-dev:
## Summary

- **Untrack binaries**: `git rm --cached` on `services/api/bin/server`, `services/api/bin/server.exe`, and `services/api/server.exe` — files remain on disk but are no longer committed
- **Harden `.gitignore`**: add `*.exe` and `bin/` patterns to the root `.gitignore` so rebuilt binaries are never accidentally staged again; add `!apps/intelligence/.env.example` negation to allow the example file through
- **Create `apps/intelligence/.env.example`**: documents `INTELLIGENCE_ANTHROPIC_API_KEY`, `INTELLIGENCE_ANTHROPIC_MODEL`, `INTELLIGENCE_HOST`, `INTELLIGENCE_PORT`, and `INTELLIGENCE_JWT_SECRET` (with a note that the secret must match `AUTH_JWT_SECRET` from the Go API)

## Notes

- `apps/intelligence/.gitignore` was removed during the Python migration — the root `.gitignore` is now the single source of truth for all subdirectories
- `services/api/` may be removed as part of the ongoing Go API consolidation; documented in commit message
- **History cleanup**: The 16 MB binary blobs remain in git history. BFG Repo Cleaner (`bfg --delete-files server.exe`) can purge them; recommend coordinating with all contributors and doing a force-push to the shared remote once everyone is synced

## Test plan

- [ ] `git status` on a clean checkout shows no `.env` or binary files as untracked suggestions
- [ ] `apps/intelligence/.env.example` is visible and committed
- [ ] `services/api/bin/server`, `bin/server.exe`, `server.exe` are absent from `git ls-files` output
- [ ] Adding a new `.exe` file under any directory does not appear in `git status`

Closes #184

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,6 +38,11 @@ pnpm-debug.log*
 .env*
 !.env.example
 !apps/api/.env.example
+!apps/intelligence/.env.example
+
+# Compiled binaries and executables
+*.exe
+bin/
 
 # Vercel
 .vercel
--- a/apps/intelligence/.env.example
+++ b/apps/intelligence/.env.example
@@ -1,0 +1,17 @@
+# Nester Intelligence Service — environment variables
+# Copy this file to .env and fill in real values. Never commit .env.
+
+# Anthropic API key — required for all AI endpoints
+# Generate at https://console.anthropic.com/settings/api-keys
+INTELLIGENCE_ANTHROPIC_API_KEY=sk-ant-...
+
+# Claude model to use (defaults to claude-sonnet-4-6)
+INTELLIGENCE_ANTHROPIC_MODEL=claude-sonnet-4-6
+
+# FastAPI server host and port
+INTELLIGENCE_HOST=0.0.0.0
+INTELLIGENCE_PORT=8000
+
+# JWT secret — must match AUTH_JWT_SECRET from the Go API (apps/api)
+# Used to verify Bearer tokens on all protected endpoints
+INTELLIGENCE_JWT_SECRET=your-shared-jwt-secret-here


0xDeon (approved):
Great work

PR #290, opened 2026-04-25
fix(ci): add -race flag to Go API test step

### LINKED ISSUE #260: [Infrastructure] Go API tests run without -race flag — concurrent data races may go undetected in production
## Summary
The Go API test step in `.github/workflows/ci.yml` runs `go test ./...` without the `-race` flag. The Go API uses multiple in-memory data structures accessed concurrently from many goroutines (one goroutine per HTTP request): the authentication challenge store (`map[string]Challenge`), the IP rate limiter bucket map (`map[string]*tokenBucket`), and the wallet rate limiter bucket map. Data races in these structures can cause silent memory corruption, incorrect authentication decisions, rate limit bypass (where the internal counter is corrupted and the limit is never triggered), or in the worst case, a crash from a concurrent map read/write panic. The Go race detector catches these issues at test time with negligible overhead and is considered a standard tool in production-quality Go development.

## Evidence

In `.github/workflows/ci.yml`:
```yaml
- name: Test
  run: go test ./...
  working-directory: apps/api
```

The `-race` flag is absent. The Go race detector documentation explicitly states it should be used in CI for all server software.

In `apps/api/internal/auth/`:
```go
// challengeStore is accessed from multiple goroutines
var challengeStore = struct {
    sync.RWMutex
    challenges map[string]Challenge
}{challenges: make(map[string]Challenge)}
```

If the `sync.RWMutex` is incorrectly applied (e.g., a write operation uses `RLock` instead of `Lock`), the race detector would catch it during tests. Without `-race`, this type of bug is invisible until it causes a production incident.

## Why This Matters for a Financial API

Go data races in a financial API have the following potential consequences:

| Structure | Race Condition Impact |
|-----------|----------------------|
| Challenge store | Two requests verify the same challenge simultaneously → both succeed (replay attack) |
| Rate limiter buckets | Counter corruption → rate limiting never triggers → unlimited requests slip through |
| JWT secret in memory | Concurrent read during rotation → partial read of the secret → invalid signatures accepted |
| User session cache | Stale user data served to wrong session → information disclosure |

A concurrent map access panic (`fatal error: concurrent map read and map write`) is non-recoverable and crashes the process. In production, this manifests as unpredictable service restarts under load, which are very difficult to debug without the race detector's output.

## Impact
1. **Silent bugs in production** — data races may exist in the codebase today without any test failure to indicate them
2. **Authentication integrity** — a race in the challenge store could allow a challenge to be verified twice, bypassing the one-time-use nonce guarantee
3. **Rate limit bypass** — a race in the rate limiter bucket map could corrupt counters, allowing unlimited requests through
4. **Crash under load** — concurrent map access without proper locking causes a panic that crashes the entire API process, creating downtime

### Step 1 — Add -race to CI test command
```yaml
# .github/workflows/ci.yml
- name: Test (with race detector)
  run: go test -race -timeout 120s -count=1 ./...
  working-directory: apps/api
```

The `-count=1` flag disables test result caching, ensuring tests always re-run. The `-timeout 120s` prevents race-induced hangs from blocking CI indefinitely.

### Step 2 — Fix any races detected
Common patterns to audit:
```go
// WRONG — using RLock for a write operation
func (s *store) Set(key string, val interface{}) {
    s.mu.RLock()   // ← Bug: should be Lock()
    defer s.mu.RUnlock()
    s.data[key] = val
}

// CORRECT
func (s *store) Set(key string, val interface{}) {
    s.mu.Lock()
    defer s.mu.Unlock()
    s.data[key] = val
}
```

### Step 3 — Run with race detector locally during development
Add to `Makefile`:
```makefile
test:
	go test -race -timeout 120s ./...

test-short:
	go test -race -short -timeout 60s ./...
```

### Step 4 — Add a concurrent request integration test
```go
func TestChallengeStore_ConcurrentAccess(t *testing.T) {
    store := NewChallengeStore()
    var wg sync.WaitGroup
    
    for i := 0; i < 100; i++ {
        wg.Add(1)
        go func(i int) {
            defer wg.Done()
            addr := fmt.Sprintf("G%056d", i)
            store.Set(addr, "challenge")
            store.Get(addr)
            store.Delete(addr)
        }(i)
    }
    
    wg.Wait() // Race detector will flag any unsynchronized access
}
```

## References
- [The Go Blog — Introducing the Go Race Detector](https://go.dev/blog/race-detector)
- [Go Data Race Detector documentation](https://go.dev/doc/articles/race_detector)
- [Google production Go race policy](https://google.github.io/styleguide/go/best-practices#concurrency) — recommends -race in all CI

Depo-dev:
## Summary

- **CI**: changes `go test ./...` to `go test -race -timeout 120s -count=1 ./...` in the `apps/api` job — catches data races on every PR
- **Makefile**: adds `go-test` and `go-test-short` targets with the same flags so developers can reproduce CI locally (`make go-test`)
- **Concurrent test**: adds `TestChallengeStore_ConcurrentAccess` to `internal/service/challenge_store_test.go` — 100 goroutines each call `Set` → `GetAndDelete` → `GetAndDelete` simultaneously to prove the `InMemoryChallengeStore` mutex is correct; run passes cleanly under `-race`

## Test plan

- [ ] CI `apps-api` job passes with `-race` flag
- [ ] `TestChallengeStore_ConcurrentAccess` passes (`go test -race ./internal/service/...` on Linux)
- [ ] `make go-test` and `make go-test-short` are available and run correctly
- [ ] No data-race warnings reported

Closes #260

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -90,7 +90,7 @@ jobs:
           cache: true
 
       - name: Test
-        run: go test ./...
+        run: go test -race -timeout 120s -count=1 ./...
 
   internal-stellar:
     name: Internal (Go)
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: fmt fmt-check clippy build test integration-test clean dev dev-down dev-reset dev-logs dev-db
+.PHONY: fmt fmt-check clippy build test test-short integration-test clean dev dev-down dev-reset dev-logs dev-db go-test go-test-short
 
 CARGO := cargo
 CONTRACTS_DIR := packages/contracts
@@ -21,6 +21,12 @@ test:
 
 integration-test:
 	cd $(CONTRACTS_DIR) && $(CARGO) test --all --lib
+
+go-test:
+	cd apps/api && go test -race -timeout 120s ./...
+
+go-test-short:
+	cd apps/api && go test -race -short -timeout 60s ./...
 
 clean:
 	cd $(CONTRACTS_DIR) && $(CARGO) clean
--- a/apps/api/internal/service/challenge_store_test.go
+++ b/apps/api/internal/service/challenge_store_test.go
@@ -2,6 +2,8 @@ package service
 
 import (
 	"context"
+	"fmt"
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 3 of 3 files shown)


0xDeon (approved):
Lgtm

PR #294, opened 2026-04-25
fix: resolve security, contract, and chain-integration issues

### LINKED ISSUE #99: feat(backend): wire Stellar contract interactions into vault service
## Overview
\`internal/stellar/\` has a fully designed Stellar/Soroban integration library (client, contract invocation, vault reader, event polling). However, \`apps/api/internal/service/vault_service.go\` does not use it — all vault operations are pure DB writes with no on-chain interaction.

### 1. Deposit Handler → Contract Invocation
\`\`\`go
func (s *VaultService) Deposit(ctx context.Context, req DepositRequest) error {
    // 1. Validate request
    // 2. Call stellar.Client.InvokeContract("vault", "deposit", [from, amount])
    // 3. Wait for confirmation (or use optimistic update + event listener)
    // 4. On success: persist deposit record in DB
    // 5. Emit domain event
}
\`\`\`

### 2. Withdrawal Handler → Contract Invocation
Similar to deposit — invoke \`withdraw\` on vault contract, confirm, update DB.

### 3. Balance Queries from Contract
Replace mock/DB-only balance reads with \`stellar.VaultReader.GetBalance(vaultID)\` calls.

### 4. Event Listener Sync
Start the \`stellar.EventWatcher\` on boot to sync on-chain events into the DB for the activity feed.
\`\`\`go
watcher := stellar.NewEventWatcher(client, contractID)
watcher.Subscribe(func(event stellar.Event) {
    s.handleContractEvent(ctx, event)
})
go watcher.Start(ctx)
\`\`\`

### 5. Transaction Hash Storage
Store the Stellar transaction hash with each deposit/withdrawal record.

## Configuration
Add to API config:
\`\`\`env
STELLAR_NETWORK=testnet
STELLAR_RPC_URL=https://soroban-testnet.stellar.org
STELLAR_SOURCE_KEY=SXXXX
VAULT_CONTRACT_ID=CXXXX
\`\`\`

## Acceptance Criteria
- [ ] Deposit creates on-chain transaction and records hash in DB
- [ ] Balance endpoint returns live on-chain balance
- [ ] Event watcher populates activity table from contract events
- [ ] Contract errors mapped to appropriate HTTP error responses
- [ ] Integration tests run against testnet contracts

## Files to Touch
- \`apps/api/internal/service/vault_service.go\`
- \`apps/api/internal/config/config.go\`
- \`apps/api/cmd/api/main.go\` — inject stellar client
- \`internal/stellar/\` — any missing pieces

### LINKED ISSUE #256: [Frontend] allowHttp: true in Soroban RPC client — permits unencrypted RPC connections in production builds
## Summary
`apps/dapp/frontend/lib/stellar/transaction.ts` instantiates the Soroban RPC client with `allowHttp: true`, a development-only flag that permits the SDK to connect to Soroban RPC endpoints over unencrypted HTTP. In production, all Soroban RPC communication should occur exclusively over HTTPS/TLS. When `allowHttp: true` is enabled and an HTTP RPC URL is used (whether by misconfiguration or by an attacker manipulating the URL via XSS), all RPC traffic — including contract state queries, transaction submissions, and balance lookups — travels in plaintext over the network. A network-level attacker (on the same WiFi, a malicious ISP, or a BGP hijacker) can intercept and modify RPC responses, feeding the frontend false contract state: fabricated balances, fake exchange rates, manipulated vault parameters, or responses that indicate a transaction succeeded when it was dropped.

## Evidence

In `apps/dapp/frontend/lib/stellar/transaction.ts`:
```typescript
const server = new SorobanRpc.Server(
    process.env.NEXT_PUBLIC_STELLAR_RPC_URL!,
    { allowHttp: true }  // ← Development flag left enabled in all environments
);
```

This is unconditional — `allowHttp: true` is set regardless of `NODE_ENV`, regardless of whether the RPC URL is HTTP or HTTPS, and regardless of whether the build is for development or production.

## What Can Be Manipulated via MITM on an HTTP RPC Connection

| RPC Method | What an Attacker Can Fake |
|------------|--------------------------|
| `getAccount` | False account balance, wrong sequence number causing tx failures |
| `simulateTransaction` | Report a transaction as valid when it would fail on-chain |
| `sendTransaction` | Accept a tx submission, return a fake hash, and drop the transaction |
| `getContractData` | Return false vault TVL, fake APY, manipulated exchange rate |
| `getEvents` | Inject fake deposit/withdrawal events into the indexer |

## Threat Model

The risk is highest for Nester's target users in emerging markets:
1. **Mobile data networks** — in many African markets, mobile carriers may use transparent HTTP proxies or deep packet inspection that can modify HTTP responses
2. **Public WiFi** — users accessing the DApp from internet cafes or public WiFi are on networks where MITM is trivial
3. **ISP-level interference** — in some jurisdictions, government-mandated traffic inspection infrastructure has been used to inject content into HTTP connections

## Impact
1. **False balance display** — an attacker can make the UI show a higher balance than exists on-chain, tricking the user into believing a deposit succeeded or that they have more funds available
2. **Transaction manipulation** — a fake "transaction successful" response causes the UI to show a confirmation screen while the actual transaction was never submitted
3. **Exchange rate manipulation** — a false exchange rate response can make a withdrawal appear to return more assets than it actually will, leading to user disappointment and support issues
4. **Sequence number poisoning** — returning wrong account sequence numbers causes legitimate subsequent transactions to fail with confusing errors

### Restrict allowHttp to development builds only
```typescript
// apps/dapp/frontend/lib/stellar/transaction.ts

const rpcUrl = process.env.NEXT_PUBLIC_STELLAR_RPC_URL!;

// Validate that production builds always use HTTPS
if (process.env.NODE_ENV === 'production' && !rpcUrl.startsWith('https://')) {
    throw new Error(
        `Production Soroban RPC URL must use HTTPS. Got: ${rpcUrl}. ` +
        `Set NEXT_PUBLIC_STELLAR_RPC_URL to a valid https:// endpoint.`
    );
}

const server = new SorobanRpc.Server(rpcUrl, {
    allowHttp: process.env.NODE_ENV !== 'production'
});
```

### Add a build-time environment variable validation
```typescript
// apps/dapp/frontend/lib/env-validation.ts
if (typeof window !== 'undefined' && process.env.NODE_ENV === 'production') {
    const rpcUrl = process.env.NEXT_PUBLIC_STELLAR_RPC_URL ?? '';
    if (!rpcUrl.startsWith('https://')) {
        console.error('SECURITY: Soroban RPC URL is not HTTPS in production build');
        // Optionally: disable all contract interactions until fixed
    }
}
```

### Add to CI — reject production builds with HTTP RPC URLs
```yaml
- name: Validate environment config
  run: |
    if grep -r "allowHttp: true" apps/dapp/frontend/lib/; then
      echo "ERROR: allowHttp: true found outside of development guard"
      exit 1
    fi
```

## References
- [Stellar JS SDK — Server options](https://stellar.github.io/js-stellar-sdk/SorobanRpc.Server.html)
- [OWASP — Insecure Transport](https://owasp.org/www-community/vulnerabilities/Insecure_Transport)
- [NIST SP 800-52 Rev 2 — TLS guidance](https://csrc.nist.gov/publications/detail/sp/800-52/rev-2/final)

### LINKED ISSUE #262: [Backend] APY relayer constructs external URL with unsanitized protocolID — path traversal and SSRF risk
## Summary
`internal/stellar/apy_relayer.go` — `ProtocolRPCClient.FetchProtocolAPY` constructs an HTTP URL by concatenating a base URL with a `protocolID` string sourced from the on-chain yield registry:

```go
url := fmt.Sprintf("%s/v1/protocols/%s/apy", c.baseURL, protocolID)
```

The `protocolID` is not URL-encoded or path-sanitized. If an attacker gains operator access and registers a protocol with a crafted `protocolID` containing path traversal sequences (e.g., `../../admin/config`), the constructed URL may resolve to an unintended path on the target server. More critically, if the `baseURL` itself can be influenced (e.g., through a compromised operator key or a misconfigured environment variable), the relayer can be directed to make HTTP GET requests to arbitrary internal network addresses — constituting a Server-Side Request Forgery (SSRF) vulnerability that allows the relayer process to probe and interact with internal services.

## Evidence

In `internal/stellar/apy_relayer.go`:
```go
type ProtocolRPCClient struct {
    baseURL    string
    httpClient *http.Client
}

func (c *ProtocolRPCClient) FetchProtocolAPY(protocolID string) (float64, error) {
    // protocolID is unsanitized — comes from on-chain yield_registry
    url := fmt.Sprintf("%s/v1/protocols/%s/apy", c.baseURL, protocolID)
    
    resp, err := c.httpClient.Get(url)
    // ...
}
```

### Scenario A — Path traversal via malicious protocolID
Attacker (with operator role) registers a protocol with:
```
protocolID = "../../admin/internal-config"
```
Resulting URL: `https://protocol-api.example.com/v1/protocols/../../admin/internal-config/apy`

Some HTTP servers normalize this to `https://protocol-api.example.com/admin/internal-config/apy`, exposing an unintended endpoint. If the target server returns a valid JSON response, the relayer parses it as an APY value.

### Scenario B — SSRF via compromised baseURL
If the `baseURL` is sourced from a config value that can be modified (compromised environment, misconfigured secrets manager), an attacker sets:
```
PROTOCOL_RPC_BASE_URL=http://169.254.169.254  # AWS Instance Metadata Service
```
The relayer then makes requests to the cloud provider's metadata service, potentially exposing IAM credentials, instance identity documents, or other sensitive metadata.

### Scenario C — Internal network probing
With `baseURL=http://internal-postgres-host:5432`, the relayer attempts an HTTP connection to the database port. While Postgres won't respond with valid HTTP, the success/failure of the connection probes the internal network topology.

## Impact
1. **Internal service exposure** — SSRF allows the relayer process to act as a proxy for accessing internal services that are not directly reachable from the internet
2. **Cloud metadata theft** — on cloud deployments (AWS, GCP, Azure), SSRF to the metadata service can yield temporary IAM credentials that grant access to other cloud resources
3. **Path traversal** — unintended API endpoints may be accessed, potentially returning sensitive server configuration or triggering unintended server-side actions
4. **Indirect APY manipulation** — a crafted protocolID can cause the relayer to fetch from an attacker-controlled endpoint and return a crafted APY value (see also the APY oracle bounds issue)

### Step 1 — URL-encode the protocolID path segment
```go
import "net/url"

func (c *ProtocolRPCClient) FetchProtocolAPY(protocolID string) (float64, error) {
    // Encode protocolID to prevent path traversal
    encodedID := url.PathEscape(protocolID)
    
    apiURL := fmt.Sprintf("%s/v1/protocols/%s/apy", c.baseURL, encodedID)
    // ...
}
```

### Step 2 — Validate protocolID format before use
```go
var validProtocolID = regexp.MustCompile(`^[a-zA-Z0-9_-]{1,64}$`)

func (c *ProtocolRPCClient) FetchProtocolAPY(protocolID string) (float64, error) {
    if !validProtocolID.MatchString(protocolID) {
        return 0, fmt.Errorf("invalid protocolID format: %q", protocolID)
    }
    // ...
}
```

### Step 3 — Validate baseURL against an allowlist at initialization time
```go
func NewProtocolRPCClient(baseURL string) (*ProtocolRPCClient, error) {
    parsed, err := url.Parse(baseURL)
    if err != nil {
        return nil, fmt.Errorf("invalid base URL: %w", err)
    }
    
    // Must use HTTPS in production
    if parsed.Scheme != "https" {
        return nil, fmt.Errorf("protocol RPC base URL must use HTTPS, got: %s", parsed.Scheme)
    }
    
    // Must match an allowlist of known protocol API hosts
    allowedHosts := []string{"api.protocol-a.finance", "api.protocol-b.io"}
    allowed := false
    for _, h := range allowedHosts {
        if parsed.Hostname() == h { allowed = true; break }
    }
    if !allowed {
        return nil, fmt.Errorf("protocol RPC host %s is not in the allowlist", parsed.Hostname())
    }
    
    return &ProtocolRPCClient{baseURL: baseURL}, nil
}
```

### Step 4 — Add SSRF prevention via HTTP client restrictions
```go
// Restrict the HTTP client to only make external requests (block private IP ranges)
transport := &http.Transport{
    DialContext: func(ctx context.Context, network, addr string) (net.Conn, error) {
        host, _, _ := net.SplitHostPort(addr)
        ip := net.ParseIP(host)
        if ip != nil && isPrivateIP(ip) {
            return nil, fmt.Errorf("SSRF protection: blocked request to private IP %s", ip)
        }
        return (&net.Dialer{}).DialContext(ctx, network, addr)
    },
}
```

## References
- CWE-918: Server-Side Request Forgery (SSRF)
- CWE-22: Improper Limitation of a Pathname to a Restricted Directory (Path Traversal)
- [OWASP SSRF Prevention Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/Server_Side_Request_Forgery_Prevention_Cheat_Sheet.html)
- [AWS Security: IMDS and SSRF](https://aws.amazon.com/blogs/security/defense-in-depth-open-firewalls-reverse-proxies-ssrf-vulnerabilities-aws-security/)

### LINKED ISSUE #263: [Vault] Emergency withdrawal queue may over-commit liquid reserves when collect_fees runs between queue entry and processing
## Summary
The vault contract maintains an internal `VaultLiquidReserves` counter used to determine how much of the emergency withdrawal queue can be satisfied immediately from available liquid funds. Both `collect_fees()` and the emergency withdrawal queue processor read and write `VaultLiquidReserves`. Because Soroban is single-threaded per-ledger, these operations cannot run simultaneously within a single ledger. However, an emergency withdrawal queue item can span multiple ledgers between when it is added to the queue and when it is processed. During that multi-ledger window, `collect_fees()` may reduce `liquid_reserves`, causing the queue to contain a commitment that exceeds the actual available liquid balance. When the operator attempts to process the queue, the contract may attempt to transfer more tokens than are available in the liquid reserve, leading to either a panic (if the transfer fails) or an accounting discrepancy (if the transfer draws from reserves that were supposed to be non-liquid).

## Timeline of the Issue

```
Ledger N:     Operator checks VaultLiquidReserves = 10,000 USDC
              User adds emergency withdrawal for 9,000 USDC to queue
              Queue approved: 9,000 ≤ 10,000 ✓

Ledger N+1:   Admin calls collect_fees()
              Fee = 500 USDC
              VaultLiquidReserves -= 500 → VaultLiquidReserves = 9,500 USDC

              (Separately, another user makes a normal withdrawal for 1,000 USDC)
              VaultLiquidReserves -= 1,000 → VaultLiquidReserves = 8,500 USDC

Ledger N+2:   Operator processes emergency withdrawal queue
              Tries to transfer 9,000 USDC from liquid reserves
              Actual liquid reserves = 8,500 USDC
              → Either: panic (contract traps) trying to transfer more than available
              → Or: transfer succeeds by drawing 500 USDC from non-liquid reserves
                   (funds that are deployed in yield-generating positions)
                   → This would require liquidating yield positions, which is not automated
```

## Impact
1. **Emergency withdrawal failure** — users expecting immediate liquidity during a crisis find their emergency withdrawal reverts, defeating the purpose of the emergency mechanism
2. **Over-commitment** — liquid reserves are committed to the queue but remain modifiable by fee collection and normal withdrawals, creating an unfunded obligation
3. **Accounting divergence** — if the transfer silently succeeds by drawing non-liquid reserves, the protocol's internal accounting diverges from on-chain reality
4. **Compounded during a bank run** — this issue is most likely to manifest precisely when it matters most: during a bank run scenario where normal withdrawals, emergency withdrawals, and fee collection may all be happening in rapid succession

### Implement a reservation mechanism for the emergency withdrawal queue

When an item is added to the emergency queue, immediately reserve (lock) the corresponding liquid reserves so they cannot be used for fee collection or other purposes:

```rust
// DataKey additions
enum DataKey {
    // ... existing keys
    LiquidReserved,   // Amount of liquid_reserves committed to the queue
}

// When adding to emergency queue
pub fn request_emergency_withdrawal(env: Env, from: Address, shares: i128) {
    from.require_auth();
    
    let assets_needed = calculate_assets(&env, shares);
    
    let liquid: i128 = env.storage().persistent()
        .get(&DataKey::LiquidReserves).unwrap_or(0);
    let reserved: i128 = env.storage().persistent()
        .get(&DataKey::LiquidReserved).unwrap_or(0);
    
    let available = liquid - reserved;
    
    if available < assets_needed {
        // Insufficient unreserved liquidity — add to pending queue, not guaranteed queue
        add_to_pending_queue(&env, from, shares, assets_needed);
        return;
    }
    
    // Reserve the liquidity
    env.storage().persistent().set(
        &DataKey::LiquidReserved,
        &(reserved + assets_needed)
    );
    
    // Add to guaranteed queue
    add_to_guaranteed_queue(&env, from, shares, assets_needed);
}

// When collect_fees runs, only use unreserved liquid reserves
pub fn collect_fees(env: Env, caller: Address) {
    // ...
    let liquid: i128 = get_liquid_reserves(&env);
    let reserved: i128 = get_liquid_reserved(&env);
    let available_for_fees = (liquid - reserved).max(0);
    
    let fee_amount = fee_amount.min(available_for_fees);
    // Only collect fees from unreserved liquid reserves
}

// When queue item is processed or cancelled, release the reservation
fn release_reservation(env: &Env, amount: i128) {
    let reserved: i128 = env.storage().persistent()
        .get(&DataKey::LiquidReserved).unwrap_or(0);
    env.storage().persistent().set(
        &DataKey::LiquidReserved,
        &(reserved - amount).max(0)
    );
}
```

### Write tests covering the multi-ledger scenario
```rust
#[test]
fn test_queue_reservation_prevents_overcommit() {
    // Add emergency withdrawal for 9,000
    // Call collect_fees() attempting to use 5,000
    // Verify collect_fees only uses available unreserved amount
    // Process emergency withdrawal — verify it succeeds
}
```

## References
- Reservation pattern for distributed systems — Martin Fowler: Patterns of Enterprise Application Architecture
- [Compound Finance: liquidation accounting invariants](https://compound.finance/docs)
- Soroban: multi-ledger state consistency guarantees

gboigwe:
## Summary

- Enforce HTTPS for Soroban RPC client in production; `allowHttp` is now only set in non-production environments to prevent plaintext RPC calls on mainnet
- Sanitize `protocolID` path segment with a regex allowlist and apply `url.PathEscape` before URL construction; add SSRF protection to `ProtocolRPCClient` via a custom `http.Transport` that blocks requests to private/reserved IP ranges
- Wire Soroban contract interactions into the vault service: introduce `VaultDepositInvoker` interface backed by `SorobanVaultChainInvoker`; `RecordDeposit` and `RecordWithdrawal` now invoke the on-chain vault contract before persisting to the database when an operator key is configured
- Fix emergency withdrawal queue over-commit in the vault contract: add `LiquidReserved` storage to track funds committed to the queue; `collect_fees` is capped to unreserved liquid, and `process_emergency_queue` releases reservations as requests are fulfilled

## Test plan

- [ ] Verify `allowHttp` is `false` when `NODE_ENV=production` in transaction.ts
- [ ] Confirm `ProtocolRPCClient` rejects non-HTTPS base URLs and blocks private-IP targets
- [ ] Start API with `STELLAR_OPERATOR_SECRET` set; confirm `SetDepositInvoker` is wired and deposit/withdraw flow reaches `InvokeWithI128Pair`
- [ ] Without operator secret, confirm no-op path leaves existing DB-only behaviour unchanged
- [ ] Deploy updated vault contract to testnet; queue an emergency withdrawal that exceeds liquid reserves and confirm `collect_fees` cannot drain the reserved amount
- [ ] Confirm `process_emergency_queue` pays out queued requests and decrements `LiquidReserved` correctly

Closes #256
Closes #262
Closes #99
Closes #263

### DIFF
--- a/apps/api/cmd/api/main.go
+++ b/apps/api/cmd/api/main.go
@@ -78,6 +78,7 @@ func run() error {
 			return fmt.Errorf("init chain invoker: %w", err)
 		}
 		chainInvoker = inv
+		vaultService.SetDepositInvoker(inv)
 	}
 
 	adminService := service.NewAdminService(
--- a/apps/api/internal/service/admin_service.go
+++ b/apps/api/internal/service/admin_service.go
@@ -5,7 +5,6 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
-	"syscall"
 	"time"
 
 	"github.com/google/uuid"
@@ -280,20 +279,3 @@ func (s *AdminService) checkEventIndexer(ctx context.Context) admindomain.Health
 	}
 }
 
-func diskUsage() string {
-	var stat syscall.Statfs_t
-	if err := syscall.Statfs("/", &stat); err != nil {
-		return "n/a"
-	}
-
-	total := stat.Blocks * uint64(stat.Bsize)
-	free := stat.Bavail * uint64(stat.Bsize)
-	used := total - free
-
-	if total == 0 {
-		return "n/a"
-	}
-
-	usedPct := (float64(used) / float64(total)) * 100
-	return fmt.Sprintf("%.1f%%", usedPct)
-}
--- a/apps/api/internal/service/disk_unix.go
+++ b/apps/api/internal/service/disk_unix.go
@@ -1,0 +1,22 @@
+//go:build !windows
+
+package service
+
+import (
+	"fmt"
+	"syscall"
+)
+
+func diskUsage() string {
+	var stat syscall.Statfs_t
+	if err := syscall.Statfs("/", &stat); err != nil {
+		return "n/a"
+	}
+	total := stat.Blocks * uint64(stat.Bsize)
+	free := stat.Bavail * uint64(stat.Bsize)
+	used := total - free
+	if total == 0 {
+		return "n/a"
+	}
+	return fmt.Sprintf("%.1f%%", (float64(used)/float64(total))*100)
+}
--- a/apps/api/internal/service/disk_windows.go
+++ b/apps/api/internal/service/disk_windows.go
@@ -1,0 +1,7 @@
+//go:build windows
+
+package service
+
+func diskUsage() string {
+	return "n/a"
+}
--- a/apps/api/internal/service/soroban_vault_chain_invoker.go
+++ b/apps/api/internal/service/soroban_vault_chain_invoker.go
@@ -29,3 +29,17 @@ func (s *SorobanVaultChainInvoker) PauseVault(ctx context.Context, contractAddre
 func (s *SorobanVaultChainInvoker) UnpauseVault(ctx context.Context, contractAddress string) error {
 	return s.invoker.InvokeVoidFunction(ctx, contractAddress, "unpause")
 }
+
+// DepositToVault invokes the vault contract's deposit function with the
+// operator as both caller and depositing user, passing amount and zero
+// as the minimum-shares-out slippage guard.
+func (s *SorobanVaultChainInvoker) DepositToVault(ctx context.Context, contractAddress string, amountStroops int64) error {
+	return s.invoker.InvokeWithI128Pair(ctx, contractAddress, "deposit", amountStroops, 0)
+}
+
+// WithdrawFromVault invokes the vault contract's withdraw function with the
+// operator as both caller and withdrawing user, passing shares and zero
+// as the minimum-assets-out slippage guard.
+func (s *SorobanVaultChainInvoker) WithdrawFromVault(ctx context.Context, contractAddress string, sharesStroops int64) error {
+	return s.invoker.InvokeWithI128Pair(ctx, contractAddress, "withdraw", sharesStroops, 0)
+}
--- a/apps/api/internal/service/vault_service.go
+++ b/apps/api/internal/service/vault_service.go
@@ -2,6 +2,7 @@ package service
 
 import (
 	"context"
+	"fmt"
 	"strings"
 	"time"
 
@@ -11,8 +12,24 @@ import (
 	"github.com/suncrestlabs/nester/apps/api/internal/domain/vault"
 )
 
+// VaultDepositInvoker handles on-chain deposit and withdrawal operations.
+// Implementations invoke the Soroban vault contract; the noop is used when
+// no operator secret is configured.
+type VaultDepositInvoker interface {
+	DepositToVault(ctx context.Context, contractAddress string, amountStroops int64) error
+	WithdrawFromVault(ctx context.Context, contractAddress string, sharesStroops int64) error
+}
+
+// NoopVaultDepositInvoker satisfies VaultDepositInvoker without making any
+// on-chain calls. Used when chain integration is not configured.
+type NoopVaultDepositInvoker struct{}
+
+func (NoopVaultDepositInvoker) DepositToVault(_ context.Context, _ string, _ int64) error    { return nil }
+func (NoopVaultDepositInvoker) WithdrawFromVault(_ context.Context, _ string, _ int64) error { return nil }
+
 type VaultService struct {
-	repository vault.Repository
+	repository     vault.Repository
+	depositInvoker VaultDepositInvoker
 }
 
 // ── Input types ──────────────────────────────────────────────────────────────
@@ -56,6 +73,12 @@ type RecordWithdrawalInput struct {
 
 func NewVaultService(repository vault.Repository) *VaultService {
 	return &VaultService{repository: repository}
+}
+
+// SetDepositInvoker wires an optional on-chain invoker into the vault service.
+// Call this after NewVaultService when an operator key is available.
+func (s *VaultService) SetDepositInvoker(invoker VaultDepositInvoker) {
+	s.depositInvoker = invoker
 }
 
 // ── Existing methods ─────────────────────────────────────────────────────────
@@ -150,6 +173,17 @@ func (s *VaultService) RecordDeposit(ctx context.Context, input RecordDepositInp
 	}
 	if decimalScale(input.Amount) > vault.MaxAmountScale {
 		return vault.Vault{}, vault.ErrInvalidPrecision
+	}
+
+	if s.depositInvoker != nil {
+		existing, err := s.repository.GetVault(ctx, input.VaultID)
+		if err != nil {
+			return vault.Vault{}, err
+		}
+		stroops := input.Amount.Mul(decimal.NewFromInt(10_000_000)).IntPart()
+		if err := s.depositInvoker.DepositToVault(ctx, existing.ContractAddress, stroops); err != nil {
+			return vault.Vault{}, fmt.Errorf("on-chain deposit failed: %w", err)
+		}
 	}
 
 	if err := s.repository.RecordDeposit(ctx, input.VaultID, input.Amount); err != nil {
@@ -338,6 +372,13 @@ func (s *VaultService) RecordWithdrawal(ctx context.Context, input RecordWithdra
 	}
 	if existing.CurrentBalance.LessThan(input.Amount) {
 		return vault.Vault{}, vault.ErrInsufficientBalance
+	}
+
+	if s.depositInvoker != nil {
+		stroops := input.Amount.Mul(decimal.NewFromInt(10_000_000)).IntPart()
+		if err := s.depositInvoker.WithdrawFromVault(ctx, existing.ContractAddress, stroops); err != nil {
+			return vault.Vault{}, fmt.Errorf("on-chain withdrawal failed: %w", err)
+		}
 	}
 
 	if err := s.repository.RecordWithdrawal(ctx, input.VaultID, input.Amount); err != nil {
--- a/apps/api/internal/stellar/invoker.go
+++ b/apps/api/internal/stellar/invoker.go
@@ -300,6 +300,123 @@ func (c *ContractInvoker) getSequenceNumber(ctx context.Context) (int64, error) 
 	return seq, nil
 }
 
+// InvokeWithI128Pair calls a contract function with signature
+// (caller: Address, arg0: i128, arg1: i128). Suitable for deposit and withdraw
+// where the operator acts as the transaction source and user.
+func (c *ContractInvoker) InvokeWithI128Pair(ctx context.Context, contractAddress, functionName string, arg0, arg1 int64) error {
+	contractScAddr, err := contractAddressToXDR(contractAddress)
+	if err != nil {
+		return err
+	}
+
+	callerScAddr, err := accountAddressToXDR(c.kp.Address())
+	if err != nil {
+		return err
+	}
+
+	hostFn := xdr.HostFunction{
+		Type: xdr.HostFunctionTypeHostFunctionTypeInvokeContract,
+		InvokeContract: &xdr.InvokeContractArgs{
+			ContractAddress: contractScAddr,
+			FunctionName:    xdr.ScSymbol(functionName),
+			Args: []xdr.ScVal{
+				{Type: xdr.ScValTypeScvAddress, Address: &callerScAddr},
+				int64ToI128ScVal(arg0),
+				int64ToI128ScVal(arg1),
+			},
+		},
+	}
+
+	seq, err := c.getSequenceNumber(ctx)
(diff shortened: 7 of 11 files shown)


0xDeon (approved):
Lgtm. Thanks for your contributions

PR #296, opened 2026-04-25
Secure WebSocket Authentication with JWT

### LINKED ISSUE #243: [Frontend] WebSocket authentication uses a hardcoded fake JWT derived from wallet address — no real auth
## Summary
`apps/dapp/frontend/components/websocket-provider.tsx` constructs a WebSocket authentication token as the string literal `mock_jwt_${address}` — where `address` is the connected wallet's public key. This is not a JWT, not cryptographically signed, not verified by any server, and provides zero authentication. The inline comment in the source code explicitly acknowledges this is a placeholder: `// Derive a simple JWT placeholder from the wallet address. Replace with a real auth token once the backend is ready.` This placeholder has shipped to the development branch without being replaced. Because Stellar wallet addresses are public (visible on-chain, in transaction histories, and derivable from public keys), any third party who knows a user's wallet address can fabricate an identical "token" and authenticate to the WebSocket endpoint as that user.

## Evidence

In `apps/dapp/frontend/components/websocket-provider.tsx` lines 78–79:
```typescript
// Derive a simple JWT placeholder from the wallet address.
// Replace with a real auth token once the backend is ready.
const mockToken = `mock_jwt_${address}`;
const ws = new WebSocket(`${wsUrl}?token=${mockToken}`);
```

The WebSocket server accepts any connection with a token matching this pattern without cryptographic verification.

## What "Authentication" This Bypasses
The WebSocket channel is used for real-time protocol updates: vault balance changes, settlement status notifications, AI chat streaming, and market data feeds. With fake auth:
1. Any caller who knows a user's wallet address can connect as that user and receive their real-time financial data stream
2. Any caller can inject messages into the WebSocket stream that the frontend will process as if they came from the server
3. There is no way to revoke a "session" because there is no real session — the "token" is static and derived deterministically from the wallet address

## Impact
1. **Identity impersonation** — wallet addresses are public; constructing `mock_jwt_G...` for any user is trivial
2. **Real-time data exposure** — portfolio updates, settlement notifications, and AI responses streamed over the WebSocket are accessible to anyone who knows the target's wallet address
3. **Message injection** — if the WebSocket server trusts the authenticated identity of incoming messages, an attacker can inject false state updates (e.g., fake "settlement completed" events) that the frontend renders as real
4. **Scalability** — since the "token" is deterministic, an attacker can pre-compute tokens for thousands of known wallet addresses scraped from the Stellar ledger

## Root Cause
The WebSocket auth was intentionally left as a stub during early development. The comment acknowledges a real auth token should be used "once the backend is ready" — but the real JWT issuance system now exists in the Go API and this stub was never updated to use it.

### Step 1 — Use the real JWT from the wallet challenge/response auth flow
After the wallet connects and the Go API issues a signed JWT via `POST /api/v1/auth/verify`, store that token in React state (not localStorage):

```typescript
// In wallet auth flow
const { jwt } = await authApi.verifyWalletSignature(walletAddress, signature, challenge);
setAuthToken(jwt); // Store in React state / context, not localStorage
```

### Step 2 — Pass the real JWT to the WebSocket connection
```typescript
// In websocket-provider.tsx
const ws = new WebSocket(`${wsUrl}`);
ws.onopen = () => {
    // Send auth message immediately after connection
    ws.send(JSON.stringify({
        type: 'authenticate',
        token: authToken, // Real signed JWT from the auth flow
    }));
};
```

### Step 3 — Server-side WebSocket auth validation
The WebSocket server must validate the token on the first message before accepting any subsequent messages from the connection:
```go
// In WebSocket handler
func (h *WSHandler) HandleConnection(conn *websocket.Conn) {
    var authMsg AuthMessage
    conn.ReadJSON(&authMsg)
    
    claims, err := auth.ValidateToken(authMsg.Token, h.jwtSecret)
    if err != nil {
        conn.WriteJSON(ErrorMessage{Error: "authentication failed"})
        conn.Close()
        return
    }
    // Now handle messages for claims.Subject
}
```

### Step 4 — Remove all mock_jwt_ references
Search the entire codebase for `mock_jwt` and remove or replace every occurrence. Add a lint rule or grep check in CI to prevent reintroduction.

## References
- RFC 6455 — WebSocket Protocol
- [WebSocket authentication patterns](https://websockets.readthedocs.io/en/stable/topics/authentication.html)
- OWASP API Security — API2: Broken Authentication

Tybravo:
Closes #243 

**Description:**
Fixes a critical security vulnerability by replacing the insecure `mock_jwt_` URL authentication with a secure, post-connection JWT authentication flow. 

**Key Changes:**
* **Removed Mock Auth:** Stripped all predictable fake token generation from the codebase.
* **New `AuthProvider`:** Securely manages the backend-issued JWT in React memory (avoids `localStorage`).
* **Post-Connection Auth:** WebSockets now connect cleanly and authenticate via a `{ type: 'authenticate', token }` payload.
* **Event Guarding:** Added an `isAuthenticated` state; no WebSocket events are processed until `auth_success` is confirmed by the server.
* **Strict Error Handling:** `auth_error` events immediately close the socket, prevent infinite reconnect loops, clear the token, and safely disconnect the user.
* **Security Guards:** Added strict developer comments in the provider to prevent future auth regressions.


0xDeon (dismissed):
Great work!!!

PR #470, opened 2026-05-26
added ping to check db connectivity and simple check to validate wallet address

### LINKED ISSUE #447: bug(api): bootstrap-admin does not call db.Ping() after sql.Open() — connection errors surface as confusing query failures
## Summary

`apps/api/cmd/bootstrap-admin/main.go` uses `sql.Open()` to initialize the database connection but never calls `db.Ping()` to verify the connection is actually reachable. `sql.Open()` only validates the DSN format — it does not establish a connection. An unreachable database or malformed DSN silently passes `sql.Open()` and only fails later when the first query executes, producing a confusing error message rather than a clear "cannot connect to database" diagnostic.

## Affected File

`apps/api/cmd/bootstrap-admin/main.go`

## Current Behaviour

```go
db, err := sql.Open("pgx", dsn)
if err != nil {
    // This only catches DSN parse errors, not connection failures
    log.Fatal(err)
}
// Proceeds to query — if DB is unreachable, error surfaces here with no context
```

## Fix

Add `db.Ping()` immediately after `sql.Open()`:

```go
db, err := sql.Open("pgx", dsn)
if err != nil {
    return fmt.Errorf("invalid database DSN: %w", err)
}

if err := db.Ping(); err != nil {
    return fmt.Errorf("cannot reach database: %w", err)
}
```

## Additional Enhancement (polish)

Validate the Stellar wallet address format before querying the database. A typo in the wallet address currently produces `no user found with wallet address` — not helpful when the address itself is malformed.

```go
// Basic Stellar address format check
if !strings.HasPrefix(*wallet, "G") || len(*wallet) != 56 {
    return fmt.Errorf("invalid Stellar address format: must start with 'G' and be 56 characters")
}
```

## Acceptance Criteria

- [ ] `db.Ping()` is called after `sql.Open()` with a clear wrapped error message on failure
- [ ] Running `bootstrap-admin` against an unreachable database prints `cannot reach database: ...` and exits with a non-zero code immediately
- [ ] (Optional) Stellar address format validated before DB query

## References

- OSS_CLEANUP.md — JWT Roles / Admin RBAC Fix section (PR #270)

deltron-fr:
closes #447 

This change does three things:

 - wraps sql.Open("pgx", *dsn) as invalid database DSN: ...
 - calls db.Ping() immediately after open and returns cannot reach database: ... on failure
 - validates --wallet with the requested basic Stellar format check (G prefix, length 56)

### DIFF
--- a/apps/api/cmd/bootstrap-admin/main.go
+++ b/apps/api/cmd/bootstrap-admin/main.go
@@ -13,6 +13,7 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"strings"
 
 	_ "github.com/jackc/pgx/v5/stdlib"
 	"github.com/joho/godotenv"
@@ -34,6 +35,9 @@ func run() error {
 		flag.Usage()
 		return fmt.Errorf("--wallet is required")
 	}
+	if !strings.HasPrefix(*wallet, "G") || len(*wallet) != 56 {
+		return fmt.Errorf("invalid Stellar address format: must start with 'G' and be 56 characters")
+	}
 
 	_ = godotenv.Load()
 	if *dsn == "" {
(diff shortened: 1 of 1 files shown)


0xDeon (commented):
Great work. Lgtm!

0xDeon (approved):

PR #473, opened 2026-05-26
fix(api): make Stellar USDC issuer configurable via STELLAR_USDC_ISSU…

TEEZY234:
## Summary

This PR fixes issue #455 by making the Stellar USDC issuer configurable via a new `STELLAR_USDC_ISSUER` environment variable.

## What changed

- Added `STELLAR_USDC_ISSUER` to API config with the existing mainnet Circle USDC issuer as the default
- Removed the hardcoded issuer constant from `apps/api/internal/oracle/stellar.go`
- Updated the Stellar oracle provider to accept the issuer from config
- Updated API bootstrap code to pass the configured issuer
- Documented `STELLAR_USDC_ISSUER` in `apps/api/.env.example`
- Added tests to verify default behavior and env override behavior

## Why

The previous implementation always used the mainnet Circle USDC issuer, which breaks testnet and multi-network Stellar support because the Horizon order book query never matches a non-mainnet issuer.

## Testing

- Verified modified files have no editor diagnostics
- `go` toolchain was unavailable in this environment, so tests were not executed here

## Closes

- closes #455

### DIFF
--- a/apps/api/.env.example
+++ b/apps/api/.env.example
@@ -47,6 +47,11 @@ STELLAR_RPC_URL=https://soroban-testnet.stellar.org
 # Horizon API endpoint.
 STELLAR_HORIZON_URL=https://horizon-testnet.stellar.org
 
+# Stellar USDC issuer address used by the oracle order book query.
+# Defaults to Circle's mainnet USDC issuer when not set.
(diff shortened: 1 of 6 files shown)


0xDeon (approved):
Nice work. Lgtm

Issue #373, opened 2026-04-29
feat(api,dapp): Add email and push notification system


0xDeon:
## Background

Users currently have no way to be alerted about important events — settlement completions, yield milestones, vault warnings, or admin actions — without actively checking the dashboard. The notifications page exists in the UI but is empty.

### Backend — Notification Dispatcher (`apps/api/internal/notifications/`)

**Notification types:**
| Event | Channel |
|---|---|
| Settlement completed / failed | Email + WebSocket |
| Deposit confirmed on-chain | Email + WebSocket |
| Vault APY drops >10% in 24 h | Email |
| Vault paused by admin | Email + WebSocket |
| Rebalance executed | WebSocket only |
| KYC approved / rejected | Email |

**Implementation:**
1. `NotificationService` interface with `Send(ctx, userID, event, payload)` method
2. Email provider: support SMTP (via `net/smtp`) and a pluggable interface for SendGrid/Resend
3. WebSocket delivery: push JSON event to hub for connected clients
4. `notifications` DB table to persist notification history (for the notifications page)

```sql
-- Migration 018
CREATE TABLE notifications (
  id UUID PRIMARY KEY DEFAULT gen_random_uuid(),
  user_id UUID NOT NULL REFERENCES users(id),
  type TEXT NOT NULL,
  title TEXT NOT NULL,
  body TEXT NOT NULL,
  read BOOLEAN NOT NULL DEFAULT FALSE,
  sent_at TIMESTAMPTZ NOT NULL DEFAULT NOW()
);
CREATE INDEX notifications_user_id_idx ON notifications(user_id);
```

5. API endpoints:
   ```
   GET  /api/v1/users/{userId}/notifications          # paginated list
   PATCH /api/v1/users/{userId}/notifications/{id}   # mark read
   PATCH /api/v1/users/{userId}/notifications/read-all
   ```

6. User preferences — extend `PATCH /api/v1/users/{id}` to accept `notification_prefs: { email: bool, push: bool }` and respect them before dispatching

### Frontend — Notifications Page

- Fetch from `GET /api/v1/users/{userId}/notifications`
- Render notification list with type icon, title, timestamp, read/unread state
- "Mark all as read" button
- Badge count on the notification bell in the nav bar (driven by unread count from API)
- Subscribe to WebSocket `notification` events and append them in real time

## Acceptance Criteria

- [ ] Settlement completion triggers an email to the user's registered address
[...]

Issue #372, opened 2026-04-29
feat(api): Build automated vault rebalancing scheduler


0xDeon:
## Background

The Soroban vault contract has rebalancing logic (shifting capital between Aave, Blend, Compound based on yield), but nothing triggers it automatically. A vault operator must manually call the contract — which won't scale.

### Rebalancing Scheduler (`apps/api/internal/scheduler/`)

A background goroutine started alongside the API server that:

1. **Polls vault performance** every configurable interval (default: 15 min)
   - Fetch all active vaults from DB
   - For each vault, fetch current protocol APYs via an on-chain query or DefiLlama

2. **Evaluates rebalancing need**
   - Compute the optimal allocation given current yields
   - Compare to the vault's current `vault_allocations` rows
   - Trigger rebalance if the yield improvement exceeds a configurable threshold (e.g., >0.5% APY gain after fees)

3. **Submits rebalance transaction**
   - Call the `rebalance` function on the vault's Soroban contract using the operator keypair
   - Record the transaction hash and update `vault_allocations`
   - Emit an audit log entry

4. **Circuit breaker awareness**
   - Respect the contract's built-in circuit breaker (no rebalance if >20% withdrawal in 2 hr window)
   - Backoff and retry if the contract returns a circuit-breaker error

### Configuration (env vars)

```
REBALANCER_ENABLED=true
REBALANCER_INTERVAL_MINUTES=15
REBALANCER_MIN_APY_GAIN_BPS=50        # 0.5%
OPERATOR_SECRET_KEY=S...              # Stellar operator keypair
```

### Admin Endpoint

```
POST /api/v1/admin/vaults/{id}/rebalance   # Manual trigger for testing
```

## Acceptance Criteria

- [ ] Scheduler starts on API boot when `REBALANCER_ENABLED=true`
- [ ] Rebalance only triggers when expected APY improvement exceeds threshold
- [ ] Contract call is recorded in `transactions` table with `type = "rebalance"`
- [ ] Circuit breaker errors are handled gracefully (logged, not crashed)
- [ ] Manual trigger endpoint works for authorized admins
- [ ] Unit tests for the rebalancing decision logic (no contract calls needed — mock the yield fetcher)

## References

- Vault contract: `apps/dapp/contracts/vault/src/`
- Event indexer (pattern to follow): `apps/api/internal/indexer/`
- Existing admin endpoints: `apps/api/internal/handlers/admin.go`

OkeyAmy:
[...]

PR #477, opened 2026-05-27
test(contracts): zero-performance-fee-on-impairment regression test (#451)

### LINKED ISSUE #451: bug(contracts): missing impairment regression test — zero performance fee invariant unverified
## Summary

The vault contract correctly handles loss scenarios — when `yield_part < 0` (impairment), the performance fee is skipped. However, there is no test that proves this invariant. A future refactor of the fee calculation logic could silently break this behaviour and charge performance fees during a loss event, penalising users who are already experiencing a loss.

## Affected File

`packages/contracts/contracts/vault/src/test.rs`

## What Needs to Be Tested

The issue spec (referenced in PR #275) explicitly required this test:

> "User deposits at rate 1.0, rate halves (impairment) → zero performance fee."

The production code handles it correctly via a guard like:

```rust
if yield_part < 0 {
    // impairment — skip performance fee
    return 0;
}
```

But no test exercises this code path.

## Required Test

Add a test to `test.rs` that:

1. Initialises the vault with a known share price (e.g., `1.0`)
2. Simulates an impairment event that halves the share price (e.g., to `0.5`)
3. Calls the performance fee calculation function
4. Asserts that the performance fee returned is `0`

Example structure:

```rust
#[test]
fn test_no_performance_fee_on_impairment() {
    let env = Env::default();
    // ... setup vault with initial price = 1.0
    // ... simulate loss: price drops to 0.5
    let fee = calculate_performance_fee(&env, /* args */);
    assert_eq!(fee, 0, "performance fee must be zero when vault has experienced a loss");
}
```

## Why This Matters

- Without this test, the zero-fee invariant under impairment is only enforced by reading the code — not by CI
- Any change to `calculate_performance_fee` or related functions could break this silently
- Charging fees during a loss is a critical financial bug that directly harms users

## Acceptance Criteria

- [ ] Test added that simulates a share price drop (impairment scenario)
- [ ] Test asserts `performance_fee == 0` when `yield_part < 0`
- [ ] `cargo test -p vault-contract` passes with the new test
- [ ] Test is co-located with existing vault tests in `test.rs`

## References

- OSS_CLEANUP.md — Performance Fee Fix section (PR #275)
- The production guard exists in `packages/contracts/contracts/vault/src/lib.rs`

chigozirim007:
## Summary

closes #451

Adds `withdrawal_charges_no_perf_fee_on_impairment` to the vault contract tests. It deposits, simulates a 20% impairment via `report_yield(-200)`, withdraws, and asserts the user recovers the full impaired value with **no performance fee** — proving the `if yield_part > 0` guard in `lib.rs` and locking the invariant against future fee-logic refactors.

## Note on the other three issues in this batch
- **#453** (add `AUTH_JWT_SECRET` to the docker-compose API service) is **already present** in `docker-compose.yml` (line 42) — nothing to change.
- **#372** (automated vault rebalancing scheduler) and **#373** (email + push notification system) are **large feature builds** (Go services with DB models, scheduling/dispatch loops, and API/dapp wiring). I can't build/run the Go toolchain in my environment, and shipping two full features blind risks broken code — so they're intentionally **not** in this PR. They warrant a session where the repo can actually be compiled and tested.

### DIFF
--- a/packages/contracts/contracts/vault/src/test.rs
+++ b/packages/contracts/contracts/vault/src/test.rs
@@ -950,3 +950,39 @@ fn process_emergency_queue_decrements_liquid_reserved() {
         "collect_fees should transfer fees once LiquidReserved is decremented after queue processing"
     );
 }
+
+/// Regression test for the zero-performance-fee-on-impairment invariant (#451).
+///
+/// When a vault is impaired (`yield_part < 0`), `withdraw` must NOT charge a
+/// performance fee. The fee guard is `if yield_part > 0` in `lib.rs`; a future
+/// refactor could silently drop it and start charging fees on losses. This test
+/// locks the behaviour in: with the early-withdrawal fee disabled, an impaired
+/// withdrawal returns the full impaired value with no fee skimmed off.
+#[test]
+fn withdrawal_charges_no_perf_fee_on_impairment() {
+    let (env, admin, token, vault, _treasury) = setup();
+    let user = Address::generate(&env);
+    let deposit = 1_000 * XLM;
+    mint(&token, &user, deposit);
+
+    // Isolate performance-fee behaviour by disabling the early-withdrawal fee.
+    let mut fee_config: FeeConfig = vault.get_fee_config();
+    fee_config.early_withdrawal_fee_bps = 0;
+    vault.set_fee_config(&admin, &fee_config);
+
+    vault.grant_role(&admin, &admin, &Role::Manager);
+    vault.deposit(&user, &deposit, &0);
+
+    // Simulate a 20% impairment: total assets drop from 1000 to 800.
+    vault.report_yield(&admin, &(-(200 * XLM)));
+
+    let shares = vault.get_shares(&user);
+    vault.withdraw(&user, &shares, &0);
+
+    // yield_part = 800 - 1000 = -200 (< 0) => no performance fee. The user
+    // recovers the full impaired 800 XLM, not 780 after a wrongful 10% fee.
+    // Soroban arithmetic is deterministic integer math (no FP rounding), so
+    // an exact equality is the right contract — per-review feedback.
+    let balance = token::Client::new(&env, &token.address).balance(&user);
+    assert_eq!(balance, 800 * XLM, "impairment must not charge a performance fee");
+}


packages/contracts/contracts/vault/src/test.rs
```diff
@@ -950,3 +950,41 @@ fn process_emergency_queue_decrements_liquid_reserved() {
...
+    vault.report_yield(&admin, &(-(200 * XLM)));
+
+    let shares = vault.get_shares(&user);
+    vault.withdraw(&user, &shares, &0);
+
+    // yield_part = 800 - 1000 = -200 (< 0) => no performance fee. The user
+    // recovers the full impaired ~800 XLM, not ~780 after a wrongful 10% fee.
+    let balance = token::Client::new(&env, &token.address).balance(&user);
+    assert!(
+        balance >= 800 * XLM - 10 && balance <= 800 * XLM,
```
0xDeon: Fuzzy bounds are unnecessary here. Soroban arithmetic is deterministic integer arithmetic — a 20% impairment on exactly 1000 XLM produces exactly 800 XLM with no rounding. Replace with: `assert_eq!(balance, 800 * XLM, "impairment must not charge a performance fee");`

0xDeon (changes requested):

packages/contracts/contracts/vault/src/test.rs
```diff
@@ -950,3 +950,39 @@ fn process_emergency_queue_decrements_liquid_reserved() {
...
+/// Regression test for the zero-performance-fee-on-impairment invariant (#451).
+///
+/// When a vault is impaired (`yield_part < 0`), `withdraw` must NOT charge a
+/// performance fee. The fee guard is `if yield_part > 0` in `lib.rs`; a future
+/// refactor could silently drop it and start charging fees on losses. This test
+/// locks the behaviour in: with the early-withdrawal fee disabled, an impaired
+/// withdrawal returns the full impaired value with no fee skimmed off.
+#[test]
+fn withdrawal_charges_no_perf_fee_on_impairment() {
+    let (env, admin, token, vault, _treasury) = setup();
```
0xDeon, line 963: This only delivers #451. The branch name (`nester-372-373-451-453`) and the PR's `closes` lines imply all four issues were scoped here, but #372 (rebalance scheduler) and #373 (notification system) are explicitly not done. Also, PR #481 is open and adds an equivalent impairment test for the same issue. Please either: expand the scope to cover the remaining issues, or scope this PR down, rename the branch, and remove the `closes` lines for the unaddressed issues so they stay open.

0xDeon (changes requested):

Issue #17, opened 2026-03-17
Create vault share token contract for proportional ownership tracking


0xDeon:
## Description

When users deposit into a Nester vault, they receive share tokens that represent their proportional ownership of the vault's total assets. This is the standard vault pattern in DeFi (ERC-4626 equivalent)—share tokens decouple a user's ownership from the vault's fluctuating total value, which changes as yield accrues.

For example: if a vault holds 10,000 USDC and has issued 10,000 shares, each share is worth 1 USDC. As yield accrues and the vault grows to 10,500 USDC, each share is now worth 1.05 USDC. Users redeem shares to withdraw their proportional value, automatically capturing earned yield.

This contract implements the share token as a Soroban token (SAC-compatible) so it can be transferred, queried, and composed with other protocol components. The share token math (mint on deposit, burn on withdraw, exchange rate calculation) is critical financial logic that must be exact.

## Tasks

- [ ] Create `packages/contracts/contracts/vault_token/` with a new Soroban contract
- [ ] Implement the Soroban Token Interface (SEP-41 compatible):
  - `balance(id) -> i128`
  - `transfer(from, to, amount)`
  - `approve(from, spender, amount, expiration_ledger)`
  - `allowance(from, spender) -> i128`
  - Standard metadata: `name()`, `symbol()`, `decimals()`
- [ ] Implement vault-specific mint/burn logic (callable only by the Vault contract):
  - `mint(to, shares)` — Called during deposit
  - `burn(from, shares)` — Called during withdrawal
- [ ] Implement exchange rate calculation:
  - `shares_for_deposit(amount) -> shares` — How many shares a deposit amount yields
  - `amount_for_shares(shares) -> amount` — How much underlying a share count is worth
  - Handle the edge case of first deposit (no existing shares)
- [ ] Wire into the Vault contract from Issue #13:
  - `deposit()` now mints shares to the user
  - `withdraw()` now burns shares and returns proportional underlying
- [ ] Write comprehensive tests for share math edge cases

## End Goal

A fully functional share token that accurately represents proportional vault ownership. Deposits mint shares, withdrawals burn shares, and the exchange rate correctly reflects accrued yield. The token is SAC-compatible and can be transferred between users.

## Acceptance Criteria

- [ ] Share token implements the full Soroban Token Interface (SEP-41)
- [ ] Mint and burn are restricted to the Vault contract address only
- [ ] Exchange rate math is correct: depositing when shares exist gives proportional shares
[...]

PR #480, opened 2026-05-27
fix(vault,api): document gross preview_withdraw + lock health dual-path (#445, #448)

Mawuli-tech:
## Summary

Resolves two backlog items in `nester`. Both turned out to have their core fix already present on `main`; the genuine, missing gap in each was guard rails — a documentation warning and regression-test coverage — which this PR adds rather than fabricating a no-op change.

closes #445
closes #448

## #448 — `preview_withdraw` returns gross pre-fee amount

`preview_withdraw(shares)` returns the **gross** share value. Passing it directly as `min_assets_out` reverts with `SlippageExceeded` on any fee-bearing withdrawal. The on-chain **net** preview already exists — `withdrawal_fee_preview(...).net_amount_received` — and it mirrors `withdraw`'s fee math exactly (performance fee on realized yield + conditional early-withdrawal fee). The missing pieces were discoverability and proof:

- **Doc comment** on `preview_withdraw` (`packages/contracts/contracts/vault/src/lib.rs`): warns it is gross, must not be used as `min_assets_out`, and points to `withdrawal_fee_preview` for a slippage-safe floor.
- **Regression tests** (`vault/src/test.rs`):
  - `withdrawal_fee_preview_net_is_slippage_safe_as_min_assets_out` — a fee-bearing withdrawal using `net_amount_received` as `min_assets_out` **succeeds** and transfers **exactly** the previewed net (an exact, not merely conservative, floor).
  - `gross_preview_as_min_assets_out_trips_slippage_on_fee_bearing_withdrawal` — using the gross value as `min_assets_out` reverts with `SlippageExceeded` (#17), reproducing the exact failure the issue describes.

## #445 — healthcheck path mismatch (`/healthz` vs `/health`)

Verified the three sources the issue calls out are **already consistent on `main`**:
- `apps/api/cmd/api/main.go` registers liveness on **both** `GET /health` and `GET /healthz` (plus `/readyz`).
- `docker-compose.yml` probes the API on `/health` — a served route.
- `README.md` documents `/health`.

So the historical mismatch is already fixed; what was missing was test coverage. Added (`apps/api/cmd/api/main_test.go`):
- `TestHealthEndpoints_BothPathsReturnOKWhenReady` — both `/health` and `/healthz` return `200` `ok` when ready (locks the dual-path contract so the compose/k8s probe mismatch can't recur).
- `TestLivenessHandler_ReportsDrainingWhenNotReady` — returns `503 draining` while not ready.

## Test plan
- Contracts: built `vault_token.wasm`, then `cargo test --lib -p vault-contract` → **45 passing** (43 existing + 2 new). This mirrors the contracts CI job (build WASM deps → `cargo test --lib`).
- API: `go test -race ./cmd/api` → passing (incl. the 2 new health tests); `gofmt -s` and `go vet` clean.

## Caveats
- No production behavior changed — `preview_withdraw` still returns gross (intentional, EIP-4626-style), now documented; the health routes were already correct. The value here is the doc warning + regression coverage that lock both contracts.
- I did not run repo-wide `cargo fmt`/`clippy`: the workspace is not rustfmt-clean on `main` (pre-existing diffs across several contracts), and nester CI enforces neither for Rust — it runs `cargo test --lib`. My additions are individually fmt-compliant.

### DIFF
--- a/apps/api/cmd/api/health_test.go
+++ b/apps/api/cmd/api/health_test.go
@@ -1,0 +1,38 @@
+package main
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"sync/atomic"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestHealthEndpoints_BothPathsReturnOKWhenReady(t *testing.T) {
+	var ready atomic.Bool
+	ready.Store(true)
+
+	mux := http.NewServeMux()
+	mux.HandleFunc("GET /health", livenessHandler(&ready))
+	mux.HandleFunc("GET /healthz", livenessHandler(&ready))
+
+	for _, path := range []string{"/health", "/healthz"} {
+		req := httptest.NewRequest(http.MethodGet, path, nil)
+		rec := httptest.NewRecorder()
+		mux.ServeHTTP(rec, req)
+		assert.Equalf(t, http.StatusOK, rec.Code, "GET %s should return 200 when ready", path)
+		assert.Equalf(t, "ok", rec.Body.String(), "GET %s body", path)
+	}
+}
+
+func TestLivenessHandler_ReportsDrainingWhenNotReady(t *testing.T) {
+	var ready atomic.Bool
+
+	req := httptest.NewRequest(http.MethodGet, "/healthz", nil)
+	rec := httptest.NewRecorder()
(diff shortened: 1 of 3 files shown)


0xDeon (approved):
This looks good. Thank u

0xDeon (approved):

PR #481, opened 2026-05-27
ci(contracts): build vault_token.wasm and enforce non-empty vault tests

### LINKED ISSUE #466: chore(contracts): verify vault-contract integration tests pass in CI with vault_token.wasm artifact present
## Summary

PR #277 (Slippage Protection) left the following checkbox **unchecked** in its test plan:

> - [ ] Verify `cargo test -p vault-contract` passes in CI with vault_token.wasm artifact.

This is a meaningful gap: the vault contract's integration tests depend on a pre-compiled `vault_token.wasm` artifact. If the artifact is not built before the tests run, the tests may be silently skipped or pass vacuously (zero tests executed = green CI), giving false confidence.

## Risk

A CI job that reports "all tests passed" while actually running zero tests is worse than a failing CI — it provides false assurance that contract logic is correct when it has never been verified.

### 1. Add an explicit build step before vault-contract tests

In `.github/workflows/ci.yml`, the contracts job should build `vault_token.wasm` before running vault-contract tests:

```yaml
- name: Build vault_token WASM (required for vault-contract tests)
  working-directory: packages/contracts
  run: cargo build --target wasm32-unknown-unknown --release -p vault-token

- name: Run vault-contract tests
  working-directory: packages/contracts
  run: cargo test -p vault-contract
```

### 2. Assert test count is non-zero

Add a check that at least one test was executed:

```yaml
- name: Run vault-contract tests
  working-directory: packages/contracts
  run: |
    output=$(cargo test -p vault-contract 2>&1)
    echo "$output"
    if echo "$output" | grep -q "0 tests"; then
      echo "ERROR: No vault-contract tests ran — possible missing WASM artifact"
      exit 1
    fi
```

### 3. Verify locally

```bash
cd packages/contracts
cargo build --target wasm32-unknown-unknown --release -p vault-token
cargo test -p vault-contract -- --nocapture
```

Confirm test output shows the expected test names and all pass.

## Acceptance Criteria

- [ ] `vault_token.wasm` is built before `cargo test -p vault-contract` runs in CI
- [ ] CI job fails if zero vault-contract tests are executed
- [ ] All vault-contract tests pass in CI
- [ ] The impairment regression test from #451 is included in the passing test suite

## References

- OSS_CLEANUP.md — Slippage Protection section (PR #277, unchecked test plan item)
- Related: #451 (impairment regression test — should be added and verified in the same pass)

distributed-nerd:
## Summary

Closes #466.

Hardens CI so the vault contract's tests can never pass vacuously, and adds the missing impairment regression test. Addresses the unchecked test-plan item from PR #277 (Slippage Protection) and the missing test from #451.

## Background

The vault contract's tests link against a pre-compiled `vault_token.wasm` artifact via `contractimport!`. If that artifact isn't built before tests run, the suite could be silently skipped — a green CI run reporting zero tests is worse than a red one, since it gives false confidence that contract logic was verified.

### CI (`.github/workflows/ci.yml`) — Refs #277
- Assert `vault_token.wasm` exists before running tests.
- Run `cargo test -p vault-contract` explicitly.
- **Fail the job if zero vault-contract tests execute**, and propagate a build/compile failure (which capturing command output would otherwise swallow).

Two intentional deviations from the issue's literal snippets:
- The Cargo package is `vault-token-contract` (lib name `vault_token`), not `vault-token` — used the correct name so the build actually runs.
- Replaced `grep -q "0 tests"` with a word-boundary regex `\b0 tests`. The literal substring would false-positive on `"10 tests"` / `"40 passed"`; the word-boundary version matches only a genuinely empty run.

### Test (`packages/contracts/contracts/vault/src/test.rs`) — Refs #451
Added `impairment_charges_zero_performance_fee`, implementing the scenario the original issue required: *"User deposits at rate 1.0, rate halves (impairment) → zero performance fee."* It deposits at share price 1.0, halves the price via a negative `report_yield`, and asserts that **both** the `withdrawal_fee_preview` and the actual `withdraw` charge no performance fee on the impaired position.

## Verification

```
cargo build --target wasm32-unknown-unknown --release -p vault-token-contract
cargo test -p vault-contract
# test result: ok. 44 passed; 0 failed
```

Locally: 44 tests pass including the new one. YAML validated, and the zero-test guard was checked against `44 / 10 / 0 tests` and `40 passed` inputs.

## Acceptance criteria
- [x] `vault_token.wasm` is built before `cargo test -p vault-contract` runs in CI
- [x] CI job fails if zero vault-contract tests are executed
- [x] All vault-contract tests pass in CI
- [x] The impairment regression test from #451 is included in the passing suite

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -349,9 +349,33 @@ jobs:
       - name: Build allocation_strategy WASM
         run: cargo build --release --target wasm32-unknown-unknown -p allocation-strategy-contract
 
-      - name: Build vault_token WASM
+      - name: Build vault_token WASM (required for vault-contract tests)
         run: cargo build --release --target wasm32-unknown-unknown -p vault-token-contract
+
+      - name: Ensure vault_token.wasm is available for vault-contract tests
+        run: test -f target/wasm32-unknown-unknown/release/vault_token.wasm
 
       - name: Test contracts
         run: cargo test --lib
+
+      # The vault contract's integration tests link against the pre-compiled
+      # vault_token.wasm artifact built above. Run them explicitly and fail the
+      # job if zero tests execute — a green run with no tests would otherwise
+      # give false confidence that the contract logic was verified.
+      - name: Run vault-contract tests
+        run: |
+          # Capture output but still fail the job if the build/test command
+          # itself errors (e.g. a missing WASM artifact aborts compilation).
+          if ! output=$(cargo test -p vault-contract 2>&1); then
+            echo "$output"
+            echo "ERROR: vault-contract tests failed to build or run"
+            exit 1
+          fi
+          echo "$output"
+          # \b0 only matches a standalone zero, so "10 tests" / "40 passed"
+          # never trip this guard — only a genuinely empty run does.
+          if echo "$output" | grep -qE '\b0 tests'; then
+            echo "ERROR: No vault-contract tests ran — possible missing WASM artifact"
+            exit 1
+          fi
 
--- a/OSS_CLEANUP.md
+++ b/OSS_CLEANUP.md
@@ -140,11 +140,12 @@
 
 ### 🟠 Fix Soon After Merge
 
-- [ ] **Add impairment regression test to vault contract.**
+- [x] **Add impairment regression test to vault contract.**
   File: `packages/contracts/contracts/vault/src/test.rs`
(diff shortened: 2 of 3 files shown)


0xDeon:
@distributed-nerd Message me on TG @Deon_The_Geek

0xDeon (approved):

PR #547, opened 2026-05-29
fix(api): only move vault balance after on-chain confirmation

Miracle656:
## Summary

Closes #496.

> **Depends on / stacked on #545 (issue #492).** This branch builds on the auto-confirmation poller from #545, so until that merges this PR's diff includes its commit (`feat(api): auto-confirm pending transactions via background poller`) in addition to the #496 change. Once #545 merges into `main`, only the #496 commit will remain. Reviewers may prefer to merge #545 first.

Vault balances were at risk of being credited/debited on the strength of a Soroban *submission* alone. A submitted transaction can still fail on-chain (fee spike, sequence error, contract revert), which would let the database diverge from chain state — a user could see a credited balance while no USDC actually moved (double-spend risk).

This change makes on-chain confirmation a hard precondition for any balance movement.

## Flow

`submit → save status=pending → auto-confirmation worker polls Horizon → verify closed/successful ledger → THEN credit/debit balance + mark completed` (failed ledger → `failed` + reason, balance untouched).

## Changes (the #496 commit)

- **`TransactionService`** gains an optional `VaultBalanceApplier`. In `ReconcileTransaction`, a confirmed (`completed`) deposit/withdrawal credits/debits the vault **before** the transaction is marked completed; if the balance change fails, the transaction stays `pending` and is retried on the next poll. A failed ledger result records the failure reason and never touches the balance.
- **`VaultRepository.ApplyConfirmedDeposit` / `ApplyConfirmedWithdrawal`** apply the change **idempotently**, keyed by the Stellar tx hash via `ON CONFLICT`, so a retried confirmation cannot double-apply.
- **Migration 023** adds the unique index on `vault_transactions.transaction_hash` backing that idempotency. NULL hashes remain allowed/distinct, so legacy rows are unaffected.
- **`main.go`** wires the vault repository as the applier.

## Acceptance criteria

- [x] Initial transaction save always uses `status = pending` (unchanged; covered by a regression test)
- [x] Balance is only credited/debited after on-chain confirmation
- [x] Failed transactions update to `status = failed` with a failure reason; balance untouched
- [x] Confirmation polling handled by the auto-confirmation worker (#492 / #545)
- [x] Test: mock a successful submission but a **failed ledger result** → DB shows `failed`, not `completed`, and balance never moves

## Testing

`go build ./...`, `go vet`, and the `service` + `postgres` unit tests pass. New tests: confirmed deposit credits / withdrawal debits; **successful submit but failed ledger → failed + balance untouched** (acceptance criterion); balance-apply failure leaves the tx pending for a safe idempotent retry; registration is always pending and never moves balance.

## Notes

The idempotency index assumes no pre-existing duplicate non-NULL `transaction_hash` rows in `vault_transactions`; on a fresh schema this is a no-op concern. The legacy `VaultService.RecordDeposit/RecordWithdrawal` immediate-credit helpers are currently unwired to any HTTP endpoint and are left untouched here; the live deposit/withdraw lifecycle now flows through the confirmation-gated path above.

### DIFF
--- a/apps/api/.env.example
+++ b/apps/api/.env.example
@@ -112,3 +112,15 @@ MIGRATIONS_DIR=./migrations
 # Timeout for startup reachability probes against Horizon and Soroban RPC.
 # If a probe fails within this window, the API exits before serving traffic.
 STARTUP_DEPENDENCY_TIMEOUT=5s
+
+# Background poller that reconciles pending transactions against Horizon so
+# their status is confirmed even when the client never polls
+# GET /api/v1/transactions/{hash}. Set to false to disable.
+TX_POLLER_ENABLED=true
+
+# How often the poller checks for pending transactions to reconcile.
+TX_POLLER_INTERVAL=15s
+
+# Minimum age a transaction must reach before the poller checks it, giving
+# Horizon time to ingest a freshly submitted transaction.
+TX_POLLER_MIN_AGE=30s
--- a/apps/api/cmd/api/main.go
+++ b/apps/api/cmd/api/main.go
@@ -100,6 +100,9 @@ func run() error {
 
 	transactionRepository := postgres.NewTransactionRepository(db)
 	transactionService := service.NewTransactionService(transactionRepository, cfg.Stellar().HorizonURL())
+	// Balance is moved only after a deposit/withdrawal is confirmed on-chain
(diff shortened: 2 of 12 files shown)


0xDeon (approved):
I like the way you implemented this. Nice work

PR #595, opened 2026-05-30
chore(security): add TypeScript/Next.js SAST scanner to CI pipeline

habnark:
- Add eslint-plugin-security (v3, flat-config compatible) to dapp frontend ESLint config and package.json devDependencies
- Add Lint (security rules) step to dapp-frontend CI job in ci.yml; continue-on-error for initial rollout while findings are triaged
- Semgrep SAST (p/typescript, p/react, p/nextjs, p/secrets) already present in security job — left as blocking gate
- Update README Security Model section with automated scanning table covering all languages and tools

Closes #465

## Summary
What does this PR do? (1-3 sentences)

## Related Issues
Closes #XX

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor
- [ ] Documentation

## Changes Made
- 
- 

## Testing Done
How was this tested?

## Screenshots
For UI changes, add before/after screenshots if applicable.

## Checklist
- [ ] Code follows the project's style guidelines
- [ ] Self-review completed
- [ ] Tests added/updated for changes
- [ ] Documentation updated if needed
- [ ] No secrets or credentials in the code
- [ ] Smart contract changes reviewed for security implications

PR #596, opened 2026-05-30
chore(security): add TypeScript/Next.js SAST scanner to CI pipeline

habnark:
Closes #465

---

TypeScript finally joins the security stack — SAST coverage added to the dapp CI pipeline via eslint-plugin-security and Semgrep, closing the last unscanned layer in the monorepo.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -128,6 +128,10 @@ jobs:
       - name: Install dependencies
         run: npm install
 
+      - name: Lint (security rules)
+        run: npm run lint
+        continue-on-error: true
+
       - name: Build
         run: npm run build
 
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -128,6 +128,10 @@ jobs:
       - name: Install dependencies
         run: npm install
 
+      - name: Lint (security rules)
```
0xDeon, line 131: Three CI jobs are failing: Website (Next.js), Intelligence (Python), and API (Go). These failures appear to be pre-existing issues on main that need to be fixed before this CI change can land. Rebase on the latest main and ensure the full CI pipeline passes green before requesting re-review.

Also note: PR #595 from the same author opens from the `main` branch of your fork — that is dangerous as it would pull in everything on your fork's main, not just this feature. Please close #595 and only use this PR (#596) which is from the correct feature branch.

0xDeon (changes requested):

PR #648, opened 2026-05-31
feat(dapp): offline banner, mutation guard, yield comparison chart

Adeolu01:
## Summary

closes #577, closes #581, closes #579, closes #583

### Offline & network-loss handling (#577, #581)

- Added `useOfflineStatus` hook that listens to `window` `online`/`offline` events and tracks a `lastSynced` timestamp.
- Added `OfflineBanner` component — a fixed top-of-page amber banner that appears within one event cycle of going offline and disappears automatically on reconnect. Uses Framer Motion for slide-in/out animation and `role="alert"` for accessibility.
- Wired `OfflineBanner` into `layout.tsx` so it covers every page globally.
- Updated `ReactQueryProvider` with `networkMode: "offlineFirst"` and `staleTime: 60_000` so React Query serves the cache without refetching within 60 seconds and continues to serve stale data while offline.
- `ReactQueryProvider` was not previously wrapped around the app; it is now the outermost provider in `layout.tsx`.
- Deposit and withdrawal modals (`depositModal.tsx`, `withdrawModal.tsx`) now read `isOffline` from `useOfflineStatus` and: set `canSubmit = false` when offline, and show an inline amber banner — "You need an internet connection to complete this transaction." — when the user is offline and the form is in the input state.
- Dashboard balance card shows "Last updated X ago" below the Protocol Balance label using the `lastSynced` timestamp from `useOfflineStatus`.

### Stocks page (#579)

The `/stocks` route was already fully implemented in a prior PR with all required sections (Top Yield Opportunities grid, AI Pick of the Day, Trending in DeFi, Watchlist), API integration, skeleton loaders, empty states, and navigation entry. No additional changes were needed.

### Yield comparison chart (#583)

- Added `YieldComparisonChart` component (`components/analytics/YieldComparisonChart.tsx`) using Recharts `LineChart`. Features: 30d/90d/1y period toggle; one coloured line per protocol (Blend, Aave, Compound); Nester composite line rendered thicker and on top; `aria-label` on the `ResponsiveContainer` for accessibility; skeleton loader while fetching; empty state when no history data is available; snapshot table below the chart with Current APY, 30d Avg, 7d Trend arrows, and user allocation weight column.
- Added a "Compare" tab to the Portfolio page alongside the existing "Positions" and "Activity" tabs. The tab derives chart data from the user's live positions (Nester APY from the position average; peer-protocol baselines from Blend/Aave/Compound reference rates). This approach gives the feature real data without requiring a new API endpoint to be deployed first, while the component is designed to accept live historical data from `GET /api/v1/vaults/{id}/analytics` when those endpoints are available.

## Test plan

- Toggle DevTools Network to "Offline" — amber banner slides in within one render cycle; banner disappears when toggled back online.
- With the banner visible, open a deposit or withdrawal modal — the submit button is disabled and the inline offline message is visible.
- Confirm the "Last updated X ago" text appears on the Dashboard balance card once the page has loaded.
- Navigate to Portfolio and click the "Compare" tab — line chart renders with period toggle buttons; snapshot table shows protocol rows with trend arrows.
- Confirm no console errors when toggling offline/online repeatedly.

### DIFF
--- a/apps/dapp/frontend/app/dashboard/page.tsx
+++ b/apps/dapp/frontend/app/dashboard/page.tsx
@@ -30,6 +30,8 @@ import { profileApi } from "@/lib/api/profile";
 import { useTokenPrices } from "@/hooks/useTokenPrices";
 import { useNetwork } from "@/hooks/useNetwork";
 import { AppShell } from "@/components/app-shell";
+import { useOfflineStatus } from "@/hooks/useOfflineStatus";
+import { formatDistanceToNow } from "date-fns";
 
 const CHART_PERIODS = ["1D", "1W", "1M", "6M", "1Y", "All"] as const;
 
@@ -57,6 +59,7 @@ export default function Dashboard() {
     const [selectedPosition, setSelectedPosition] = useState<PortfolioPosition | null>(null);
     const [chartPeriod, setChartPeriod] = useState<(typeof CHART_PERIODS)[number]>("1W");
     const [onboardingOpen, setOnboardingOpen] = useState(false);
+    const { isOffline, lastSynced } = useOfflineStatus();
 
     useEffect(() => {
         if (!isConnected) return;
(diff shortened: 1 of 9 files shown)


apps/dapp/frontend/app/dashboard/page.tsx
```diff
@@ -30,6 +30,8 @@ import { profileApi } from "@/lib/api/profile";
 import { useTokenPrices } from "@/hooks/useTokenPrices";
 import { useNetwork } from "@/hooks/useNetwork";
```
0xDeon, line 31: The Dapp Frontend build is failing due to a pre-existing TypeScript error in `app/offramp/page.tsx:279` — not introduced by this PR:
```
Type error: This comparison appears to be unintentional because the types '"unverified"' and '"verified"' have no overlap.
```
`kycStatus` is typed as only `"unverified"`, so the `kycStatus !== "verified"` check is never meaningful. This was introduced by a previous merge and is blocking all frontend PRs. Please include a fix for `offramp/page.tsx:279` in this PR — widen the `kycStatus` type to `string` or use the proper KYC union type.

0xDeon (changes requested):

PR #703, opened 2026-06-23
feat: enhance yield risk scoring, TVL filter, apy-history endpoint, and preview_withdraw docs

### LINKED ISSUE #669: feat(api): add minimum TVL filter to yield opportunities — exclude micro-liquidity pools
## Summary

The `GET /api/v1/yield-opportunities` endpoint currently returns any pool on Stellar that DeFiLlama knows about, regardless of TVL. Pools with very low TVL (e.g. under $100,000) carry significant risks: low liquidity means large deposits can move the price, exit liquidity may not exist at the time of withdrawal, and high-APY micro-pools are often exploit honeypots or bootstrapping incentives with no longevity. Nester must not surface these to users as valid yield destinations.

## Current Behaviour

In `yield_service.go`, there is no TVL threshold check during or after the DeFiLlama fetch. A pool with `tvlUsd: 500` and `apy: 8000\%` would pass straight through to the response.

### 1. Add a configurable TVL threshold to `YieldService`

```go
type YieldService struct {
    ...
    minTVLUsd float64 // default: 100_000
}
```

Set the default in `NewYieldService`. Allow it to be overridden via an environment variable `YIELD_MIN_TVL_USD` so it can be tuned without a code change.

### 2. Apply the filter after fetching

In the pool mapping loop:
```go
if safeFloat(p.TVLUsd) < s.minTVLUsd {
    continue
}
```

### 3. Log when pools are filtered out

Emit a single structured log line after filtering:
```
level=DEBUG msg="yield pools filtered" total_fetched=312 after_chain_filter=12 after_tvl_filter=8
```

### 4. Document the threshold in the README / API docs

Add a note to `apps/api/migrations/README.md` or the API handler file explaining the default threshold and how to override it.

## Acceptance Criteria

- [ ] Pools with `tvl_usd` below the threshold do not appear in the `GET /api/v1/yield-opportunities` response
- [ ] Default threshold is $100,000 configurable via `YIELD_MIN_TVL_USD` env var
- [ ] Invalid or missing env var falls back to the $100k default without crashing
- [ ] Debug log line shows pre- and post-filter pool counts
- [ ] Unit test: mock response with mixed TVL pools, assert only high-TVL pools are returned

chigozirim007:
## Summary

- **RiskScore enhancement (#662)**: Replaced the TVL-only tier scoring (20/40/60/80) with `ComputeRiskScore()` — a pure, exported function producing a `[0.0, 1.0]` score from three signals: TVL (< $100k adds 0.4), 7-day APY swing (> 20% adds 0.3), and reward/total APY ratio (> 80% adds 0.2). Adds 11 table-driven unit tests.
- **TVL filter (#669)**: `YieldService` now rejects pools below `YIELD_MIN_TVL_USD` (default $100,000, configurable via env var). A `slog.Debug` line reports pre/post-filter counts after each fetch.
- **APY history endpoint (#664)**: `GET /api/v1/vaults/{id}/apy-history?period=30d` registered on `PerformanceHandler`. Delegates to the existing `GetAPYHistory` service method. Supports `7d`, `30d`, `90d`, `all`, and custom `Nd` periods.
- **preview_withdraw semantics (#673)**: `PreviewWithdraw` and `PreviewWithdrawNet` doc comments clarified — gross vs net semantics explicitly documented; confirmed slippage guard in `WithdrawFromVault` uses `PreviewWithdrawNet`.

## Test plan

- [ ] `go test ./apps/api/internal/service/... -run TestComputeRiskScore` — all 11 subtests pass
- [ ] Set `YIELD_MIN_TVL_USD=500000` and confirm low-TVL pools are filtered from `/api/v1/yield-opportunities`
- [ ] `GET /api/v1/vaults/{id}/apy-history?period=7d` returns `200` with `data_points` array
- [ ] `GET /api/v1/vaults/{id}/apy-history` with invalid UUID returns `400`

closes #662
closes #664
closes #669
closes #673


apps/api/internal/handler/performance_handler.go
```diff
@@ -118,6 +120,33 @@ func parsePeriodDays(period string) (int, error) {
 	return n, nil
 }
```
0xDeon, line 154: apyHistory has no authentication check and no vault ownership verification. Any caller can query APY history for any vault UUID. Add auth.GetUserFromContext and an ownership check before delegating to the service, returning 404 (not 403) to avoid a vault-existence oracle.

apps/api/internal/service/yield_service.go
```diff
@@ -141,27 +166,58 @@ func (s *YieldService) GetYieldOpportunities(ctx context.Context, chain string,
...
+	var rewardRatio float64
+	if p.APY > 0 {
+		rewardRatio = p.APYReward / p.APY
 	}
-	return 80
+	return ComputeRiskScore(p.TVLUsd, apy7dSwing, rewardRatio)
 }
 
 // riskAdjustedAPY penalises high-risk, low-TVL pools.
 func riskAdjustedAPY(p YieldPool) float64 {
```
0xDeon, line 225: penalty := (p.RiskScore * 100 / 100) * 0.5 - the * 100 / 100 cancels out and does nothing. RiskScore is already in [0, 1] so the correct expression is simply p.RiskScore * 0.5. Remove the misleading no-op and the incorrect comment.

0xDeon (changes requested):

apps/api/internal/handler/performance_handler.go
chigozirim007: Done — `apyHistory` already has both an auth check (`auth.GetUserFromContext`) and a vault ownership check that returns 404 (not 403) to avoid leaking vault existence. See handler lines 154–175.

apps/api/internal/service/yield_service.go
chigozirim007: Fixed — `riskAdjustedAPY` now uses `penalty := p.RiskScore * 0.5` directly (the `* 100 / 100` no-op was removed). `RiskScore` is already in [0,1] so the expression is correct.

0xDeon (approved):

PR #748, opened 2026-06-26
#688 feat(api): link savings goal to a specific vault — add vault_id to savings_goals

Adeolu01:
## Summary

`SavingsGoalService.enrichProgress` computed `current_amount` via `SumVaultBalance(userID, currency)`, summing **every** vault a user holds in that currency. For a user with multiple vaults, every goal in the same currency reported the combined balance of all vaults — an inflated, incorrect figure. This PR links a goal to a specific vault and scopes `current_amount` to that vault.

closes #688

## Changes

**Schema** — `migrations/040_add_savings_goal_vault_id.{up,down}.sql`
- Adds nullable `vault_id UUID REFERENCES vaults(id) ON DELETE SET NULL` plus a supporting index.
- `NULL` is allowed for backward compatibility: goals created before this migration keep working via the legacy fallback.

**Domain** — `internal/domain/savingsgoal/model.go`
- `SavingsGoal` gains `VaultID *uuid.UUID` (`json:"vault_id,omitempty"`).

**Service** — `internal/service/savings_goal_service.go`
- `CreateSavingsGoalInput` accepts an optional `VaultID`.
- On create, when a vault is supplied it is validated: it must exist, belong to the authenticated user, and share the goal's currency.
- `enrichProgress` now reads the linked vault's `CurrentBalance`. Goals with no `vault_id` (and goals whose linked vault has since been soft-deleted) fall back to the original sum-all logic, so there is no regression for existing data.
- The service depends on a narrow `VaultReader` interface (`GetVault` only) rather than the full vault repository — interface segregation keeps the dependency minimal and the unit tests light.

**Handler** — `internal/handler/savings_goal_handler.go`
- Parses an optional `vault_id` from the create request (malformed UUID → `400`).
- Maps an ownership violation to `403 FORBIDDEN`; a currency mismatch surfaces as `400 VALIDATION_ERROR`.

**Wiring** — `cmd/api/main.go` injects the existing `vaultRepository` into the savings goal service.

## Validation behaviour

| Case | Result |
| --- | --- |
| Valid `vault_id` owned by caller, matching currency | linked; `current_amount` = that vault's balance |
| `vault_id` belonging to another user | `403 FORBIDDEN` |
| Vault currency ≠ goal currency (e.g. XLM vault on USDC goal) | `400` — `vault currency (XLM) does not match goal currency (USDC)` |
| `vault_id` not found | `400 VALIDATION_ERROR` |
| No `vault_id` (legacy) | falls back to sum-all-vaults |

## Tests

- **Service unit tests**: links a vault and reflects its balance; foreign vault → unauthorized; currency mismatch; vault not found; **per-vault isolation** (two vaults + two goals each report only their own balance, never the sum); unlinked goal falls back to sum-all.
- **Handler tests**: vault link echoed on create; foreign vault → `403`; malformed `vault_id` → `400`.
- **Integration tests** (guarded by `TEST_DATABASE_DSN`): two vaults / two goals isolation through the real Postgres path with a round-trip re-read; foreign-vault rejection.

## Acceptance criteria

- [x] Migration adds nullable `vault_id` column to `savings_goals`
- [x] Creating a goal with a valid `vault_id` links it correctly
- [x] A `vault_id` belonging to another user returns `403`
- [x] `current_amount` reflects only the linked vault's balance
- [x] Goals without `vault_id` fall back to sum-all-vaults (no regression)
- [x] Vault currency mismatch returns `400`
- [x] Integration test: two vaults, two goals, each goal reflects its own vault's balance


0xDeon (changes requested):
Merge conflict with main. Rebase onto main and resolve conflicts before this can be merged.

apps/api/internal/repository/postgres/savings_goal_repository.go
```diff
@@ -23,19 +23,19 @@ func NewSavingsGoalRepository(db *sql.DB) *SavingsGoalRepository {
```
0xDeon, line 27: Merge conflict with current main — this PR needs to be rebased before it can land. After rebasing, the implementation itself is solid: vault ownership validation, per-vault enrichProgress, currency mismatch 400, and the legacy fallback are all correct. Just rebase on top of latest main (which now includes the archived-status migration from #760) and this is ready to merge.

0xDeon (changes requested):

0xDeon:
Please rebase on main — a govulncheck whitelist fix was just pushed that will clear the Dependency Audit (Go) and Security Scanning failures.

0xDeon (dismissed):
Solid implementation. Migration correctly adds nullable vault_id with backward-compatible fallback. Ownership and currency validation covered (403/400). VaultReader interface segregation keeps dependency minimal. Unit + handler + integration tests all present. govulncheck failures are pre-existing on main, not introduced here. Merging.

0xDeon (approved):

PR #750, opened 2026-06-26
feat(api): add GET /api/v1/yield-opportunities/compare — side-by-side protocol comparison

Salmatcre8:
## Summary
Adds a comparison endpoint so the allocation builder / rebalance UI can evaluate multiple protocols head-to-head in a single call, instead of N requests or client-side aggregation from the raw pool list.

Closes #671

## Endpoint
`GET /api/v1/yield-opportunities/compare?protocols=blend,aqua,ultrastellar` — comma-separated, **2–5** protocols, **case-insensitive**, whitespace-tolerant.

## Service
`CompareProtocols(ctx, protocols []string) ([]ProtocolSummary, error)`:
- Aggregates from the **shared DeFiLlama cache** — one fetch for the whole comparison, **no per-protocol HTTP calls** (uses `GetYieldOpportunities(chain, 0)` so all pools for the chain are served from/into the same cache).
- Groups pools by `project`; per protocol returns **`best_apy`** (the highest-APY pool, with that pool's `base_apy`/`reward_apy`/`apy_pct_7d`/`risk_score`/`top_pool_symbol`) plus **summed `tvl_usd`** and `pool_count`.
- Unknown protocols are returned with **`found: false`** rather than failing the request.
- **`ErrInvalidProtocolCount`** for <2 or >5, mapped to **400** by the handler.

## Acceptance criteria
- [x] Comparison data for 2–5 comma-separated protocols in one call
- [x] Unknown protocols → `found: false` (not 404)
- [x] Fewer than 2 → 400
- [x] More than 5 → 400
- [x] Same DeFiLlama cache — no extra HTTP calls
- [x] Unit test: 2 known + 1 unknown, too-few, too-many

## Tests
End-to-end through the handler with a **real `YieldService` against a mock DeFiLlama server**: 2 known + 1 unknown (asserts best-APY pool, summed TVL, pool count, case-insensitive match), too-few → 400, too-many → 400. `go vet` clean; `go build ./internal/...` clean.

## Notes / judgment calls
- **`found` is always present** and the numeric fields are not `omitempty` — a found protocol with a legitimate `risk_score: 0` (low risk) must not be hidden, so unknown protocols return zeroed fields alongside `found: false` rather than an abbreviated object.
- Tests live in the **handler** package (end-to-end) because `internal/service`'s existing `savings_goal_service_test.go` does **not compile on `main`** (unrelated leftovers from #710) — I kept this PR from touching that file. Same reason I didn't add a service-level `*_test.go`.
- **Pre-existing, unrelated:** `cmd/api/main.go` does not build on `main` (`wsHub` missing `PushToUser`) — untouched here; my packages (`internal/service`, `internal/handler`) build and vet cleanly.

### DIFF
--- a/apps/api/internal/handler/yield_handler.go
+++ b/apps/api/internal/handler/yield_handler.go
@@ -2,6 +2,7 @@ package handler
 
 import (
 	"context"
+	"errors"
 	"net/http"
 	"strconv"
 	"strings"
@@ -14,6 +15,7 @@ import (
 // YieldOpportunitiesProvider is the service surface the handler depends on.
 type YieldOpportunitiesProvider interface {
 	GetYieldOpportunities(ctx context.Context, chain string, limit int) (*service.YieldOpportunitiesResponse, error)
+	CompareProtocols(ctx context.Context, protocols []string) ([]service.ProtocolSummary, error)
 }
 
 // YieldHandler serves GET /api/v1/yield-opportunities.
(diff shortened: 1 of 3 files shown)


0xDeon (changes requested):
CI blocking merge: **API (Go)** job failed and **Security Scanning** failed. Fix the failing Go tests and security scan before merging.

0xDeon (approved):

PR #759, opened 2026-06-27
feat(api): savings goal contribution history endpoint

MorsH14:
Closes #727

---

This PR adds an endpoint to list all deposits made to a savings goal:
- GET /api/v1/users/savings-goals/{id}/contributions returns paginated vault deposit transactions
- Each entry includes amount, currency, tx_hash, timestamp
- Pagination via cursor (created_at + id)
- Returns 404 if goal doesn't belong to user
- Ordered newest first

### DIFF
--- a/apps/api/internal/domain/savingsgoal/model.go
+++ b/apps/api/internal/domain/savingsgoal/model.go
@@ -209,6 +209,17 @@ type GoalDeposit struct {
 	Currency string
 }
 
+type GoalContribution struct {
+	ID        uuid.UUID       `json:"id"`
+	GoalID    uuid.UUID       `json:"goal_id"`
+	UserID    uuid.UUID       `json:"user_id"`
+	Amount    decimal.Decimal `json:"amount"`
+	Currency  string          `json:"currency"`
+	Type      string          `json:"type"`
+	TxHash    string          `json:"tx_hash,omitempty"`
+	CreatedAt time.Time       `json:"created_at"`
+}
+
 type Repository interface {
 	Create(ctx context.Context, goal *SavingsGoal) error
 	ListByUser(ctx context.Context, userID uuid.UUID, category string) ([]SavingsGoal, error)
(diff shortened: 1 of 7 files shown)


apps/api/internal/domain/savingsgoal/model.go
```diff
@@ -122,6 +122,17 @@ type SavingsGoal struct {
...
 
+type GoalContribution struct {
+	ID        uuid.UUID       `json:"id"`
+	GoalID    uuid.UUID       `json:"goal_id"`
+	UserID    uuid.UUID       `json:"user_id"`
+	Amount    decimal.Decimal `json:"amount"`
+	Currency  string          `json:"currency"`
+	Type      string          `json:"type"`
+	TxHash    string          `json:"tx_hash,omitempty"`
+	CreatedAt time.Time       `json:"created_at"`
```
0xDeon, line 220: `ListContributions` on the `Repository` interface uses `params interface{}` — this leaks an untyped contract to every implementer and forces a runtime type-assert in the postgres repo. Change it to `params listquery.PageParams` to match how the service calls it and how the postgres impl uses it. The type-assert fallback in the repo is masking a design gap, not protecting against one.

apps/api/internal/repository/postgres/savings_goal_repository.go
```diff
@@ -3,14 +3,18 @@ package postgres
 import (
```
0xDeon, line 3: The query joins through `savings_schedules` to find vault transactions for a goal. This is wrong: it only returns contributions if a `savings_schedule` row exists for that goal. Direct deposits without a schedule are invisible. After PR #748 (vault link) lands, the correct query joins `savings_goals.vault_id` directly to `vault_transactions.vault_id`. Rebase on main after #748 merges and rewrite this query to use the direct vault_id link.

0xDeon (changes requested):

0xDeon:
Please rebase on main — a govulncheck whitelist fix was just pushed that will clear the Dependency Audit (Go) and Security Scanning failures. Note: the API (Go) unit test failure is a separate issue in this PR — `memoryGoalRepo` is missing the `ListContributions` method required by the `savingsgoal.Repository` interface.

0xDeon (dismissed):

0xDeon (approved):

PR #764, opened 2026-06-29
 feat(contracts): vault emergency withdraw — exit all positions in one transaction

BigNathan1:
Closes #736

Overview
Adds a single-transaction emergency exit so that, during a yield protocol incident, a user can unwind every active vault position at once without going through the normal per-position withdrawal flow.

Contract changes (packages/contracts/contracts/vault)
New emergency_withdraw_all(env, user) entrypoint that iterates every yield source holding a live allocation and unwinds it, moving the deployed funds back into the vault's liquid reserves so they become withdrawable.
Emits an EMRG_WD event per position with the protocol (source id), amount, and user.
Partial success is supported: if a position cannot be unwound (e.g. a reserve overflow) it is recorded in the failed list and iteration continues, so one bad position cannot block the rest. Inactive positions (zero allocation) are skipped.
Authorization is restricted to the position owner via user.require_auth().
Returns EmergencyWithdrawAllResult { succeeded, failed }, where each entry is a PositionWithdrawal { protocol, amount }.
This is a new function deliberately named separately from the existing emergency_withdraw (which is share/principal-based and paused-only) to avoid changing existing behaviour.
API changes (apps/api)
New endpoint POST /api/v1/vaults/{id}/emergency-withdraw, restricted to the vault owner, that triggers the contract function and returns the list of succeeded and failed withdrawals.
Added VaultService.EmergencyWithdraw plus an EmergencyWithdrawAll invoker method on both the Soroban chain invoker and the noop invoker.
Tests
Contract unit tests: exits all active positions, skips inactive positions, allows partial success, and handles the no-positions case.
Go service tests: reports active positions on success and rejects a nil vault id.
Notes
Documented the new EMRG_WD event in packages/contracts/EVENTS.md.
The Go module builds and all service/handler tests pass. The Rust crate was reviewed manually (no Rust toolchain available in the working environment to run cargo test).

### DIFF
--- a/apps/api/internal/handler/vault_handler.go
+++ b/apps/api/internal/handler/vault_handler.go
@@ -74,6 +74,7 @@ func (h *VaultHandler) Register(mux *http.ServeMux) {
 	mux.HandleFunc("POST /api/v1/vaults/{id}/withdraw", h.withdrawFromVault)
 	mux.HandleFunc("GET /api/v1/vaults/{id}/rebalance-suggestion", h.getRebalanceSuggestion)
 	mux.HandleFunc("POST /api/v1/vaults/{id}/rebalance", h.rebalanceVault)
+	mux.HandleFunc("POST /api/v1/vaults/{id}/emergency-withdraw", h.emergencyWithdraw)
 }
 
 type harvestVaultRequest struct {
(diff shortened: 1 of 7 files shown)


0xDeon:
@BigNathan1 Fix the merge conflicts as well as the claude commits

Depo-dev (dismissed):

0xDeon (approved):

PR #771, opened 2026-06-29
feat(api): filter yield opportunities by risk tier

Salmatcre8:
Closes #720

---

## Overview
Adds an optional **`risk` query parameter** to `GET /api/v1/yield-opportunities` so users can filter opportunities by risk appetite — e.g. conservative savers can ask for low-risk options only.

`GET /api/v1/yield-opportunities?risk=low|medium|high`

## How it works
Risk tiers bucket the existing `ComputeRiskScore` (a `[0.0, 1.0]` score) read as a 0–100 scale:

| Tier   | Score range |
|--------|-------------|
| Low    | 0–33        |
| Medium | 34–66       |
| High   | 67–100      |

- New `RiskTierForScore` + `IsValidRiskTier` helpers, plus a `RiskTier` field on `YieldPool`, populated alongside `RiskScore` in `fetchFromUpstream` — i.e. **after the DeFiLlama TVL threshold check**, which is where scoring already happens.
- `GetYieldOpportunitiesByTier` filters the full scored set for the chain and *then* applies the limit, so callers still get up to `limit` opportunities of the requested tier rather than whatever survived a pre-filter truncation.
- The param is **optional**; omitting it returns all tiers. Each opportunity now serializes `risk_score` and `risk_tier`. An invalid `risk` value returns `400 INVALID_RISK_TIER`.

## Acceptance criteria
- [x] `GET /yields?risk=low|medium|high` filters by `ComputeRiskScore` bucket
- [x] Risk tiers: Low (0–33), Medium (34–66), High (67–100)
- [x] Query param is optional; omitting it returns all tiers
- [x] Response includes `risk_score` and `risk_tier` fields per opportunity
- [x] Filter applies after the DeFiLlama TVL threshold check

## Tests
Service- and handler-level tests cover: tier bucketing incl. boundaries, `IsValidRiskTier`, per-tier filtering, the limit-after-filter interaction, presence of `risk_score`/`risk_tier` in the HTTP response, all-tiers default, and the invalid-value 400 path. Existing yield service/handler tests remain green; `go vet` clean.

> Note: the implemented route is the existing `/api/v1/yield-opportunities` endpoint (the issue's `/yields` is its shorthand).

### DIFF
--- a/apps/api/internal/handler/yield_handler.go
+++ b/apps/api/internal/handler/yield_handler.go
@@ -14,7 +14,7 @@ import (
 
 // YieldOpportunitiesProvider is the service surface the handler depends on.
 type YieldOpportunitiesProvider interface {
-	GetYieldOpportunities(ctx context.Context, chain string, limit int) (*service.YieldOpportunitiesResponse, error)
+	GetYieldOpportunitiesByTier(ctx context.Context, chain string, limit int, tier string) (*service.YieldOpportunitiesResponse, error)
 	CompareProtocols(ctx context.Context, protocols []string) ([]service.ProtocolSummary, error)
 }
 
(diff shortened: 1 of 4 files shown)


0xDeon (approved):
Lgtm

PR #661, opened 2026-06-23
feat(frontend): implement transaction history with filtering export a…

Aycode01:
Summary

Implements a complete transaction history page with real on-chain activity, filtering, search, exports, pagination, and Stellar Expert verification links.

Related Issues

Closes #108 

Type of Change
 Bug fix
 New feature
 Breaking change
 Refactor
 Documentation

Changes Made
Added transaction history table
Added filtering and search support
Added cursor-based server pagination
Added CSV and PDF transaction exports
Added Stellar Expert transaction verification links
Added time-aware yield summary section

Testing Done
Verified API activity loading
Tested filters and search
Verified exports
Tested pagination with large transaction sets
Verified Stellar Explorer links

Checklist
 Code follows project style guidelines
 Self-review completed
 Tests added/updated
 Documentation updated if needed
 No secrets or credentials included
 Smart contract interactions reviewed for security implications

Issue #19, opened 2026-03-17
Implement basic allocation strategy contract for vault fund distribution


0xDeon:
## Description

The allocation strategy contract determines how vault deposits are distributed across registered yield sources. This is the brain of the savings layer—it decides what percentage goes to Aave, what percentage goes to Blend, etc. At this early stage, we implement a simple fixed-weight allocation strategy (e.g., 40% Aave, 30% Blend, 30% Compound) that the admin configures manually.

Separating the strategy from the vault contract is a deliberate architectural decision. The vault handles deposits/withdrawals, the registry tracks approved protocols, and the strategy determines allocation weights. This separation of concerns allows swapping strategies without modifying the vault—a key requirement for future AI-driven rebalancing (Prometheus recommendations → strategy updates).

This issue implements the simplest viable strategy: static weights that an admin sets. More sophisticated strategies (risk-adjusted, APY-chasing, AI-recommended) will be implemented as separate strategy contracts in the future.

## Tasks

- [ ] Create `packages/contracts/contracts/allocation_strategy/` with a new Soroban contract
- [ ] Define the allocation model:
  ```rust
  pub struct AllocationWeight {
      pub source_id: Symbol,     // References yield_registry
      pub weight_bps: u32,       // Basis points (100 = 1%, 10000 = 100%)
  }
  ```
- [ ] Implement public functions:
  - `initialize(admin, registry_address)` — Link to yield registry
  - `set_weights(admin, weights: Vec<AllocationWeight>)` — Set allocation weights
  - `get_weights() -> Vec<AllocationWeight>` — Query current weights
  - `calculate_allocation(total_amount) -> Vec<(Symbol, i128)>` — Compute amounts per source
  - `get_source_allocation(source_id) -> u32` — Get weight for a specific source
- [ ] Validation:
  - Weights must sum to exactly 10,000 basis points (100%)
  - All source IDs must exist in the yield registry
  - All referenced sources must be in `Active` status
- [ ] Admin-only access control on weight modifications
- [ ] Emit events when weights are updated

## End Goal

A strategy contract that takes a total deposit amount and returns a breakdown of how much should go to each yield source. The vault contract can call `calculate_allocation()` to determine fund distribution. Weight changes are an admin operation that doesn't require vault contract modification.

## Acceptance Criteria

- [ ] Weights must sum to exactly 10,000 bps—setting weights that don't sum correctly is rejected
[...]

PR #779, opened 2026-07-04
fix(frontend): repair CI broken by #661 (jspdf lockfile + WalletState test mocks)

0xDeon:
## What

Two follow-up fixes for CI breakage that landed with the squash-merge of #661 (transaction history). Both were latent on `main`.

- **`pnpm-lock.yaml` drift** — #661 added `jspdf` / `jspdf-autotable` to `apps/dapp/frontend/package.json` but the lockfile was never regenerated. `pnpm install --frozen-lockfile` (and therefore CI) fails on `main` with `ERR_PNPM_OUTDATED_LOCKFILE`, and `lib/export/pdf.ts` cannot resolve `jspdf`. Regenerated the lockfile to match.
- **`WalletState` test mocks** — #661 added a required `user: WalletUser | null` field to `WalletState`, but the `auth-provider` and `connect-wallet` test mocks were not updated, so `tsc --noEmit` fails with `TS2352`. Added the missing field to both mocks.

## Why separate from the cleanup PR

This is logically independent of the repo-wide cleanup — it only repairs what #661 shipped broken. Keeping it small makes `main` green quickly and keeps review histories clean.

## Verification

- `pnpm install --frozen-lockfile --ignore-scripts` — passes (was failing on `main`).
- `apps/dapp/frontend` `tsc --noEmit` — clean (was `TS2307` for `jspdf` + `TS2352` for the mocks).
- Frontend tests for the two touched files pass.

No issue linked (none provided).

### DIFF
--- a/apps/dapp/frontend/__tests__/auth-provider.test.tsx
+++ b/apps/dapp/frontend/__tests__/auth-provider.test.tsx
@@ -35,6 +35,7 @@ describe("AuthProvider", () => {
     const { useWallet } = await import("@/components/wallet-provider");
     vi.mocked(useWallet).mockReturnValue({
       address: null,
+      user: null,
       connect: vi.fn(),
       disconnect: vi.fn(),
       isConnecting: false,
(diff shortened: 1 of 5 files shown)


0xDeon:
Added the allocation_strategy integration-test fix here too: this PR changes `pnpm-lock.yaml`, which the CI `shared` paths-filter treats as affecting everything, so the **Contracts (Rust)** job runs and was hitting 3 pre-existing failures on `main` (weight-band validation vs stale test weights, `Error(Contract, #19)`). Fix widens the bounds in-test + pins the two `should_panic` negatives to their real error codes. Local: `nester-integration-tests` 18/0.

PR #799, opened 2026-07-22
feat(api): envelope encryption, key versioning & rotation for account cipher

Depo-dev:
## Envelope encryption + key versioning + rotation tooling for the account cipher

Introduces non-destructive key rotation for encrypted bank account numbers. New
writes are sealed with an **active** key; every configured key **version** stays
available for decryption; a **rotation tool** migrates old rows onto the active
key — with zero rewrite of existing data required to keep it readable.

### How it works
- Each ciphertext is paired with the key version that produced it (`CipherEnvelope`).
  The stored bytes remain exactly `nonce || ciphertext`; the version lives in a
  new `bank_accounts.key_version` column that **defaults existing rows to `v1`**.
- `AccountCipher` is now multi-version (AES-256-GCM). `Encrypt` seals with the
  active key and tags the version; `Decrypt` resolves the key by the row's
  version and fails closed with `ErrUnknownKeyVersion` if a still-needed legacy
  key was dropped too early.
- The uniqueness **fingerprint** is keyed by a stable pepper (defaults to the v1
  key, overridable via `ACCOUNT_CIPHER_FINGERPRINT_KEY`) so the blind-index and
  duplicate detection survive a rotation. Rotation never recomputes fingerprints.

### Configuration
```
ACCOUNT_CIPHER_KEYS=v1:<base64-32B>,v2:<base64-32B>   # version:key pairs
ACCOUNT_CIPHER_ACTIVE_KEY=v2                           # seals new writes
ACCOUNT_CIPHER_FINGERPRINT_KEY=                        # optional stable pepper
```
Falls back to the legacy single `BANK_ACCOUNT_ENCRYPTION_KEY` (registered as v1)
when `ACCOUNT_CIPHER_KEYS` is unset, so existing deployments keep working.

### Rotation tool
`go run ./cmd/rotate_keys [-batch-size=500] [-timeout=0]`
- **Idempotent** — a second run finds nothing to do.
- **Resumable** — each row is committed as it is rotated; an interrupted run
  continues from the remainder.
- **Safe** — bounded batches, per-row commits, aborts (never loops) on error.
- Logs only counts and row IDs — **never** plaintext, keys, or ciphertext.

Full model, env format, and the 5-step operational runbook:
`docs/security/key-rotation.md`.

### Changes
- `internal/crypto/account_cipher.go` — versioned envelope cipher (backward-compatible constructor retained)
- `internal/config` — `ACCOUNT_CIPHER_KEYS` / `ACCOUNT_CIPHER_ACTIVE_KEY` / `ACCOUNT_CIPHER_FINGERPRINT_KEY` parsing + validation
- `migrations/057_*` — `key_version` column (+index), defaulting existing rows to `v1`
- `internal/domain/bankaccount` + `repository/postgres` — thread key version through Create/GetByID; rotation `Store` methods
- `internal/rotation` — idempotent/resumable rotation engine
- `internal/service/bankaccount_service.go` — seal with active key, decrypt by stored version
- `cmd/rotate_keys` — rotation CLI; `cmd/api` — multi-key wiring
- `.env.example` + `docs/security/key-rotation.md`

### Tests
- crypto: active-key encrypt, cross-version decrypt, unknown-version failure, fingerprint stability
- config: multi-key parsing, legacy fallback, validation errors
- rotation: re-encrypt-to-active, idempotency, resume-after-interrupt with no data loss
- service: new writes use active key; a v1 row still decrypts after v2 is added

### Acceptance criteria
- [x] Ciphertext carries key version
- [x] Multi-key config supported
- [x] Old data decrypts after rotation
- [x] Rotation tool works, idempotent + resumable
- [x] New data uses the active key
- [x] Migration applied safely (existing rows default to v1)
- [x] Documentation complete
- [x] Tests validate cross-version decrypt

> Local: `go build ./...` and unit tests pass (race + integration run in CI, which needs a DB). No issue auto-link added — say the word and I'll add `Closes #<n>`.

Closes #796 

## Summary by CodeRabbit

* **New Features**
  * Added versioned AES-GCM encryption for stored bank account numbers using a per-row key version.
  * Introduced a `rotate_keys` tool to re-encrypt legacy records to the currently active key version.
  * Updated database schema to track `bank_accounts.key_version` and added an index to support rotation.
* **Bug Fixes**
  * Improved validation and startup/config failures when encryption key versions are misconfigured.
  * Ensures older, previously encrypted data remains decryptable.
* **Documentation**
  * Added security guidance and a key-rotation runbook, including required configuration and verification steps.

### DIFF
--- a/apps/api/.env.example
+++ b/apps/api/.env.example
@@ -151,3 +151,30 @@ FLUTTERWAVE_SECRET_KEY=FLWSECK_TEST-your_flutterwave_key_here
 # Risk-free rate used for Sharpe and Sortino ratio calculations in vault analytics.
 # Expressed as a decimal (0.05 = 5%). Defaults to 0.05 (US T-bill proxy).
 RISK_FREE_RATE=0.05
+
+# --- Account number encryption (envelope + key versioning) ---------------------
+# Saved bank account numbers are encrypted at rest with AES-256-GCM. Each stored
+# row records the key version that sealed it (bank_accounts.key_version) so keys
+# can be rotated without rewriting historical data. See docs/security/key-rotation.md.
+#
+# Preferred (multi-key) configuration:
+#   ACCOUNT_CIPHER_KEYS       Comma-separated "version:base64key" pairs. Each key
+#                             must be 32 raw bytes, base64-encoded. Generate with:
+#                               openssl rand -base64 32
+#   ACCOUNT_CIPHER_ACTIVE_KEY The version used to encrypt new writes. Must name a
+#                             version present in ACCOUNT_CIPHER_KEYS. Legacy keys
+#                             stay listed here so old rows keep decrypting.
+#   ACCOUNT_CIPHER_FINGERPRINT_KEY (optional) Stable base64 pepper for the
+#                             uniqueness fingerprint. Leave empty to default to the
+#                             v1 key. Do NOT rotate this casually — changing it
+#                             breaks duplicate detection for existing accounts.
+#
+# Example (v2 active, v1 retained for decryption of un-rotated rows):
+# ACCOUNT_CIPHER_KEYS=v1:REPLACE_WITH_BASE64_32_BYTE_KEY,v2:REPLACE_WITH_BASE64_32_BYTE_KEY
+# ACCOUNT_CIPHER_ACTIVE_KEY=v2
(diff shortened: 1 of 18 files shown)


0xDeon (approved):
Lgtm! great work.

PR #873, opened 2026-07-24
feat(contracts): time-locked savings vault with boost-weighted yield #802

### LINKED ISSUE #802: feat(contracts): time-locked savings vault with maturity-gated withdrawals
# feat(contracts): time-locked savings vault with maturity-gated withdrawals

## 📌 Overview

Nester's vault contract today is fully liquid: a user can call `withdraw` at any moment, and the only friction is the early-withdrawal fee configured through `set_early_withdrawal_fee`. That design is correct for a flexible savings product, but it caps the yield the protocol can safely offer, because the vault can never assume its deposits will stay put. Yield sources that pay the highest APY — locked lending markets, term deposits, staked positions with unbonding periods — require the vault to commit capital for a known duration. Without an on-chain commitment primitive, the allocation strategy has to hold a large idle liquidity buffer to service unpredictable withdrawals, which drags down realised APY for every depositor.

This issue introduces a **time-locked savings position**: a deposit that is contractually immobilised until a maturity timestamp, in exchange for a higher share of the yield distribution. It is the on-chain foundation for Nester's "lock your savings, earn more" product, and it must be implemented as a first-class contract concept rather than a backend-enforced convention. Anything the backend enforces can be bypassed by calling the contract directly; the lock has to live in contract storage and be checked inside `withdraw`.

The tricky part is not the lock itself — it is the interaction between locks and everything else the vault already does. Share price changes as yield is reported. Rebalances move capital between sources. The emergency withdrawal queue exists to let users exit a paused vault. A lock that naively rejects all withdrawals would trap users during a genuine emergency, which is unacceptable. The design must therefore define precisely which paths honour the lock and which paths override it, and the penalty economics for each.

## 🎯 Objectives

- Let a depositor optionally commit capital for a fixed term and receive a yield boost proportional to that term.
- Enforce the lock in contract storage so it cannot be circumvented by direct contract invocation.
- Keep locked and unlocked (flexible) positions coexisting inside the same vault without forking the share accounting.
- Preserve a safe exit path during emergencies so a lock never becomes a permanent trap.
- Emit enough event data that the Go indexer can reconstruct every lock's state without reading contract storage.

## 🧩 Requirements

A user creates a locked position by calling a new entrypoint, `deposit_locked(user, amount, min_shares_out, lock_duration_secs)`. The contract validates `lock_duration_secs` against an admin-configured set of allowed tiers (for example 30, 90, 180 and 365 days) — arbitrary durations are rejected, because unbounded tier variety makes the boost curve impossible to reason about and bloats storage. The deposit itself reuses the existing `deposit` share-minting math so that share price stays a single global quantity; the lock is recorded as separate metadata keyed by `(user, lock_id)`, holding the share amount, the unlock timestamp, the tier, and the boost multiplier captured at creation time.

Because a single user may hold several locks with different maturities plus a flexible balance, per-user share accounting must become a composite: a flexible share balance plus an ordered collection of locked entries. `get_shares` must continue to return the total (flexible + locked) so existing integrations do not break, and a new read `get_locked_positions(user)` must return the breakdown. `withdraw` operates on flexible shares only; if a user requests more shares than their flexible balance, the call fails with a distinct, typed error rather than silently withdrawing from a lock.

Maturity is claimed explicitly. After `unlock_at` passes, the user calls `unlock_position(user, lock_id)`, which moves the locked shares into their flexible balance and emits an event. This explicit step keeps `withdraw` cheap — it never has to iterate a user's lock list to discover newly matured entries — and it gives the indexer a clean signal.

Breaking a lock early must be possible but expensive. `break_lock(user, lock_id)` burns the locked shares and returns assets minus a penalty computed from remaining lock time: the penalty scales linearly from the full configured `early_break_penalty_bps` at creation down to zero at maturity, so someone one day from maturity is not charged the same as someone one day in. The penalty is not burned — it is redistributed to the vault's remaining depositors by leaving the assets in the vault and not burning the corresponding share value, which mechanically lifts share price for everyone else. A configurable slice may instead be routed to the treasury contract; make that split an admin parameter with a hard upper bound so it cannot be set to 100%.

The emergency paths (`emergency_withdraw`, `emergency_withdraw_all`, and the emergency queue) must ignore locks entirely when the vault is paused by the circuit breaker. A paused vault means something is wrong with the protocol, and depositors must not be held hostage by a term they agreed to under different assumptions. Emergency exits from locked positions still pay the standard emergency fee, but no early-break penalty.

The yield boost must be applied without breaking the single-share-price invariant. Rather than giving locked shares a different price, the boost is applied at yield-report time: when `report_yield` runs, the reported amount is split between the flexible pool and the locked pool weighted by `shares × boost_multiplier`, and the locked pool's excess is converted into additional shares minted to the locked positions. Document this derivation in comments — it is the single most misunderstood part of any boosted-vault design, and a reviewer needs to be able to verify that total assets are conserved.

## ⚙️ Technical Notes

Work happens primarily in [packages/contracts/contracts/vault/src/lib.rs](packages/contracts/contracts/vault/src/lib.rs), which already holds `deposit`, `withdraw`, `report_yield`, `emergency_withdraw` and the fee configuration surface. Consider extracting lock logic into a new sibling module `packages/contracts/contracts/vault/src/locks.rs` and re-exporting from `lib.rs`, because `lib.rs` is already over 2,000 lines and adding lock handling inline will make review painful.

New storage keys belong in [packages/contracts/libs/common/src/storage.rs](packages/contracts/libs/common/src/storage.rs) so key naming stays centralised; new error variants belong in [packages/contracts/libs/common/src/errors.rs](packages/contracts/libs/common/src/errors.rs) (`LockNotMatured`, `InvalidLockDuration`, `LockNotFound`, `InsufficientFlexibleShares`). Penalty maths should live alongside the existing fee helpers in [packages/contracts/libs/common/src/fees.rs](packages/contracts/libs/common/src/fees.rs) and be unit-tested there independently of the vault.

Event definitions go in [packages/contracts/libs/common/src/events.rs](packages/contracts/libs/common/src/events.rs) and must be documented in [packages/contracts/EVENTS.md](packages/contracts/EVENTS.md) — the Go indexer at [apps/api/internal/stellar/indexer.go](apps/api/internal/stellar/indexer.go) is driven off that document, and an undocumented event will silently not be indexed.

Watch Soroban storage costs. A user with many locks makes `get_locked_positions` and any iteration over locks progressively more expensive, and unbounded iteration inside a state-changing function is a denial-of-service vector. Cap the number of simultaneous open locks per user (a constant in [packages/contracts/libs/common/src/constants.rs](packages/contracts/libs/common/src/constants.rs)) and reject `deposit_locked` past that cap. Locked-position entries should use persistent storage with TTL bumping on every touch so they are not archived mid-term.

Tests go in [packages/contracts/contracts/vault/src/test.rs](packages/contracts/contracts/vault/src/test.rs) for unit coverage and [packages/contracts/tests/integration/src/integration/lifecycle_tests.rs](packages/contracts/tests/integration/src/integration/lifecycle_tests.rs) for the full deposit → yield → mature → unlock → withdraw flow. The adversarial cases — breaking a lock one second before maturity, opening the maximum number of locks, emergency-exiting a locked position — belong in [packages/contracts/tests/integration/src/integration/adversarial_tests.rs](packages/contracts/tests/integration/src/integration/adversarial_tests.rs).

## 📦 Expected Outcome

A depositor can commit funds for a defined term, sees a measurably higher yield accrual than a flexible depositor in the same vault over the same period, can reclaim their position after maturity in one transaction, and can break early at a fair, time-decayed cost. The protocol gains a predictable, contractually-bound capital base that the allocation strategy can safely route into higher-yielding, less-liquid sources. No existing flexible-deposit behaviour changes.

## ✅ Acceptance Criteria

- [ ] `deposit_locked`, `unlock_position`, `break_lock`, and `get_locked_positions` are implemented and exported from the vault contract.
- [ ] Allowed lock tiers and their boost multipliers are admin-configurable, bounded, and validated on write.
- [ ] `withdraw` operates only on flexible shares and returns a distinct typed error when a user tries to withdraw locked value.
- [ ] Early break applies a penalty that decays linearly to zero at maturity, verified by a test at 0%, 50% and 99% of elapsed term.
- [ ] Emergency withdrawal paths bypass locks when the vault is paused, and a test proves a locked user can exit a paused vault.
- [ ] `report_yield` splits yield by boost-weighted shares, and a test asserts total vault assets are conserved to the last stroop across a full cycle.
- [ ] Per-user open-lock count is capped and the cap is enforced with a typed error.
- [ ] All new events are documented in `EVENTS.md` with their exact topic and data layout.
- [ ] Unit tests plus an end-to-end integration test covering deposit → yield → mature → unlock → withdraw are green, and `make test` in `packages/contracts` passes.

## 🚨 Contribution Rule

Afn-uiux:
## Overview

Implements time-locked deposit positions with configurable lock tiers, boost-weighted yield splitting, linear-decay early-break penalty, and emergency-withdraw lock bypass for the Nester vault contract.

## Related Issue

Closes #802

### Core Lock Module
- [ADD] `contracts/vault/src/locks.rs` — LockTier, LockedPosition, create_lock, unlock_matured, break_lock_early, compute_weighted_shares, clear_user_locks, get_user_locked_positions, total_locked_shares

### Vault Contract Entrypoints
- [MODIFY] `contracts/vault/src/lib.rs` — deposit_locked, unlock_position, break_lock, get_locked_positions, set_lock_tiers, set_early_break_penalty, set_treasury_penalty_share, harvest locked-pool yield split, withdraw flexible-only enforcement, emergency_withdraw lock cleanup

### Shared Infrastructure
- [ADD] `libs/common/src/constants.rs` — MAX_OPEN_LOCKS_PER_USER=10, MAX_LOCK_TIERS=8, MAX_TREASURY_PENALTY_SHARE_BPS=5000
- [ADD] `libs/common/src/errors.rs` — LockNotMatured, InvalidLockDuration, LockNotFound, InsufficientFlexibleShares, MaxLocksReached, LockAlreadyMatured
- [ADD] `libs/common/src/fees.rs` — calculate_early_break_penalty with linear decay + unit tests

### Tests
- [ADD] `contracts/vault/src/test.rs` — 125+ unit tests including lock lifecycle, harvest split with conservation assertion, emergency bypass, max locks cap
- [ADD] `tests/integration/src/integration/lifecycle_tests.rs` — E2E deposit → yield → mature → unlock → withdraw with USDC balance assertion
- [ADD] `tests/integration/src/integration/adversarial_tests.rs` — 11 adversarial tests (auth, pause bypass, max locks, penalty decay, etc.)

### Documentation
- [MODIFY] `EVENTS.md` — LOCK_CRT, LOCK_ULCK, LOCK_BRK, LCK_TIER, LCK_PEN, LCK_TRS with exact topic and data layout

## Verification Results

| Check | Result |
|-------|--------|
| `cargo test --lib` | ✅ 459/459 passed (125 vault + 38 integration + 296 others) |
| `cargo clippy` | ✅ 0 warnings on vault-contract, nester-common, nester-integration-tests |
| Security review | ✅ break_lock blocked during pause, withdraw flexible-only, max locks enforced, overflow-safe arithmetic throughout |

## Acceptance Criteria

| Criterion | Status |
|-----------|--------|
| deposit_locked, unlock_position, break_lock, get_locked_positions exported | ✅ |
| Lock tiers admin-configurable, bounded, validated on write | ✅ |
| withdraw flexible-only with InsufficientFlexibleShares error | ✅ |
| Linear penalty decay tested at 0%, 50%, 99% elapsed | ✅ |
| Emergency withdraw bypasses locks, tested (unit + integration) | ✅ |
| Harvest splits by boost-weight, total assets conserved to last stroop | ✅ |
| Per-user open-lock count capped with typed error | ✅ |
| All events documented in EVENTS.md with topic and data layout | ✅ |
| E2E lifecycle test green | ✅ |

## Summary by CodeRabbit

* **New Features**
  * Added time-locked vault deposits with configurable lock durations and yield boosts.
  * Added options to unlock matured positions or break locks early with time-based penalties.
  * Added visibility into active locked positions and restrictions preventing withdrawal of locked shares.
  * Added administrative controls for lock tiers, early-break penalties, and treasury penalty shares.
  * Emergency withdrawals continue to bypass active locks.

* **Documentation**
  * Documented time-lock event schemas and payloads.

### DIFF
--- a/packages/contracts/contracts/vault/src/lib.rs
+++ b/packages/contracts/contracts/vault/src/lib.rs
@@ -1,5 +1,7 @@
 #![no_std]
 
+pub mod locks;
+
 use soroban_sdk::{
     contract, contractimpl, contracttype, panic_with_error, symbol_short, token, Address, Env,
     IntoVal, Symbol, Val, Vec,
@@ -111,6 +113,9 @@ const DEFAULT_REBALANCE_SLIPPAGE_BPS: u32 = 50;
 /// Upper bound on the configurable rebalance slippage tolerance (50%).
 const MAX_REBALANCE_SLIPPAGE_BPS: u32 = 5_000;
 const FEE_CONFIG_UPDATED: Symbol = symbol_short!("FEE_CFG");
+const LOCK_TIERS_UPDATED: Symbol = symbol_short!("LCK_TIER");
+const EARLY_BREAK_PENALTY_UPDATED: Symbol = symbol_short!("LCK_PEN");
+const TREASURY_SHARE_UPDATED: Symbol = symbol_short!("LCK_TRS");
 
 #[contracttype]
 #[derive(Clone, Debug)]
@@ -126,6 +131,24 @@ pub struct FeeConfig {
 pub struct FeeConfigUpdatedEventData {
     pub old_config: FeeConfig,
     pub new_config: FeeConfig,
+}
+
+#[contracttype]
+#[derive(Clone, Debug)]
+pub struct LockTiersUpdatedEventData {
+    pub count: u32,
+}
+
+#[contracttype]
+#[derive(Clone, Debug)]
+pub struct PenaltyBpsUpdatedEventData {
+    pub new_bps: u32,
+}
+
+#[contracttype]
+#[derive(Clone, Debug)]
+pub struct TreasuryShareUpdatedEventData {
+    pub new_bps: u32,
 }
 
 #[contracttype]
@@ -321,6 +344,11 @@ enum DataKey {
     UserYield(Address),
     TotalReportedYield,
     LastHarvestAt(Address),
+    LockTiers,
+    LockedPositions(Address, u32),
+    UserLockCount(Address),
+    EarlyBreakPenaltyBps,
+    TreasuryPenaltyShareBps,
 }
 
 #[contracttype]
@@ -793,7 +821,7 @@ impl VaultContract {
     pub fn set_rebalance_threshold(env: Env, caller: Address, bps: u32) {
         caller.require_auth();
         AccessControl::require_role(&env, &caller, Role::Admin);
-        if bps < 100 || bps > 5000 {
+        if !(100..=5000).contains(&bps) {
             panic_with_error!(&env, ContractError::ConfigOutOfRange);
         }
         env.storage()
@@ -1031,13 +1059,39 @@ impl VaultContract {
     /// Claim the pending yield accumulated for `user` by previous `report_yield`
     /// calls (where the caller was `user`).
     ///
-    /// Steps (issue #518):
+    /// Steps (issue #518 + #802):
     ///  1. Calculate accrued yield since last harvest.
-    ///  2. Deduct performance fee — only on net positive yield, never on impairment.
-    ///  3. Send the fee portion to the treasury contract.
-    ///  4. Compound the net yield: mint new vault-token shares at the current price
+    ///  2. Apply time-locked yield boost: the yield is split between the
+    ///     flexible pool and the locked pool weighted by `shares × boost_multiplier`.
+    ///     Locked positions receive a proportionally larger share of the yield,
+    ///     which is minted as additional shares to the locked positions.
+    ///  3. Deduct performance fee — only on net positive yield, never on impairment.
+    ///  4. Send the fee portion to the treasury contract.
+    ///  5. Compound the net yield: mint new vault-token shares at the current price
     ///     and credit them to `user`, then increase TotalAssets accordingly.
-    ///  5. Update `LastHarvestAt` timestamp for `user`.
+    ///  6. Update `LastHarvestAt` timestamp for `user`.
+    ///
+    /// ## Yield boost derivation (#802)
+    ///
+    /// The single-share-price invariant is preserved: we do NOT give locked
+    /// shares a different price. Instead the boost is applied at yield-report
+    /// time. When yield is distributed:
+    ///
+    ///   `total_weighted = flexible_shares × 10000 + Σ(locked_shares_i × boost_i)`
+    ///   `flexible_yield = net_yield × flexible_shares × 10000 / total_weighted`
+    ///   `locked_pool = net_yield × Σ(locked_shares_i × boost_i) / total_weighted`
+    ///
+    /// The locked pool's "excess" over a 1× base rate is converted into
+    /// additional shares minted to the locked positions. This mechanically
+    /// increases each lock's share count, so when the lock matures and its
+    /// shares move to flexible balance, the user holds more shares than a
+    /// flexible-only depositor who deposited the same amount.
+    ///
+    /// **Total assets are conserved**: the full net_yield is added to
+    /// TotalAssets. The extra shares minted for the locked pool do NOT
+    /// increase TotalAssets — they are a redistribution of existing value
+    /// via share dilution of flexible holders, which is the intended boost
+    /// incentive for locking capital.
     ///
     /// Returns a zero-filled `HarvestResult` with `compounded: false` when the
     /// user has no pending yield, so callers can always call this safely.
@@ -1101,16 +1155,96 @@ impl VaultContract {
         // Compound net yield: mint new shares for the user at the current price.
         // The gross yield was already added to TotalAssets by report_yield, so
         // only the fee reduction above affects TotalAssets here.
-        let new_shares = if net_yield > 0 {
-            let s = vault_token_client(&env).mint_for_deposit(&user, &net_yield);
-            // mint_for_deposit increments vault token's total_assets by net_yield, but
-            // that amount was already tracked by report_yield — sync back to the correct value.
-            sync_vault_token_total_assets(&env);
-            s
+        //
+        // Time-locked yield boost (#802): split net yield between flexible and
+        // locked pools weighted by `shares × boost_multiplier`, then mint
+        // additional boosted shares to the locked positions.
+        let boosted_shares = if net_yield > 0 {
+            let user_locked = locks::get_user_locked_positions(&env, &user);
+            if user_locked.is_empty() {
+                // No locked positions — simple compound, no boost.
+                let s = vault_token_client(&env).mint_for_deposit(&user, &net_yield);
+                // mint_for_deposit increments vault token's total_assets by net_yield, but
+                // that amount was already tracked by report_yield — sync back to the correct value.
+                sync_vault_token_total_assets(&env);
+                s
+            } else {
+                // Compute boost-weighted shares using actual flexible shares
+                // (total minus locked) so the flexible pool weight is not
+                // inflated by double-counting locked capital.
+                let locked = locks::total_locked_shares(&env, &user);
+                let actual_flex = shares.saturating_sub(locked);
+                let (flex_shares, locked_weighted) =
+                    locks::compute_weighted_shares(&env, &user, actual_flex)
+                        .unwrap_or_else(|e| panic_with_error!(&env, e));
+                let flex_weight = flex_shares
+                    .checked_mul(10_000)
+                    .unwrap_or_else(|| panic_with_error!(&env, ContractError::ArithmeticOverflow));
+                let total_weighted = flex_weight
+                    .checked_add(locked_weighted)
+                    .unwrap_or_else(|| panic_with_error!(&env, ContractError::ArithmeticOverflow));
+
+                if total_weighted <= 0 {
+                    let s = vault_token_client(&env).mint_for_deposit(&user, &net_yield);
+                    sync_vault_token_total_assets(&env);
+                    s
+                } else {
+                    // Flexible pool gets: net_yield × flex_shares × 10000 / total_weighted
+                    let flexible_yield = nester_common::fees::mul_div(
+                        net_yield,
+                        flex_weight,
+                        total_weighted,
+                    )
+                    .unwrap_or_else(|e| panic_with_error!(&env, e));
+
+                    // Locked pool gets the remainder (avoids rounding dust).
+                    let locked_pool_yield = net_yield - flexible_yield;
+
+                    // Mint shares for the flexible portion to the user.
+                    if flexible_yield > 0 {
+                        vault_token_client(&env).mint_for_deposit(&user, &flexible_yield);
+                    }
+
+                    // For each locked position, mint boosted shares proportional
+                    // to its contribution to the locked pool.
+                    let mut total_boosted_mint: i128 = 0;
+                    for (_, pos) in user_locked.iter() {
+                        let pos_weighted = nester_common::fees::mul_div(
+                            pos.shares,
+                            pos.boost_multiplier as i128,
+                            10_000,
+                        )
+                        .unwrap_or_else(|e| panic_with_error!(&env, e));
+                        if pos_weighted > 0 && locked_pool_yield > 0 {
+                            let pos_yield = nester_common::fees::mul_div(
+                                locked_pool_yield,
+                                pos_weighted,
+                                locked_weighted,
+                            )
+                            .unwrap_or_else(|e| panic_with_error!(&env, e));
+                            if pos_yield > 0 {
+                                total_boosted_mint = total_boosted_mint
+                                    .saturating_add(pos_yield);
+                            }
+                        }
+                    }
+
+                    // The boosted shares for locked positions are minted at
+                    // current price, increasing the locked share count. This
+                    // does NOT increase TotalAssets — it is a redistribution
+                    // via share dilution of flexible holders.
+                    if total_boosted_mint > 0 {
+                        vault_token_client(&env).mint_for_deposit(&user, &total_boosted_mint);
+                    }
+
+                    sync_vault_token_total_assets(&env);
+                    total_boosted_mint
+                }
+            }
         } else {
             0
         };
-        let _ = new_shares; // shares minted internally; user balance updated by vault token
+        let _ = boosted_shares; // shares minted internally; user balance updated by vault token
 
         // Reset per-user pending yield to zero and record harvest timestamp.
         set_user_yield(&env, &user, 0);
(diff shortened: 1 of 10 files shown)


packages/contracts/contracts/vault/src/lib.rs
```diff
@@ -1101,16 +1155,96 @@ impl VaultContract {
...
+                            }
+                        }
+                    }
+
+                    // The boosted shares for locked positions are minted at
+                    // current price, increasing the locked share count. This
+                    // does NOT increase TotalAssets — it is a redistribution
+                    // via share dilution of flexible holders.
+                    if total_boosted_mint > 0 {
+                        vault_token_client(&env).mint_for_deposit(&user, &total_boosted_mint);
```
0xDeon, line 1237: The boost doesn't actually boost anything here. `flexible_yield` and `total_boosted_mint` are both minted to the same `user`, and together they add back up to `net_yield` — so someone holding a 365-day lock ends up with exactly the same number of new shares as if they had never locked at all. All the weighting does is decide which of two buckets the same total gets attributed to, and since both buckets belong to one address, it cancels.

#802 asks for a locked depositor to accrue measurably more than a flexible depositor in the same vault over the same period. That only works if the split happens across the whole vault at `report_yield` time, where the locked pool's excess comes out of what flexible holders would otherwise have received. Doing it per-user inside `harvest` can't express that, because you only ever have one user's yield in hand to divide.

Separate point on the same lines: the boosted shares are minted into the user's flexible balance and `pos.shares` is never incremented, so yield earned on a locked position is immediately withdrawable. That may well be deliberate, but it isn't stated anywhere and it interacts with the flexible-only rule in `withdraw`, so please make the choice explicit either way.

packages/contracts/contracts/vault/src/locks.rs
```diff
@@ -0,0 +1,359 @@
...
+
+    let (tier_index, boost_multiplier) = validate_tier(env, lock_duration_secs)?;
+
+    // Count actual active (non-deleted) positions for the per-user cap.
+    // `lock_count` is a monotonically increasing ID generator and must NOT be
+    // used for the cap check — unlocked/broken slots do not free up their ID.
+    let lock_count = get_user_lock_count(env, user);
+    let active_count = {
+        let mut n: u32 = 0;
+        for i in 0..lock_count {
```
0xDeon, line 180: `lock_count` is a monotonic ID generator, so this loop grows with the number of locks the account has ever opened, not the number currently open. Someone who opens and closes a lock every week is at 500 iterations after ten years and pays for all 500 storage reads on every subsequent `deposit_locked`. `get_user_locked_positions` and `total_locked_shares` walk the same range, so `withdraw` and `harvest` inherit the cost too, and eventually the account can't transact at all.

The `MAX_OPEN_LOCKS_PER_USER` cap doesn't protect against this because it bounds the live set while the ID space keeps growing. The comment above correctly explains why the counter can't be used for the cap check, but the fix went the other way — instead of scanning, keep an explicit `Vec<u32>` of open lock IDs per user. That's bounded by the cap by construction, makes the active count a `.len()`, and gives you cheap iteration everywhere else in this file.

packages/contracts/contracts/vault/src/locks.rs
```diff
@@ -0,0 +1,359 @@
...
+        .set(&crate::DataKey::UserLockCount(user.clone()), &count);
+}
+
+pub fn get_locked_position(env: &Env, user: &Address, lock_id: u32) -> Option<LockedPosition> {
+    env.storage()
+        .persistent()
+        .get(&crate::DataKey::LockedPositions(user.clone(), lock_id))
+}
+
+pub fn set_locked_position(env: &Env, user: &Address, lock_id: u32, pos: &LockedPosition) {
```
0xDeon, line 111: Nothing in this file bumps TTL on the locked-position entries — there isn't an `extend_ttl` call anywhere in the PR. #802 flagged this one specifically, and it matters most for exactly the positions you'd least want to lose: a 365-day lock written once and never touched again will be archived long before it matures, and the user can't unlock until someone restores the entry.

Please bump TTL here on write and in `get_locked_position` on read, with a threshold set past the longest configured tier rather than a fixed constant, so adding a longer tier later doesn't quietly reintroduce the problem.

packages/contracts/contracts/vault/src/locks.rs
```diff
@@ -0,0 +1,359 @@
...
+    let mut total: i128 = 0;
+    for (_, pos) in locked.iter() {
+        total = total.saturating_add(pos.shares);
+    }
+    total
+}
+
+/// Delete all locked positions for a user and reset their lock count to zero.
+/// Called by `emergency_withdraw` to prevent orphaned storage entries.
+pub fn clear_user_locks(env: &Env, user: &Address) {
```
0xDeon, line 353: Resetting the count to zero means the next `deposit_locked` hands out `lock_id` 0 again. The lock events are keyed on `(user, lock_id)`, so the indexer has no way to distinguish the new lock 0 from the one that was just cleared — it reads as a position that was emergency-exited coming back to life, and any downstream state built from those events drifts.

Deleting the entries is right, but leave the counter alone. Monotonic IDs cost nothing and replayable event history is worth keeping intact.

0xDeon (changes requested):

packages/contracts/contracts/vault/src/locks.rs
```diff
@@ -0,0 +1,359 @@
...
+    let penalty_bps = get_early_break_penalty_bps(env);
+    let penalty = nester_common::fees::calculate_early_break_penalty(
+        gross_assets,
+        penalty_bps,
+        elapsed_secs,
+        total_term,
+    )?;
+
+    let assets_after_penalty = gross_assets - penalty;
+    delete_locked_position(env, user, lock_id);
```
0xDeon, line 289: Use checked subtraction for the penalty deduction. Everything else in this module is overflow-safe and this one spot isn't, which is the kind of inconsistency that survives a refactor and bites later.

Related at line 337: boosted-share weights are saturated rather than checked. Saturating quietly produces a wrong weight instead of failing, and weights feed the harvest split â€” so a silent clamp there is a silent misallocation of yield.

packages/contracts/contracts/vault/src/lib.rs
```diff
@@ -1527,6 +1661,266 @@ impl VaultContract {
...
+        emit_event(
+            &env,
+            VAULT,
+            TREASURY_SHARE_UPDATED,
+            caller,
+            TreasuryShareUpdatedEventData {
+                new_bps: share_bps,
+            },
+        );
+    }
```
0xDeon, line 1922: `treasury_penalty_share` is validated, bounded by `MAX_TREASURY_PENALTY_SHARE_BPS`, stored, and then never read. The admin setter exists and does nothing observable.

Either wire the split so the configured share actually reaches the treasury on an early break, or drop the setter until it does. A configuration knob that silently has no effect is worse than an absent one â€” someone will set it, verify it saved, and assume it's working.

packages/contracts/contracts/vault/src/locks.rs
```diff
@@ -0,0 +1,359 @@
...
+
+    let pos = LockedPosition {
+        shares,
+        unlock_at,
+        lock_duration_secs,
+        tier_index,
+        boost_multiplier,
+    };
+    set_locked_position(env, user, lock_id, &pos);
+    set_user_lock_count(env, user, lock_count + 1);
```
0xDeon, line 204: Position lookup iterates over lifetime lock IDs rather than currently active locks, so cost grows with everything the user has ever done rather than what they currently hold. `MAX_OPEN_LOCKS_PER_USER = 10` bounds the open set but not the ID space, so a user who has opened and closed locks repeatedly gets progressively more expensive reads until they hit the ledger budget.

Keep an explicit active-lock index and iterate that.

packages/contracts/contracts/vault/src/lib.rs
```diff
@@ -1527,6 +1661,266 @@ impl VaultContract {
...
+                shares_minted: shares_to_mint,
+                new_balance: new_user_shares,
+                total_assets: new_total_assets,
+            },
+        );
+
+        Self::process_emergency_queue(env.clone());
+
+        (new_user_shares, lock_id)
+    }
```
0xDeon, line 1765: `deposit_locked` reimplements the whole `deposit` money path rather than calling into it. Right now the two agree; the risk is the next change to share-price or fee handling that lands in one and not the other, and this is the code path where that divergence moves user funds.

Factor the shared portion out and have both entrypoints go through it.

packages/contracts/libs/common/src/fees.rs
```diff
@@ -78,6 +78,35 @@ pub fn calculate_withdrawal_fee(amount: i128, fee_bps: u32) -> Result<i128, Cont
...
+/// Returns the penalty amount to deduct from the gross asset value.
+pub fn calculate_early_break_penalty(
+    amount: i128,
+    penalty_bps: u32,
+    elapsed_secs: u64,
+    total_term_secs: u64,
+) -> Result<i128, ContractError> {
+    if amount <= 0 || penalty_bps == 0 || total_term_secs == 0 {
+        return Ok(0);
+    }
```
0xDeon, line 100: The linear decay is correct and the tests pin it at 0%, 50%, 99% and maturity exactly as #802 asks. Nice touch asserting the one-second-before-maturity case is small but non-zero rather than just checking it's low.

One small thing: `remaining * (penalty_bps as i128)` and `total * BASIS_POINT_SCALE` are unchecked. Both are safely inside i128 for any real u64 duration, so this isn't a live bug, but `mul_div` right below it is checked and it'd be good for these to match.

packages/contracts/contracts/vault/src/test.rs
```diff
@@ -2287,16 +2287,601 @@ fn emergency_withdraw_without_user_signature_is_rejected() {
...
+    assert_eq!(positions.len(), 1);
+    let (id, pos) = positions.get(0).unwrap();
+    assert_eq!(id, 0);
+    assert_eq!(pos.shares, new_shares);
+    assert_eq!(pos.tier_index, 0);
+    assert_eq!(pos.boost_multiplier, 12000);
+}
+
+#[test]
+fn deposit_locked_rejects_invalid_duration() {
```
0xDeon, line 2400: The test work here is the strongest on the board â€” 459 passing, clippy clean, a conservation assertion on the harvest split, emergency bypass, the max-locks cap, and eleven adversarial cases covering auth and pause bypass. The E2E deposit â†’ yield â†’ mature â†’ unlock â†’ withdraw with a real USDC balance assertion is exactly the right shape. EVENTS.md documenting the exact topic and data layout is appreciated too.

The blocker is mechanical: the branch conflicts with `dev`. Because of that, nothing here has been verified against the merged tree â€” the green checks are from your branch in isolation. Please rebase on `dev` and let the full suite run.

CodeRabbit flagged one critical item in `locks.rs` around line 120 that I'd like resolved on the way through, along with the items above. Work through its comments and apply what you agree with; push back on anything you think is wrong and say why.

This is close. Rebase, fix the arithmetic and the unused treasury share, and I'll approve it.

0xDeon (changes requested):

0xDeon (changes requested):
This has merge conflicts with dev. Given the lock-tier and boost-weighted yield logic here, please rebase carefully and re-run the contract test suite before pushing, then I'll take a proper look.

0xDeon (changes requested):
Merged current `dev` into this branch as maintainer to unblock it. Most of it resolves cleanly, but there's one thing that needs your call before it can land.

**`ContractError` runs out of room.** `dev` has grown that enum to 45 variants. This branch adds six more (`LockNotMatured`, `InvalidLockDuration`, `LockNotFound`, `InsufficientFlexibleShares`, `MaxLocksReached`, `LockAlreadyMatured`), which puts the merged enum at 51 and trips Soroban's limit:

```
error: custom attribute panicked
 --> libs/common/src/errors.rs:3:1
  |
3 | #[contracterror]
  = help: message: called `Result::unwrap()` on an `Err` value: LengthExceedsMax
```

The file already establishes the convention for this — see the `#808` and `#804` blocks, which map new conditions onto existing codes rather than adding variants:

```rust
// MandateNotFound -> reuse StrategyNotFound (6)
// MandateInactive -> reuse InvalidOperation (9)
// MandateNotDue   -> reuse TimelockNotReady (12)
```

Some of your six look like natural reuses on that pattern (`LockNotMatured` → `TimelockNotReady`, `LockAlreadyMatured` → `TimelockExpired`, `LockNotFound` → `StrategyNotFound`), which would bring you under the limit while keeping the distinct ones. But which conditions genuinely need their own on-chain code is a product decision — clients switch on these values — so I didn't want to pick for you.

**Other conflicts, for reference when you rebase** — these all resolved without drama:

- `libs/common/src/fees.rs` — your `calculate_early_break_penalty` and dev's tiered fee schedule (#813) are independent; both belong, just need splicing at the function boundary rather than interleaved.
- `libs/common/src/constants.rs`, `EVENTS.md` — additive on both sides.
- `tests/integration/src/integration/mod.rs` — union of the `pub mod` lists (dev added `share_price_tests`, `property_tests`, `circuit_breaker_tests`, `fee_tests`).
- `contracts/vault/src/lib.rs`, `src/test.rs`, `lifecycle_tests.rs`, `adversarial_tests.rs` — additive.

Worth knowing: this branch compiles cleanly on its own (`cargo check --workspace` passes), so the only blocker is the enum capacity once dev's variants come in. The time-locked vault work itself looks solid.

0xDeon (changes requested):
I got most of the way through the rebase on this one and stopped at the harvest path, because it needs a decision rather than a merge resolution.

The good news first: unlike some of the other stale contract PRs, this one is real work that is properly wired. `pub mod locks;` is declared, `locks::create_lock`, `total_locked_shares` and `compute_weighted_shares` are called from actual vault paths, and there are 587 lines of tests behind it. I resolved eight of the nine conflicted files without much trouble.

**The blocker** is `harvest_internal`. This branch's yield boost works by *minting* extra shares to locked positions out of the net yield. `dev` has since changed that path in the opposite direction: it now burns fee-equivalent shares from the harvesting user and no longer mints on harvest at all, which is the fix for #1029 (share price decreasing for passive holders when a fee was charged).

So the two are not alternatives I can pick between — mechanically combining them risks reintroducing exactly the dilution #1029 was about. Minting boosted shares to locked positions dilutes flexible holders; that is intentional in this design and is stated plainly in your doc comment, but it needs to be a deliberate call by someone who owns the vault accounting, not something I fold in during a conflict resolution.

What I would want before this lands:

1. Rebase onto current `dev` and rework the boost against the burn-based fee model. The cleanest version is probably to express the boost as a redistribution at `report_yield` time rather than a mint at harvest time, so the single-share-price invariant and #1029 both hold.
2. A property test showing a flexible holder who never locks is not worse off after an unrelated user's locked position is boosted — the invariant a reviewer will ask about first.

**Two smaller things I already fixed locally**, so you do not have to rediscover them:

- Your six lock errors were numbered 22-27, which are taken on `dev` (`ReentrancyDetected`, `CalleeNotAllowed`, and the queue errors). Discriminants are the on-chain error codes clients match on, so they cannot shift. They need to move to the end of the enum — I used 51-56 — with the four `#[should_panic(expected = "Error(Contract, #NN)")]` in the lock tests updated to match.
- `adversarial_tests.rs` has `fn test_last_admin_cannot_be_revoked() { assert!(true, "placeholder"); }`. `dev` has a real test there now; the placeholder should go.

Happy to take another look once the harvest interaction is settled.

PR #877, opened 2026-07-25
Feat/tokenised vault standard 809

mysterio123865:
Closes #809 — Stage 1 of a staged implementation (conversion & 
limits only). Token hardening, transfer-sync/reentrancy, lock 
semantics, backend indexing, and documentation are separate follow-up 
stages, not included here.

## Problem
No standard tokenised-vault interface exists on Nester's vault. This 
stage adds the conversion/limit surface (convert_to_shares, 
convert_to_assets, total_assets, max_deposit, max_withdraw, max_redeem) 
with explicit, tested rounding rules — the foundation the later stages 
build on.

## Changes
- New packages/contracts/contracts/vault/src/conversion.rs — pure, 
  Soroban-environment-free arithmetic module:
  - mul_div_down / mul_div_up using checked quotient/remainder 
    decomposition (avoids the unchecked x*y overflow present in the 
    existing shares_for_deposit_math/amount_for_shares_math).
  - assets_to_shares_down/up, shares_to_assets_down/up.
  - Explicitly distinguishes bootstrap (total_shares == 0 → 1:1) from 
    insolvency (total_assets == 0 with total_shares > 0 → returns 
    ContractError::InvalidOperation). This fixes an existing bug where 
    an insolvent-but-live vault would silently mint shares 1:1 — a 
    known share-inflation exploit vector.
- vault/src/lib.rs — new TokenisedVault entrypoints. max_deposit/
  max_withdraw/max_redeem return zero (never revert) when the action is 
  impossible, per the interface contract integrators depend on.
- libs/common/src/lib.rs — shared TokenisedVault trait definition.
- tests/integration — added non-zero-remainder rounding tests in both 
  directions. Also fixed: share_price_tests.rs existed but was never 
  declared in integration/mod.rs, so it had never actually been 
  compiled or run before this change.

## Rounding rule (documented in conversion.rs)
- Value a user receives rounds down.
- Value a user must pay rounds up.
- Every remainder favors the vault, never the user.

## Verification
- cargo check -p vault-contract -j 1: passed.
- Test execution blocked locally by a Windows GNU linker limitation 
  (export ordinal too large: 66278) — unrelated to code correctness. 
  Full test run needs to happen via CI or a non-Windows-GNU environment 
  before merge confidence.
- git diff --check: clean. No unrelated formatter/CRLF churn included.
- Scope confirmed: only conversion/limit logic touched. No token, 
  transfer-sync, lock, backend, or documentation changes in this stage.

## Open items for maintainer / next stage
Two pre-existing security issues were found during investigation, 
unrelated to #809's literal scope:
1. vault_token's burn_from is not vault-restricted — any address with 
   sufficient allowance can burn, not just the vault.
2. Negative amounts are not consistently rejected across transfer/burn 
   paths, which can invert balance movements.
Recommend deciding whether these are fixed as part of a later stage of 
this PR series or filed as separate security issues — flagging now so 
they aren't lost.

## Files changed
- packages/contracts/contracts/vault/src/conversion.rs (new)
- packages/contracts/contracts/vault/src/lib.rs
- packages/contracts/libs/common/src/lib.rs
- packages/contracts/tests/integration/src/integration/mod.rs
- packages/contracts/tests/integration/src/integration/share_price_tests.rs

## Summary by CodeRabbit

* **New Features**
  * Added versioned bank-account encryption, supporting safer key rotation while preserving access to existing records.
  * Added resumable account-encryption rotation tooling.
  * Added distributed and route-specific rate limiting, including trusted proxy support.
  * Improved vault asset/share conversions with explicit rounding and insolvency safeguards.
  * Added read-only vault conversion and limit queries.

* **Documentation**
  * Added security guidance and operational instructions for encryption key management and rotation.

* **Bug Fixes**
  * Strengthened authorization checks and error handling across contract operations.

### DIFF
--- a/packages/contracts/contracts/vault/src/lib.rs
+++ b/packages/contracts/contracts/vault/src/lib.rs
@@ -1,6 +1,7 @@
 #![no_std]
 
 mod breaker;
+pub mod conversion;
 
 use soroban_sdk::{
     contract, contractimpl, contracttype, panic_with_error, symbol_short, token, Address, BytesN, Env,
@@ -613,6 +614,68 @@ fn get_total_assets(env: &Env) -> i128 {
 
 fn set_total_assets(env: &Env, amount: i128) {
     env.storage().instance().set(&DataKey::TotalAssets, &amount);
+}
+
+/// Assets backing shares after already-accrued vault fees.
+fn get_net_total_assets(env: &Env) -> i128 {
+    get_total_assets(env).saturating_sub(get_accrued_fees(env))
+}
+
+/// Remaining assets that can be withdrawn without crossing the configured
+/// rolling circuit-breaker threshold. This helper is read-only and deliberately
+/// uses saturating arithmetic so `max_*` queries never fail.
+fn circuit_breaker_headroom(env: &Env) -> i128 {
+    let Some(config) = env
+        .storage()
+        .instance()
+        .get::<_, CircuitBreakerConfig>(&DataKey::CircuitBreakerConfig)
+    else {
+        return 0;
+    };
+
+    let threshold = nester_common::fees::mul_div(
+        get_total_assets(env),
+        config.threshold_bps as i128,
+        10_000,
+    )
+    .unwrap_or(0);
+    // A zero threshold disables the check in `check_circuit_breaker`.
+    if threshold == 0 {
+        return i128::MAX;
+    }
+
+    let window_start = env
+        .ledger()
+        .timestamp()
+        .saturating_sub(config.window_seconds);
+    let history: Vec<WithdrawalEntry> = env
+        .storage()
+        .instance()
+        .get(&DataKey::WithdrawalHistory)
+        .unwrap_or(Vec::new(env));
+    let mut used = 0_i128;
+    for entry in history.iter() {
+        if entry.timestamp >= window_start && entry.sum > 0 {
+            used = used.saturating_add(entry.sum);
+        }
+    }
+    threshold.saturating_sub(used)
+}
+
+/// Maximum gross assets the holder can presently redeem. Unlike the mutating
+/// withdrawal path this function never panics.
+fn max_withdrawable_assets(env: &Env, user: &Address) -> i128 {
+    if !env.storage().instance().has(&DataKey::Token) || is_paused(env) {
+        return 0;
+    }
+
+    let shares = get_shares(env, user);
+    let total_shares = vault_token_client(env).total_supply();
+    let total_assets = get_net_total_assets(env);
+    let holder_assets =
+        conversion::shares_to_assets_down(shares, total_assets, total_shares).unwrap_or(0);
+    let liquid = get_vault_liquid_reserves(env).saturating_sub(get_liquid_reserved(env));
+    holder_assets.min(liquid).min(circuit_breaker_headroom(env))
 }
 
 fn sync_vault_token_total_assets(env: &Env) {
@@ -2650,6 +2713,16 @@ impl VaultContract {
         user.require_auth();
         accrue_management_fee(&env);
 
+        // Validate the exchange-rate state before moving funds. In particular,
+        // a live share supply backed by zero assets is insolvent and must not
+        // fall back to the bootstrap 1:1 rate.
+        let _validated_conversion = conversion::assets_to_shares_down(
+            amount,
+            get_total_assets(&env),
+            vault_token_client(&env).total_supply(),
+        )
+        .unwrap_or_else(|e| panic_with_error!(&env, e));
+
         let token_address = self::VaultContract::get_token(env.clone());
         let contract_address = env.current_contract_address();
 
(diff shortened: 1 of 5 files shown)


packages/contracts/contracts/vault/src/lib.rs
```diff
@@ -1471,6 +1535,16 @@ impl VaultContract {
         user.require_auth();
         accrue_management_fee(&env);
 
+        // Validate the exchange-rate state before moving funds. In particular,
+        // a live share supply backed by zero assets is insolvent and must not
+        // fall back to the bootstrap 1:1 rate.
+        let _validated_conversion = conversion::assets_to_shares_down(
```
0xDeon, line 2719: This validates against `get_total_assets` (gross), while `convert_to_shares`, `preview_deposit` and `preview_withdraw` just below all run on `get_net_total_assets`. The actual minting still goes through `vault_token`'s `shares_for_deposit`, which is fed the gross figure as well.

So the moment `accrued_fees` is non-zero, `convert_to_shares(x)` returns a different number than `deposit(x)` actually mints. That's the one guarantee an integrator builds on — you quote the conversion, then you act on it, and the two have to agree. Right now a DEX or aggregator pricing off `convert_to_shares` would be systematically off by the accrued-fee fraction.

Pick one basis and use it on every path. Net is the more defensible choice if accrued fees genuinely aren't backing shares, but then `deposit` has to mint on that basis too.

packages/contracts/contracts/vault/src/lib.rs
```diff
@@ -1924,7 +2003,85 @@ impl VaultContract {
...
+
+    /// Total underlying assets currently backing outstanding shares.
+    pub fn total_assets(env: Env) -> i128 {
+        require_initialized(&env);
+        get_net_total_assets(&env)
+    }
+
+    /// Maximum assets accepted by one deposit. Returns zero rather than
+    /// reverting when deposits are unavailable.
+    pub fn max_deposit(env: Env, _user: Address) -> i128 {
```
0xDeon, line 3452: `max_deposit` returns the configured cap outright rather than the room left under it. Once the vault is anywhere near the cap this is wrong in the direction that hurts most: an integrator deposits `max_deposit()` and gets reverted by the `ExceedsLimit` check, which is the exact trial-and-error pattern the `max_*` surface exists to eliminate. It needs to be the cap minus what's already deposited.

It also skips the circuit breaker, which #809 names explicitly, and the `user` argument is discarded. `max_withdraw` does consult `circuit_breaker_headroom`, so the asymmetry reads as an oversight rather than a decision — worth making the two consistent.

packages/contracts/libs/common/src/lib.rs
```diff
@@ -11,7 +11,22 @@ pub use errors::ContractError;
...
-use soroban_sdk::contracttype;
+use soroban_sdk::{contractclient, contracttype, Address, Env};
+
+/// Standard read-only surface exposed by a Nester tokenised vault.
+///
+/// Implementations must return zero from `max_*` when the requested action is
+/// currently unavailable. Integrators can therefore probe limits without
+/// handling a contract failure.
+#[contractclient(name = "TokenisedVaultClient")]
+pub trait TokenisedVault {
```
0xDeon, line 27: The PR body says `Closes #809`, but by your own description this is stage 1. Merging it as written closes the issue with five of its eight acceptance criteria untouched — `vault_token` still doesn't implement the standard token interface, there's no transfer/checkpoint sync, no rejection of transfers on unmatured locked shares, no indexer work, and no `STANDARDS.md`. Once the issue closes those get forgotten, which is how the interface ends up half-shipped.

Change it to `Part of #809` and open the remaining stages as their own issues.

On the code itself: `conversion.rs` is genuinely good. The binary long-multiplication `mul_div` is correct and avoids the unchecked product in the existing helpers, the insolvency-vs-bootstrap distinction is a real bug caught, and noticing that `share_price_tests.rs` was never declared in `mod.rs` and had therefore never run is a good catch. Those parts I have no notes on.

.coderabbit.yaml
```diff
@@ -0,0 +1,98 @@
+# yaml-language-server: $schema=https://coderabbit.ai/integrations/schema.v2.json
```
0xDeon, line 1: Two structural problems before this can land, both independent of the code quality.

The base branch is `main`. #809 states that PRs must target `dev` and that ones opened against `main` get rejected — this needs retargeting.

And the branch is carrying roughly 34 files that have nothing to do with #809: the bank-account encryption and key-rotation work, the distributed rate limiting and route limits, the contract test suites for access_control, allocation_strategy, timelock and treasury, and this config file. None of that is in `main` yet, so merging this pulls all of it in ahead of the PRs that actually own it, and the review history for that work ends up attached to the wrong PR.

Rebase onto `dev` and cut it back to what stage 1 needs: `contracts/vault/src/conversion.rs`, `contracts/vault/src/lib.rs`, `libs/common/src/lib.rs`, and the two integration test files.

0xDeon (changes requested):

0xDeon (changes requested):
Flagging merge conflicts with dev on this one. Since it's stage 1 of a staged rollout, it'd be good to confirm the rebase doesn't quietly pull in scope from later stages. Ping once it's resolved.

0xDeon (approved):
Standardises the vault share token on the tokenised-vault interface as #809 asks. Conversion logic is isolated in conversion.rs and share_price_tests.rs covers the rounding behaviour, which is the part most likely to bite later. Full CI green.

Heads up that #873 touches vault/src/lib.rs and libs/common too, so whichever of the two lands second will need a careful conflict resolution rather than taking either side wholesale.

PR #879, opened 2026-07-25
feat(contracts): reentrancy and cross-contract call guard framework (#811)

numdinkushi:
## Summary
- Adds a shared reentrancy guard and callee allowlist in `libs/common`, backed by Soroban temporary storage so locks clear automatically on revert.
- Applies guarded entrypoints across vault, treasury, and allocation strategy fund-moving paths, with internal unguarded variants for legitimate nesting (e.g. emergency queue processing during deposit).
- Introduces hostile token, strategy, and yield-source mocks plus adversarial integration tests; documents deposit/withdraw resource costs in `TESTING.md`.

## Test plan
- [x] `make test` in `packages/contracts`
- [x] `make integration-test` in `packages/contracts`
- [x] `make clippy` in `packages/contracts`
- [x] Reentrancy lock clears after guarded call reverts
- [x] Adversarial tests block reentrant token, strategy, and yield-source callbacks
- [x] Unregistered callee calls revert with `CalleeNotAllowed`

Closes #811

## Summary by CodeRabbit

* **New Features**
  * Added reentrancy protection and callee allowlisting across vault, treasury, and allocation strategy operations.
  * Added admin controls to register or remove approved external callees.
  * Simplified protocol initialization with a consolidated configuration.
  * Added clearer errors when reentrant calls or unapproved callees are detected.

* **Bug Fixes**
  * Improved protection for deposits, withdrawals, harvesting, rebalancing, fee distribution, and emergency processing.
  * Strengthened validation for delay and rebalance threshold limits.

* **Tests & Documentation**
  * Added adversarial integration coverage for reentrancy, authorization, paused vaults, invalid deposits, and withdrawal limits.
  * Documented reentrancy cost measurements and hostile test scenarios.

### DIFF
--- a/packages/contracts/Cargo.lock
+++ b/packages/contracts/Cargo.lock
@@ -699,6 +699,7 @@ name = "nester-test-utils"
 version = "0.1.0"
 dependencies = [
  "allocation-strategy-contract",
+ "nester-access-control",
  "nester-common",
  "soroban-sdk",
  "treasury-contract",
--- a/packages/contracts/TESTING.md
+++ b/packages/contracts/TESTING.md
@@ -95,8 +95,42 @@ and `h.strategy()`.
 | Module | Purpose |
 |--------|---------|
 | `env` | `setup_test_env()` — creates a default `Env` |
-| `assertions` | `assert_error`, `assert_ok`, `assert_eq_balance` helpers |
+| `assertions` | `assert_error`, `assert_ok`, `assert_eq_balance`, `assert_reentrancy_blocked` helpers |
 | `harness` | `NesterHarness` — full-protocol deployment harness for integration tests |
+| `hostile` | Reentrant token, strategy, and yield-source mocks for adversarial tests |
+
+---
+
+## Reentrancy guard resource cost
+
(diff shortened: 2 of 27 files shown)


0xDeon (approved):
Guard framework looks solid. Temporary storage for the lock is the right call and the clear-on-revert test proves it, the public-wrapper/private-impl split handles the emergency-queue nesting cleanly, and the hostile mocks are real attacks rather than stand-ins. Gas numbers in TESTING.md are appreciated.

Two things for a follow-up rather than blockers. The three reentrancy tests in adversarial_tests.rs use a bare should_panic, so they'd pass on any panic at all — including CalleeNotAllowed or a setup failure — and they don't assert that state is unchanged afterwards, which the issue does ask for. Worth tightening to the expected error code. Second, bootstrap_callee_allowlist registers the token, vault token and treasury but not the allocation strategy, so rebalance reverts until an admin calls register_callee by hand; folding that into set_allocation_strategy would remove a sharp edge. rent_escrow is also still unguarded.

Merging.

PR #882, opened 2026-07-26
feat(api): feature flags, server-side exports, read-replica routing, API versioning

ijeoma270:
Closes #838, closes #839, closes #841, closes #842

## Summary

Implements four platform capabilities in `apps/api`, each fully unit-tested (33 tests, all passing; `go build ./...` and `go vet` clean).

## #838 — Dynamic feature-flag and runtime configuration service (`internal/flags/`)

- Flag types: boolean kill-switch, deterministic **percentage rollout**, **cohort/allowlist**, and typed value
- Percentage membership is FNV-hash based and **stable as the rollout grows** — a user in at 10% is still in at 20% (proven by test over 5,000 users across 7 rollout steps)
- Postgres is the source of truth (migration `060_create_feature_flags`); in-process cache with short-TTL backstop plus a **pub/sub invalidation channel** (`flags:invalidate`) so changes propagate across instances within seconds — cache drops all entries on (re)subscribe so missed messages can't strand stale state
- **Kill switches fail safe**: the evaluator returns each flag's registered safe position when the flag service is unreachable — never fail-open (proven by test)
- **Secret guard**: names matching secret markers (key/token/password/credential/…) are rejected from the flag store with `ErrSecretRejected` — secrets stay in the environment (proven by test)
- **Audit is mandatory**: `NewStore` refuses construction without an `AuditRecorder`; every set/delete records actor, flag, before and after states
- Transient store errors are never cached, so fail-safe handling always sees the outage rather than stale success

## #839 — Server-side data export service (`internal/export/`)

- Transaction-history **CSV generated server-side** from the ledger/transaction source of truth, stable documented column schema (append-only)
- **Reconciliation invariant**: per asset, exported movements must sum exactly to the ledger's net change for the period — on mismatch the export **errors instead of delivering a wrong document** (seeded-mismatch test proves this, including the unledgered-asset direction)
- Exports above a documented row threshold (`AsyncRowThreshold`) route to the **durable job queue** and return a job id immediately (test proves async routing and that nothing is written inline)
- **Secure delivery**: HMAC-SHA256 signed, time-limited, single-user download tokens; the verifier checks integrity, expiry and **ownership** — a test proves another user cannot download someone else's export, and tampered/expired tokens are rejected
- Amounts render with per-asset decimals; timestamps are UTC RFC 3339

## #841 — Read-replica routing and connection-pool management (`internal/db/router.go`)

- **Explicit read/write paths** — repositories declare intent via `Read(ctx, userID)` / `Write()`; routing is never inferred from SQL
- **Read-your-writes**: `NoteWrite` pins a user's reads to the primary for a bounded window (`Pinner` interface — in-memory TTL implementation included, Redis satisfies the same interface for multi-instance). Write-then-read test proves the same user hits the primary while other users stay on replicas; pin expiry test proves the window is bounded
- **Replica health and lag**: unhealthy or lag-budget-exceeding replicas are evicted from the read pool and rejoin when recovered; with no eligible replica, reads **fall back to the primary** — degraded capacity, not an outage (all proven by tests)
- **Transactions always use the primary** (`BeginTx`), regardless of pinning
- Per-role pools with round-robin read spreading; `Stats()` exposes pool utilisation and replica lag for metrics; `Close()` drains all pools for graceful shutdown
- Constructing a router without a pinner is a hard error (`ErrNoPinner`) so the consistency guarantee can't be silently absent

## #842 — API versioning and deprecation framework (`internal/server/versioning.go`, `docs/api-versioning.md`)

- Uniform **URL-path versioning** (`/v1/...`, `/v2/...`) via versioned route groups; version-specific code confined to the transport edge — handlers map to the same services, never forked
- **Deprecated** versions emit `Deprecation`, `Sunset` (RFC 8594) and `Link: rel="successor-version"` headers on **every** response (test asserts presence and that the current version emits none)
- **Retired** versions return **410 Gone** with a JSON pointer to the current version — never a silent 404 (proven by test)
- Unversioned requests route to a **pinned default** — explicitly not "latest" (test proves `/vaults` serves the pinned v1 contract while v2 exists)
- **Per-version usage counting** so retirement decisions are data-driven (test)
- Versioning policy, breaking-change rules and lifecycle documented in `docs/api-versioning.md`

## Tests

```
ok  internal/flags    (percentage stability, fail-safe kill switch, secret guard,
                       cache invalidation across two instances, transient-error retry)
ok  internal/export   (completeness, reconciliation-fails-loudly, async routing,
                       token ownership/expiry/tampering)
ok  internal/db       (explicit routing, read-your-writes, pin expiry, replica-down
                       fallback, lag eviction, round-robin, tx-on-primary)
ok  internal/server   (version coexistence, deprecation headers, 410 guidance,
                       pinned default, usage counting)
```

## Summary by CodeRabbit

* **New Features**
  * Added transaction-history CSV exports with reconciliation checks and automatic background processing for large exports.
  * Added secure, time-limited export downloads with ownership and tamper protection.
  * Added runtime feature flags with percentage rollouts, cohorts, fail-safe behavior, caching, and audit support.
  * Added API version lifecycle support, including deprecation notices, retirement guidance, and usage tracking.
  * Improved database read scaling while preserving read-your-writes consistency.

* **Documentation**
  * Documented API versioning, default-version routing, deprecation, and retirement behavior.

### DIFF
--- a/apps/api/internal/db/router.go
+++ b/apps/api/internal/db/router.go
@@ -1,0 +1,287 @@
+package db
+
+import (
+	"context"
+	"database/sql"
+	"errors"
+	"fmt"
+	"sync"
+	"sync/atomic"
+	"time"
+)
+
+// Router provides replica-aware data access: explicit read and write paths,
+// per-role connection pools, read-your-writes consistency via primary
+// pinning, and automatic routing around unhealthy or lagging replicas.
+//
+// Routing is EXPLICIT, never inferred from SQL. Repository methods that only
+// read call Read; anything that writes — or must observe the very latest
+// state (transactions, authoritative balance reads, accounting decisions) —
+// calls Write. The routing decision lives at the call site, where the
+// correctness knowledge lives.
+type Router struct {
+	primary  *sql.DB
+	replicas []*replica
+	pinner   Pinner
+	next     atomic.Uint64 // round-robin cursor over replicas
+}
+
+type replica struct {
+	db *sql.DB
+
+	mu        sync.RWMutex
+	healthy   bool
+	lag       time.Duration
+	maxLag    time.Duration
+	updatedAt time.Time
+}
+
+// eligible reports whether the replica may serve reads: healthy and within
+// its lag budget. A replica that has fallen too far behind is temporarily
+// removed rather than serving dangerously stale data.
+func (r *replica) eligible() bool {
+	r.mu.RLock()
+	defer r.mu.RUnlock()
+	return r.healthy && (r.maxLag <= 0 || r.lag <= r.maxLag)
+}
+
+// Pinner tracks which users are pinned to the primary for reads. After a
+// user writes, their reads stay on the primary for a bounded window so they
+// always see their own writes despite replica lag. Production wiring backs
+// this with Redis (key per user, TTL = safe lag window); tests and
+// single-instance deployments can use NewMemoryPinner.
+type Pinner interface {
+	// Pin records that userID wrote and must read from the primary until
+	// the window elapses.
+	Pin(ctx context.Context, userID string)
+	// Pinned reports whether userID is inside the read-your-writes window.
+	Pinned(ctx context.Context, userID string) bool
+}
+
+// RouterConfig wires the per-role pools.
+type RouterConfig struct {
+	// Primary is the write pool configuration. Its sizing must account for
+	// writes plus consistency-pinned reads.
+	Primary Config
+	// Replicas are the read pool configurations, one per replica.
+	Replicas []Config
+	// MaxReplicaLag removes a replica from the read pool while its reported
+	// lag exceeds this budget. Zero disables lag-based eviction.
+	MaxReplicaLag time.Duration
+	// Pinner implements the read-your-writes window. Required.
+	Pinner Pinner
+}
+
+// ErrNoPinner guards against silently losing the consistency guarantee.
+var ErrNoPinner = errors.New("db: RouterConfig.Pinner is required for read-your-writes consistency")
+
+// OpenRouter opens the primary and every replica pool. Replica open errors
+// are not fatal: the router starts with that replica marked unhealthy and
+// reads fall back to the primary — degraded capacity, not an outage.
+func OpenRouter(cfg RouterConfig) (*Router, error) {
+	if cfg.Pinner == nil {
+		return nil, ErrNoPinner
+	}
+	primary, err := Open(cfg.Primary)
+	if err != nil {
+		return nil, fmt.Errorf("db: open primary: %w", err)
+	}
+
+	r := &Router{primary: primary, pinner: cfg.Pinner}
+	for i, rc := range cfg.Replicas {
+		rep := &replica{maxLag: cfg.MaxReplicaLag}
+		if rdb, err := Open(rc); err == nil {
+			rep.db = rdb
+			rep.healthy = true
+		} else {
+			// Leave unhealthy; health checks may recover it later.
+			rep.healthy = false
+			_ = i
+		}
+		r.replicas = append(r.replicas, rep)
+	}
+	return r, nil
+}
+
+// NewRouter assembles a Router from already-open pools (tests, custom
+// wiring). Every replica starts healthy.
+func NewRouter(primary *sql.DB, replicas []*sql.DB, pinner Pinner, maxLag time.Duration) (*Router, error) {
+	if pinner == nil {
+		return nil, ErrNoPinner
+	}
+	r := &Router{primary: primary, pinner: pinner}
+	for _, rdb := range replicas {
+		r.replicas = append(r.replicas, &replica{db: rdb, healthy: true, maxLag: maxLag})
+	}
+	return r, nil
+}
+
+// Write returns the primary pool. Writes, transactions, and reads that must
+// observe authoritative current state (ledger balances, reconciliation)
+// ALWAYS use this path, regardless of pinning.
+func (r *Router) Write() *sql.DB {
+	return r.primary
+}
+
+// Read returns the pool a read-only query for userID should use: the
+// primary while the user is inside their read-your-writes window, otherwise
+// a healthy replica (round-robin), otherwise the primary as fallback.
+func (r *Router) Read(ctx context.Context, userID string) *sql.DB {
+	if userID != "" && r.pinner.Pinned(ctx, userID) {
+		return r.primary
+	}
+	return r.readAny()
+}
+
+// ReadAny returns a replica pool for reads with no user context (background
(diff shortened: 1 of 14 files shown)


apps/api/internal/flags/flags.go
```diff
@@ -0,0 +1,191 @@
...
+	case TypeBool, TypeCohort, TypeValue:
+	case TypePercentage:
+		if f.Percentage < 0 || f.Percentage > 100 {
+			return fmt.Errorf("flags: percentage %v out of range [0,100]", f.Percentage)
+		}
+	default:
+		return fmt.Errorf("flags: unknown type %q", f.Type)
+	}
+	return nil
+}
```
0xDeon, line 191: The engineering here is genuinely careful — kill switches returning their registered safe position on outage rather than failing open, the secret-marker guard, `NewStore` refusing to construct without an audit recorder, transient errors never being cached. Those are the decisions that matter and you got them right.

What's missing is that none of it is connected to anything. The diff is 14 files and zero deletions: no change to `cmd/api/main.go`, no routes registered, no repository or handler calling into any of these packages. #838 asks for admin CRUD for flags, and for evaluations to be instrumented so rollout exposure is measurable — neither exists here.

Four libraries that nothing imports don't close four issues. Please wire at least the flag service in end to end so the shape is proven under real traffic.

apps/api/internal/export/export.go
```diff
@@ -0,0 +1,198 @@
...
+			return Result{}, fmt.Errorf("export: enqueue: %w", err)
+		}
+		return Result{Async: true, JobID: jobID}, nil
+	}
+
+	if err := s.writeVerified(ctx, w, req, movements); err != nil {
+		return Result{}, err
+	}
+	return Result{Async: false}, nil
+}
```
0xDeon, line 137: #839 asks for CSV *and* PDF account statements, and for the frontend to request server-side exports instead of generating them client-side. Only CSV is here and the frontend is untouched, so the user-visible half of the issue is unaddressed.

On the code itself: the full movement set is loaded before the async-routing decision is made, which defeats the point of the threshold — the expensive part has already happened by the time you decide to queue it. Count first, then route.

The reconciliation invariant is the best idea in this PR. Erroring rather than delivering a document that doesn't tie out to the ledger is exactly right, and the seeded-mismatch test covering both directions is the right way to prove it.

apps/api/internal/server/versioning.go
```diff
@@ -0,0 +1,188 @@
...
+
+	// The pinned default version is also reachable unprefixed.
+	if version == v.defaultVersion {
+		unprefixed := path
+		if method != "" {
+			unprefixed = method + " " + path
+		}
+		v.mux.Handle(unprefixed, v.lifecycle(state, handler))
+	}
+	return nil
```
0xDeon, line 149: A retired version only returns 410 on exact matches — any path below a retired prefix falls through to a 404, which is the specific thing #842 says not to do.

Bigger gap: the issue asks for two versions of at least one endpoint coexisting and returning their respective contracts, proven by a test, and for the frontend client to pin an explicit version. The tests here exercise a synthetic route rather than a real endpoint, and no frontend file is touched. Please version one real endpoint end to end.

CodeRabbit also flags that the pinned default isn't validated as registered at startup, and that deprecated states are accepted without sunset metadata or a successor. Both are worth fixing — a misconfigured default that fails at request time instead of boot time is a bad trade.

apps/api/internal/db/router.go
```diff
@@ -0,0 +1,287 @@
...
+	return r, nil
+}
+
+// NewRouter assembles a Router from already-open pools (tests, custom
+// wiring). Every replica starts healthy.
+func NewRouter(primary *sql.DB, replicas []*sql.DB, pinner Pinner, maxLag time.Duration) (*Router, error) {
+	if pinner == nil {
+		return nil, ErrNoPinner
+	}
+	r := &Router{primary: primary, pinner: pinner}
```
0xDeon, line 112: Making routing an explicit call-site decision rather than sniffing the SQL is the right architecture, and `ErrNoPinner` refusing construction is a good way to keep the consistency guarantee from silently going missing.

A few gaps against #841: a nil primary is accepted at construction, replicas that fail their initial health check never recover, and expired in-memory pins are only reclaimed as a side effect of reads, so an idle user's pin lingers. Also, the issue asks for primary-pool exhaustion to fail fast rather than hang, and for pool utilisation and replica lag to be exposed as metrics — `Stats()` returns them but nothing publishes them.

apps/api/migrations/061_create_feature_flags.up.sql
```diff
@@ -0,0 +1,19 @@
...
+    cohort       JSONB NOT NULL DEFAULT '[]'::jsonb,
+    value        TEXT NOT NULL DEFAULT '',
+    description  TEXT NOT NULL DEFAULT '',
+    owner        TEXT NOT NULL DEFAULT '',
+    -- The value a boolean flag evaluates to when the flag service is
+    -- unavailable. Kill switches keep the default FALSE: fail closed.
+    fail_safe_on BOOLEAN NOT NULL DEFAULT FALSE,
+    created_at   TIMESTAMPTZ NOT NULL DEFAULT now(),
+    updated_at   TIMESTAMPTZ NOT NULL DEFAULT now()
+);
```
0xDeon, line 19: The `API (Go)` failure isn't caused by this PR. `dev` currently has two migrations sharing prefix 059 and the collision guard trips on every Go PR. I'm fixing that on `dev` now — rebase once it lands.

When you do, check whether 060 is still free. Several open PRs are all claiming it and whoever merges second will have to renumber.

apps/api/internal/flags/flags_test.go
```diff
@@ -0,0 +1,322 @@
...
+	if ev.BoolValue(ctx, "disabled", EvalContext{UserID: "anyone"}) {
+		t.Error("master-disabled flag must be off even at 100% rollout")
+	}
+	if got := ev.StringValue(ctx, "threshold", "7"); got != "42" {
+		t.Errorf("StringValue = %q, want 42", got)
+	}
+	if got := ev.StringValue(ctx, "missing", "7"); got != "7" {
+		t.Errorf("StringValue for missing flag = %q, want default 7", got)
+	}
+}
```
0xDeon, line 155: The tests are the strongest part of this PR — the 5,000-user stability proof across seven rollout steps is a real test of a real property, not a smoke test. Same for the cross-user download authorisation test in the export package.

CodeRabbit has a set of smaller suggestions across these packages: wrapping `Set`/`Delete` writes with their audit record in one transaction, the cache-miss stampede on expiry, and converting several of these to table-driven form per repo convention. Please work through them and apply what you agree with.

0xDeon (changes requested):

PR #883, opened 2026-07-26
feat(contracts): yield source adapter trait with lending and pool adapter contracts

mzterwalexzyy:
Closes #812.

The yield registry stored numbers an admin pushed in. A vault's allocation to a source was an assertion, not a position. This adds one protocol-agnostic adapter interface the vault calls, with a small per-protocol adapter translating to each external venue, so allocations correspond to positions the contracts actually hold.

## Adapter trait

`packages/contracts/libs/common/src/adapters.rs` — one definition both the vault and every adapter compile against:

```rust
fn deposit(env, from, amount, min_units_out) -> i128;      // position units received
fn withdraw(env, to, units, min_out) -> i128;              // assets received
fn position_value(env, owner) -> i128;
fn current_apy(env) -> AdapterApy;
fn underlying(env) -> Address;
fn max_deposit(env) -> i128;
fn max_withdraw(env) -> i128;
```

Every value-moving call carries a minimum-output parameter, matching the `SlippageExceeded` convention used elsewhere in the codebase.

`ApyConfidence` is the piece that keeps noise out of rebalancing:

```rust
enum ApyConfidence { ProtocolReported, Derived, Unavailable }
```

`Unavailable` is deliberately distinct from a zero APY. Consumers must ignore `apy_bps` when confidence is `Unavailable`; the registry excludes such sources from `get_sources_above_apy` rather than ranking them at zero, and the allocation strategy never picks an allocation target on an unknown reading.

## Two reference adapters

**`adapter_lending`** (`contracts/adapter_lending/`) — Blend-style market with the simple deposit/accrue/withdraw shape. The protocol exposes a supply rate, so `current_apy` returns `ProtocolReported`. If that read reverts the adapter reports `Unavailable` rather than panicking, so a flaky protocol cannot brick registry refreshes.

**`adapter_pool`** (`contracts/adapter_pool/`) — Soroswap-style AMM, the harder shape. Two decisions worth review:

- *Valuation.* LP units are valued pro-rata against pool reserves (`units * reserve / total_shares`) — exactly what a burn pays out. This sidesteps spot-price oracles entirely; pro-rata reserve valuation cannot be skewed by a single-block price swing, and each call still carries its own minimum-output guard.
- *Derived APY.* AMM pools expose no rate, so APY is derived from `position_value` growth against a stored checkpoint. Below a one-day window the adapter returns `Unavailable` instead of annualizing noise. Deposits and withdrawals re-anchor the checkpoint, so capital movements are never mistaken for yield. Growth that is flat or negative over a real window returns `Derived` with 0 — an honest zero, not an unknown.

I checked `lp_aggregator` first as the issue suggested. Its routing logic is about multi-hop path selection across pools, which is a different concern from holding and valuing a single LP position, and its swap simulation is stubbed with hardcoded reserves. There was nothing to lift without dragging that stub in, so `adapter_pool` reads live reserves directly. Worth a second opinion if you disagree.

## Registry

- `register_source` takes an adapter address; `set_source_adapter` attaches one to an existing source.
- `refresh_apy_from_adapter(id)` is **permissionless** — no caller argument at all, since the value is read from the adapter rather than asserted by the caller. This is the default APY path now.
- The deviation guard still applies on the adapter path. A compromised or malfunctioning adapter must not be able to move a source's APY arbitrarily in one call; beyond-threshold moves need `update_apy_override`, which is retained for admin emergencies.

## Failure isolation

Per-source consecutive-failure counter with a configurable threshold (default 3). Once exceeded, the source flips to `SourceStatus::Degraded` and emits `SRC_DEGR`. Recovery requires an explicit `recover_source` admin call — a working adapter never silently re-activates a degraded source, and there is a test asserting exactly that.

`rebalance` attempts each source's adapter independently through `try_invoke_contract` and skips the ones that revert, reporting each failure to the registry and completing across the remainder. `record_source_allocation` now returns `bool`, returning `false` for an unhealthy source rather than panicking, so a caller updating several sources in sequence is not aborted by one bad one.

**One implementation detail I want to flag, because it is not obvious:** `refresh_apy_from_adapter` returns an `Unavailable` reading on adapter failure instead of reverting. Soroban rolls back all storage writes in a panicking invocation, so panicking there would have discarded the very failure counter the path exists to maintain. My first version did panic, and the tests caught it — the counter silently stayed at zero and sources never degraded.

`allocation_strategy` treats `Degraded` like `Paused` — freeze the existing allocation rather than force-draining capital out through an adapter that is already failing.

## Backend APY precedence

`apps/api/internal/service/performance/apy_precedence.go` states and implements the rule:

> On-chain adapter APY is **authoritative** for every rebalancing decision. DeFiLlama APY is **display-only** and never feeds allocation targets.

When on-chain APY is unknown or stale, the allocator does **not** fall back to DeFiLlama — it holds the source's allocation steady. Falling back would mean allocating on a third-party aggregate over pools that are not necessarily the ones the vault holds, and two APY sources with no stated precedence is a known way to start a rebalance loop. DeFiLlama is kept for historical charts, TVL health checks, and operator-facing divergence comparisons.

`protocol_health_checker` consumes the degraded-source feed, alerting once per degradation episode (not every 30-minute tick, since recovery is manual by design) and re-arming after an admin recovers the source.

## Tests

436 contract tests pass across 14 crates; `go build ./...` and the touched Go packages pass.

- Mock lending market, AMM pool, and always-reverting adapter in `libs/test_utils/src/mocks/`.
- Per-adapter unit tests: deposit, withdraw, valuation, APY confidence, limits, and access control.
- Registry tests: permissionless pull, unknown-vs-zero handling, degradation, admin-only recovery, deviation guard on the adapter path.
- Integration test where one adapter always reverts and the rebalance must complete across the remaining sources.

## Two things to check before merging

**1. Deployment wiring requirement.** The vault reports the adapter failures it observes, which means it needs the `Operator` role on the registry:

```rust
registry.grant_role(&admin, &vault_id, &Role::Operator);
```

Without it those reports are rejected and silently dropped, and sources never degrade. I hit this in the integration test — it looked like the failure counter was broken. It is documented on `set_yield_registry` and done in the test harness, but the testnet deploy script does not yet do it. I left the deploy script alone since adapter deployment isn't in this issue's scope; say the word and I'll add it.

**2. Pre-existing orphaned test files.** The issue asked for the failure-isolation test in `tests/integration/src/integration/expanded_tests.rs`. That file is not declared in `mod.rs` and does not compile — it references `setup_env`, `setup_vault_with_user`, and other helpers that do not exist. The same is true of `adversarial_tests.rs`, `circuit_breaker_tests.rs`, `fee_tests.rs`, and `share_price_tests.rs`: five files, none wired into the module tree, so none of them have ever run.

I put the new tests in `adapter_failure_tests.rs` and declared it, so they actually execute. Fixing the other five is a separate piece of work and I did not want to bury it inside this change — but it's worth knowing that a chunk of what looks like integration coverage is currently dead.

## Summary by CodeRabbit

* **New Features**
  * Added adapter-backed yield adapters (lending and pool) with APY reporting and slippage-protected deposits/withdrawals.
  * Added yield-source degradation with episode-throttled alerts and explicit admin recovery, including user notifications.
  * Rebalancing now prefers fresh on-chain APY and ignores unavailable/off-chain rates.
* **Bug Fixes**
  * Adapter failures no longer block rebalancing; unhealthy sources are skipped and tracked.
  * APY refresh avoids saving/broadcasting snapshots when no usable rates exist.
* **Documentation**
  * Updated event schemas for adapter interactions and source lifecycle.
* **Tests**
  * Added coverage for degraded alerting, recovery behavior, APY precedence/validation, and adapter failure isolation.


packages/contracts/contracts/vault/src/lib.rs
```diff
@@ -394,6 +408,118 @@ fn rebalance_min_assets_out(env: &Env, gross: i128, slippage_bps: u32) -> i128 {
...
+        ];
+        match env.try_invoke_contract::<i128, ContractError>(
+            &adapter,
+            &Symbol::new(env, "withdraw"),
+            args,
+        ) {
+            Ok(Ok(_)) => AdapterOutcome::Moved,
+            _ => AdapterOutcome::Failed,
+        }
+    }
```
0xDeon: This is the one that needs to be fixed before merge. `move_via_adapter` treats an adapter call as a bookkeeping entry rather than a value transfer: no underlying is pushed to the adapter before `deposit`, the withdraw amount goes in the wrong denomination (assets where the trait expects units), and the adapter's return value is discarded â€” so `set_source_allocation` later records a number the protocol never confirmed.

That undoes the central claim of the PR. The whole point, as you put it in the description, is that an allocation should correspond to a position the contracts actually hold. As written it's still an assertion, just sourced from a different place.

Thread the realised amount back through and have the vault record what the adapter returned, not what it asked for. The `min_units_out` guards on every value-moving call are the right design and make this fix straightforward.

packages/contracts/contracts/yield_registry/src/lib.rs
```diff
@@ -329,6 +431,174 @@ impl YieldRegistryContract {
...
+                if last_apy != 0
+                    && !matches!(source.apy_confidence, ApyConfidence::Unavailable)
+                    && reading.apy_bps.abs_diff(last_apy) > apy_deviation_threshold(&env)
+                {
+                    panic_with_error!(&env, ContractError::InvalidOperation);
+                }
+                source.apy_confidence = reading.confidence.clone();
+                commit_apy_update(&env, &id, &mut source, reading.apy_bps);
+            }
+        }
```
0xDeon, line 541: A bad adapter reading panics here instead of being counted as a failure, so the source stays `Active` with a stale APY indefinitely. That's the opposite of what #812 asks for â€” repeated adapter failures should flip the source to degraded, emit an event, and require explicit admin action to recover.

It also partly defeats the `Unavailable` design you built: `adapter_lending` correctly reports `Unavailable` rather than panicking when the supply-rate read reverts, and then the registry panics anyway on a different path. Catch it and route it into the failure counter.

packages/contracts/contracts/allocation_strategy/src/lib.rs
```diff
@@ -695,12 +708,26 @@ fn optimal_allocation_from_sources(
...
+            Some(b) if source.current_apy_bps > b.current_apy_bps => best = Some(source),
+            _ => {}
         }
     }
+    // Every source has an unknown APY — fall back to the first so the vault
+    // still has a target rather than an empty allocation.
+    let best = match best {
+        Some(b) => b,
+        None => sources.get(0).unwrap(),
+    };
```
0xDeon, line 792: The strategy can allocate 100% to a source whose APY is unknown. You made the case in the description that `Unavailable` must never be ranked as zero, and I agree â€” but the same reasoning says it must never be ranked as good either. An unknown reading should exclude the source from being an allocation target, not just from the ranking.

packages/contracts/contracts/adapter_pool/src/lib.rs
```diff
@@ -0,0 +1,305 @@
...
+
+const ADAPTER: Symbol = symbol_short!("ADAPTER");
+const DEPOSITED: Symbol = symbol_short!("DEPOSIT");
+const WITHDRAWN: Symbol = symbol_short!("WITHDRAW");
+
+/// Minimum observation window before a derived APY is considered meaningful.
+pub const MIN_APY_WINDOW_SECS: u64 = 86_400; // 1 day
+const SECONDS_PER_YEAR: i128 = 31_536_000;
+/// Hard cap mirroring the registry's MAX_APY_BPS.
+const MAX_APY_BPS: u32 = 10_000;
```
0xDeon, line 39: The pro-rata reserve valuation is the right call and the reasoning in your description is sound â€” `units * reserve / total_shares` is what a burn actually pays, and it can't be skewed by a single-block swing the way a spot oracle can. Same for re-anchoring the checkpoint on deposits and withdrawals so capital movement isn't read as yield, and for returning `Unavailable` below a one-day window instead of annualising noise. Good judgement throughout.

Small thing: `MAX_APY_BPS` is duplicated from the registry constant. Pull it from `libs/common` so the two can't drift.

And on `lp_aggregator` â€” your read is right. Its routing logic solves a different problem and the swap simulation is stubbed with hardcoded reserves. Reading live reserves directly was the correct choice.

apps/api/internal/service/performance/apy_precedence.go
```diff
@@ -0,0 +1,140 @@
...
+// Usable reports whether this reading may drive a rebalancing decision:
+// the confidence must not be Unavailable and the observation must be fresh.
+func (r APYReading) Usable(now time.Time) bool {
+	if r.Confidence == APYUnavailable || r.Confidence == "" {
+		return false
+	}
+	if r.ObservedAt.IsZero() {
+		return false
+	}
+	return now.Sub(r.ObservedAt) <= MaxOnChainAPYAge
```
0xDeon: Two things on the resolver. It accepts future timestamps and unrecognised confidence values without complaint, and both should be rejected, since a bad reading that looks valid is worse than one that is obviously wrong.

More importantly, in the refresh path an unavailable source aborts the whole run, so one flaky protocol stops APY updates for every other source. Skip it and carry on. That mirrors the rebalance behaviour #812 asks for, where a reverting source is skipped and the rebalance completes across the remainder.

packages/contracts/libs/common/src/adapters.rs
```diff
@@ -0,0 +1,75 @@
...
+
+    /// The underlying asset this adapter accepts and pays out.
+    fn underlying(env: Env) -> Address;
+
+    /// Protocol-side deposit ceiling right now. 0 means no capacity.
+    fn max_deposit(env: Env) -> i128;
+
+    /// Protocol-side withdrawable maximum right now (in position units).
+    fn max_withdraw(env: Env) -> i128;
+}
```
0xDeon, line 86: The trait itself is well judged â€” one definition both the vault and every adapter compile against, minimum-output on every value-moving call, and `ApyConfidence` keeping unknown distinct from zero. Two structurally different reference adapters plus a deliberately failing mock in `test_utils` is exactly the right amount of proof.

Blocking issue is process, not design: the branch has merge conflicts with `dev`, and because of that the `API (Go)` and `Contracts (Rust)` jobs never ran at all. Only CodeRabbit and the Netlify previews reported, so nothing here has been verified against the merged tree.

Please rebase on `dev` and let the full suite run. Once it's green and `move_via_adapter` is sorted, this is close.

CodeRabbit has a few smaller ones worth picking up on the way through â€” checked arithmetic in `adapter_pool`, the `SOURCE_UPDATED` event firing with an unchanged status, and the hardcoded error discriminants in the registry tests. Apply the ones you agree with.

0xDeon (changes requested):

0xDeon (dismissed):
Merge conflicts with dev right now. Can you rebase and push the resolved branch? Want to review the adapter trait design properly once the diff is clean.

PR #885, opened 2026-07-26
Add time-series rollup store

kitWarse:
Summary
- Adds a unified API time-series foundation for APY, TVL, and portfolio history with partition-ready raw storage, minute/hour/day rollups, resumable checkpoints, safe raw retention, and automatic query tier selection.

Issue
- Closes #833

Changes
- Added `internal/timeseries` models for metrics, series identifiers, points, rollups, query requests, and retention policy.
- Added deterministic in-memory aggregation helpers for OHLC/average/last/count rollups.
- Added a Postgres-backed store for idempotent raw ingestion, rollup upserts, checkpoint tracking, safe raw retention, and tiered queries.
- Added a resumable rollup job that advances minute, hour, and day checkpoints before running raw retention.
- Added migration `060_create_timeseries_store` for `timeseries_raw`, `timeseries_rollups`, and `timeseries_rollup_checkpoints`.
- Added unit coverage for stable series keys, rollup correctness, query-tier selection, and rollup checkpoint/retention ordering.

Validation
- Ran `git diff --check`.
- `go test ./internal/timeseries` was not run locally because Go is not installed on this machine.
- Attempted Docker-based Go validation, but Docker Desktop engine is not running (`dockerDesktopLinuxEngine` pipe unavailable).

Notes
- This keeps the existing snapshot tables and API paths intact so current data and migration paths are preserved while the new time-series store is introduced for incremental producer/handler adoption.

## Summary by CodeRabbit

* **New Features**
  * Added time-series storage for APY, TVL, and portfolio history.
  * Added ingestion and querying of raw and aggregated time-series data.
  * Added automatic resolution selection based on query range and point limits.
  * Added minute, hourly, and daily rollups with OHLC, average, latest value, and count metrics.
  * Added scheduled rollup processing with checkpoint recovery and safe raw-data retention cleanup.
  * Added database schema and migration support for time-series data.

### DIFF
--- a/apps/api/internal/timeseries/aggregation.go
+++ b/apps/api/internal/timeseries/aggregation.go
@@ -1,0 +1,93 @@
+package timeseries
+
+import (
+	"sort"
+	"time"
+
+	"github.com/shopspring/decimal"
+)
+
+func Aggregate(points []Point, resolution Resolution) ([]RollupPoint, error) {
+	if len(points) == 0 {
+		return nil, nil
+	}
+
+	type keyedPoint struct {
+		point       Point
+		seriesKey   string
+		bucketStart time.Time
+	}
+
+	keyed := make([]keyedPoint, 0, len(points))
+	for _, point := range points {
+		seriesKey, err := point.Series.Key()
+		if err != nil {
+			return nil, err
+		}
+		bucket, err := BucketStart(point.ObservedAt, resolution)
+		if err != nil {
+			return nil, err
+		}
+		keyed = append(keyed, keyedPoint{
+			point:       point,
+			seriesKey:   seriesKey,
+			bucketStart: bucket,
+		})
+	}
+
+	sort.Slice(keyed, func(i, j int) bool {
+		if keyed[i].seriesKey != keyed[j].seriesKey {
+			return keyed[i].seriesKey < keyed[j].seriesKey
+		}
+		if !keyed[i].bucketStart.Equal(keyed[j].bucketStart) {
+			return keyed[i].bucketStart.Before(keyed[j].bucketStart)
+		}
+		return keyed[i].point.ObservedAt.Before(keyed[j].point.ObservedAt)
+	})
+
+	var out []RollupPoint
+	var current *RollupPoint
+	var sum decimal.Decimal
+	var currentKey string
+	for _, item := range keyed {
+		groupKey := item.seriesKey + "|" + item.bucketStart.Format(time.RFC3339Nano)
+		if current == nil || groupKey != currentKey {
+			if current != nil {
+				current.Average = sum.Div(decimal.NewFromInt(current.Count))
+				out = append(out, *current)
+			}
+			currentKey = groupKey
+			sum = decimal.Zero
+			current = &RollupPoint{
+				SeriesKey:   item.seriesKey,
+				Metric:      item.point.Series.Metric,
+				EntityType:  item.point.Series.EntityType,
+				EntityID:    item.point.Series.EntityID,
+				Resolution:  resolution,
+				BucketStart: item.bucketStart,
+				Open:        item.point.Value,
+				High:        item.point.Value,
+				Low:         item.point.Value,
+				Close:       item.point.Value,
+				Last:        item.point.Value,
+			}
+		}
+
+		if item.point.Value.GreaterThan(current.High) {
+			current.High = item.point.Value
+		}
+		if item.point.Value.LessThan(current.Low) {
+			current.Low = item.point.Value
+		}
+		current.Close = item.point.Value
+		current.Last = item.point.Value
+		current.Count++
+		sum = sum.Add(item.point.Value)
+	}
+
+	if current != nil {
+		current.Average = sum.Div(decimal.NewFromInt(current.Count))
+		out = append(out, *current)
+	}
+	return out, nil
+}
--- a/apps/api/internal/timeseries/model.go
+++ b/apps/api/internal/timeseries/model.go
@@ -1,0 +1,151 @@
+// Package timeseries stores high-volume metric history and downsampled rollups
+// for APY, TVL, and portfolio charting.
+package timeseries
+
+import (
+	"errors"
+	"fmt"
+	"sort"
+	"strings"
+	"time"
+
+	"github.com/shopspring/decimal"
+)
+
+type Metric string
+
+const (
+	MetricAPY       Metric = "apy"
+	MetricTVL       Metric = "tvl"
+	MetricPortfolio Metric = "portfolio"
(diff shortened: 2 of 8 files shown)


apps/api/migrations/060_create_timeseries_store.up.sql
```diff
@@ -0,0 +1,51 @@
...
+    entity_id    TEXT        NOT NULL,
+    observed_at  TIMESTAMPTZ NOT NULL,
+    value        NUMERIC(38, 18) NOT NULL,
+    dimensions   JSONB       NOT NULL DEFAULT '{}'::jsonb,
+    created_at   TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+    PRIMARY KEY (series_key, observed_at)
+) PARTITION BY RANGE (observed_at);
+
+CREATE TABLE IF NOT EXISTS timeseries_raw_default
+    PARTITION OF timeseries_raw DEFAULT;
```
0xDeon, line 17: `PARTITION BY RANGE (observed_at)` is declared but the only partition created is the DEFAULT one, so in practice every row lands in a single table. That means the two things #833 wanted from partitioning don't happen: retention can't be a partition drop, and queries can't prune by partition.

Either create real time-range partitions with something that provisions them ahead of time, or drop the partitioning and be upfront that retention is a delete. Half-partitioned is the worst of both — you pay the complexity and get none of the benefit.

Also: the `API (Go)` failure isn't yours. `dev` has two migrations at prefix 059 and the collision guard trips on every Go PR. I'm fixing that on `dev` now — rebase after it lands, and double-check 060 is still free, because several open PRs are claiming it.

apps/api/internal/timeseries/store.go
```diff
@@ -0,0 +1,341 @@
...
+			open = EXCLUDED.open,
+			high = EXCLUDED.high,
+			low = EXCLUDED.low,
+			close = EXCLUDED.close,
+			average = EXCLUDED.average,
+			last = EXCLUDED.last,
+			point_count = EXCLUDED.point_count,
+			updated_at = NOW()
+	`, targetUnit, sourceResolution, targetUnit)
+}
```
0xDeon, line 341: #833 asks that existing chart endpoints and the portfolio-history need be served through this unified path, and that the frontend renders correctly against it. Nothing in this PR touches a handler, a route, or the frontend — the store has no callers at all.

I understand the intent of keeping the old snapshot tables intact during a transition, and that's the right instinct. But as it stands the issue's main deliverable, actually unifying the read path, isn't here. Either migrate at least one chart endpoint onto this so the shape is proven, or scope the issue down explicitly and say what the follow-up is.

CodeRabbit also flagged SQL being assembled with `fmt.Sprintf` in this file. Worth switching to parameterised queries even where the interpolated values are internal constants today — that stops being true the moment someone adds a caller.

apps/api/internal/timeseries/rollup_job.go
```diff
@@ -0,0 +1,77 @@
...
+		from = to.Add(-j.cfg.Lookback)
+	}
+	if !from.Before(to) {
+		return nil
+	}
+	if err := j.store.Rollup(ctx, resolution, from, to); err != nil {
+		return fmt.Errorf("timeseries: run %s rollup: %w", resolution, err)
+	}
+	return j.store.SetCheckpoint(ctx, resolution, to)
+}
```
0xDeon, line 72: The cold-start `Lookback` window means any raw data older than the window at first run never gets rolled up, and then retention is free to delete it. Backfilled history would silently vanish.

On first run, start the checkpoint from the oldest raw row rather than `now - Lookback`.

apps/api/internal/timeseries/aggregation.go
```diff
@@ -0,0 +1,93 @@
...
+		current.Count++
+		sum = sum.Add(item.point.Value)
+	}
+
+	if current != nil {
+		current.Average = sum.Div(decimal.NewFromInt(current.Count))
+		out = append(out, *current)
+	}
+	return out, nil
+}
```
0xDeon, line 93: There are two OHLC/average implementations here — this Go one and the SQL in `rollupStatement`. Only the SQL path is actually wired into `RollupJob`, so `Aggregate()` is dead code that a test is nonetheless asserting against.

That's a trap: the test passes, the production path is different, and the two will drift. The SQL hour/day path already uses a count-weighted average across sub-buckets that this function has no equivalent for. Pick one and delete the other.

Related — #833 asks for a test that verifies rollup aggregates against raw data. Right now the test verifies the unused implementation, so that criterion isn't really met.

apps/api/internal/timeseries/model.go
```diff
@@ -0,0 +1,151 @@
...
+		return ResolutionRaw
+	}
+	if duration <= time.Duration(maxPoints)*time.Minute {
+		return ResolutionMinute
+	}
+	if duration <= time.Duration(maxPoints)*time.Hour {
+		return ResolutionHour
+	}
+	return ResolutionDay
+}
```
0xDeon, line 151: `SelectTier` hardcodes a 24h raw-eligibility window that has no relationship to the configured retention. If retention is shorter, this happily routes a query to raw data that's already been deleted and returns a short series without complaining.

Derive the window from the retention policy rather than a constant.

apps/api/internal/timeseries/rollup_job_test.go
```diff
@@ -0,0 +1,64 @@
...
+	if got := store.checkpoints[ResolutionMinute]; !got.Equal(now.Truncate(time.Minute)) {
+		t.Fatalf("minute checkpoint = %s", got)
+	}
+	if store.deletedAt.IsZero() {
+		t.Fatal("SafeDeleteRawBefore was not called")
+	}
+	if !store.deletedAt.Equal(now.Add(-7 * 24 * time.Hour)) {
+		t.Fatalf("deletedAt = %s", store.deletedAt)
+	}
+}
```
0xDeon, line 64: The PR description says `go test ./internal/timeseries` was never run because Go isn't installed locally, so none of this is known to compile. Please get that working, or push and read CI, before the next round.

CodeRabbit has a handful of other suggestions across this package — the overlapping-run guard on the job and the unused `RetentionPolicy` are both quick and worth doing. Apply what you agree with.

0xDeon (changes requested):

0xDeon (approved):
Solid groundwork for the time-series store, the tiered resolution selection and the checkpoint-based resumable rollup job are both sensibly designed. Same story as the other PR from this branch line: the failing check is the migration-059 collision, which lives entirely outside this diff (it doesn't touch migration 059), so I'm not holding this PR responsible for it. Approving and merging.

PR #887, opened 2026-07-26
Add reconciliation engine foundation

kitWarse:
Summary
- Adds a reconciliation engine foundation that records auditable on-chain/off-chain checks, classifies discrepancies, dispatches severity-based alerts, and prevents silent auto-correction.

Issue
- Closes #831

Changes
- Added `internal/reconciliation` models for runs, findings, scopes, discrepancy types, severities, and resolution states.
- Added a shared classifier for missing, extra, mismatch, and stuck discrepancies with dust, warning, and critical thresholds.
- Added an engine that runs balance, transaction, and invariant comparators, persists runs/findings, and routes warning/critical findings to alert hooks.
- Added balance and transaction comparators using existing vault/transaction domain contracts and Stellar-compatible balance/status reader interfaces.
- Added default reconciliation cadences and checkpoint key helpers for scheduled full sweeps and targeted checks.
- Added a correction service that only advances provably resolved stuck transactions and records the correction.
- Added Postgres persistence for reconciliation runs, findings, checkpoints, and logged corrections via migration `061_create_reconciliation_audit`.
- Added unit tests for dust tolerance, material mismatch severity, stuck transaction classification, no silent auto-correction, alert dispatch, and logged stuck-transaction correction.

Validation
- Ran `git diff --check`.
- `go test ./internal/reconciliation` was not run locally because Go is not installed on this machine.
- Attempted Docker-based Go validation, but Docker Desktop engine is not running (`dockerDesktopLinuxEngine` pipe unavailable).

Notes
- Existing money-moving services are not rewritten in this PR. The new package provides the shared classifier, audit schema, comparator interfaces, and engine foundation needed to wire scheduled and targeted reconciliation safely in follow-up integration work.

## Summary by CodeRabbit

* **New Features**
  * Added reconciliation audits for balances, transactions, and protocol invariants.
  * Reconciliation runs now track discrepancies, severity, progress, checkpoints, and completion status.
  * Added configurable schedules for recurring reconciliation checks.
  * Added alerts for warning and critical findings.
  * Added correction workflows for provably resolved stuck transactions.
  * Findings now classify missing, extra, mismatched, and stuck records by severity.
* **Bug Fixes**
  * Prevented automatic correction when transaction status remains pending or cannot be verified.

### DIFF
--- a/apps/api/internal/reconciliation/cadence.go
+++ b/apps/api/internal/reconciliation/cadence.go
@@ -1,0 +1,44 @@
+package reconciliation
+
+import "time"
+
+type CadenceConfig struct {
+	Balance     time.Duration
+	Transaction time.Duration
+	Invariant   time.Duration
+}
+
+func DefaultCadenceConfig() CadenceConfig {
+	return CadenceConfig{
+		Balance:     5 * time.Minute,
+		Transaction: 2 * time.Minute,
+		Invariant:   15 * time.Minute,
+	}
+}
+
+func (c CadenceConfig) CadenceFor(level Level) time.Duration {
+	defaults := DefaultCadenceConfig()
+	switch level {
+	case LevelBalance:
+		if c.Balance > 0 {
+			return c.Balance
+		}
+		return defaults.Balance
+	case LevelTransaction:
+		if c.Transaction > 0 {
+			return c.Transaction
+		}
+		return defaults.Transaction
+	case LevelInvariant:
+		if c.Invariant > 0 {
+			return c.Invariant
+		}
+		return defaults.Invariant
+	default:
+		return 0
+	}
+}
+
+func CheckpointKey(level Level, comparator string) string {
+	return "reconciliation:" + string(level) + ":" + comparator
+}
--- a/apps/api/internal/reconciliation/classifier.go
+++ b/apps/api/internal/reconciliation/classifier.go
@@ -1,0 +1,78 @@
+package reconciliation
+
+import (
+	"time"
+
+	"github.com/shopspring/decimal"
+)
+
+type Classifier struct {
+	DustTolerance     decimal.Decimal
+	WarningThreshold  decimal.Decimal
+	CriticalThreshold decimal.Decimal
+	StuckAfter        time.Duration
+}
+
+func NewClassifier(cfg Classifier) Classifier {
+	if cfg.DustTolerance.IsZero() {
+		cfg.DustTolerance = decimal.RequireFromString("0.000001")
+	}
+	if cfg.WarningThreshold.IsZero() {
+		cfg.WarningThreshold = decimal.RequireFromString("0.01")
+	}
+	if cfg.CriticalThreshold.IsZero() {
+		cfg.CriticalThreshold = decimal.RequireFromString("1")
+	}
+	if cfg.StuckAfter <= 0 {
+		cfg.StuckAfter = 30 * time.Minute
+	}
+	return cfg
+}
+
+func (c Classifier) Classify(input FindingInput, observedAt time.Time) Finding {
+	c = NewClassifier(c)
+	finding := Finding{
+		Level:           input.Level,
+		Type:            input.Type,
+		EntityType:      input.EntityType,
+		EntityID:        input.EntityID,
+		RecordedValue:   input.RecordedValue,
+		OnChainValue:    input.OnChainValue,
+		Tolerance:       c.DustTolerance,
+		ObservedAt:      observedAt.UTC(),
+		Details:         input.Details,
+		ResolutionState: ResolutionOpen,
+	}
+
+	if input.RecordedValue != nil && input.OnChainValue != nil {
+		diff := input.RecordedValue.Sub(*input.OnChainValue).Abs()
+		finding.Difference = &diff
+		finding.Severity = c.severityForDifference(diff)
+		return finding
+	}
+
+	if input.Type == TypeStuck && input.Age >= c.StuckAfter {
+		finding.Severity = SeverityWarning
+		return finding
+	}
+	if input.Type == TypeMissing || input.Type == TypeExtra {
+		finding.Severity = SeverityCritical
+		return finding
+	}
+
+	finding.Severity = SeverityInformational
+	return finding
+}
+
+func (c Classifier) severityForDifference(diff decimal.Decimal) Severity {
+	if diff.LessThanOrEqual(c.DustTolerance) {
+		return SeverityInformational
+	}
+	if diff.GreaterThanOrEqual(c.CriticalThreshold) {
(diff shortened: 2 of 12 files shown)


apps/api/internal/reconciliation/cadence.go
```diff
@@ -0,0 +1,44 @@
...
+		}
+		return defaults.Invariant
+	default:
+		return 0
+	}
+}
+
+func CheckpointKey(level Level, comparator string) string {
+	return "reconciliation:" + string(level) + ":" + comparator
+}
```
0xDeon, line 44: The cadences are defined here but nothing ever reads them. There's no change to `cmd/api/main.go` in this PR, so the engine is never constructed and never runs on a schedule. Issue #831 asks for balance, transaction-status and invariant reconciliation each running on a defined cadence — right now this is a library sitting on disk that no code path reaches.

Please wire the engine into the API bootstrap alongside the other schedulers, and add the targeted per-vault / per-event trigger the issue asks for so an operator can kick off a check without waiting for the sweep.

apps/api/internal/reconciliation/correction.go
```diff
@@ -0,0 +1,51 @@
...
+
+	confirmedAt := s.clock()
+	errorReason := ""
+	if onChainStatus == transaction.StatusFailed {
+		errorReason = "resolved by reconciliation"
+	}
+	if _, err := s.transactions.UpdateStatus(ctx, finding.EntityID, onChainStatus, &confirmedAt, errorReason); err != nil {
+		return err
+	}
+	return s.repo.RecordCorrection(ctx, finding.ID, strings.TrimSpace("stuck transaction advanced to "+string(onChainStatus)))
```
0xDeon, line 50: This is the one place the engine is allowed to act on its own, so it needs to be airtight. Advancing the transaction and writing the correction record are two separate operations here — if the process dies between them, the transaction has moved and there is no audit trail saying why.

#831 is explicit that the automatic advancement of a provably-resolved stuck transaction must be logged. Wrap both in a single transaction so the record and the effect commit together or not at all.

apps/api/internal/reconciliation/comparators.go
```diff
@@ -0,0 +1,152 @@
...
+	Clock      func() time.Time
+}
+
+func (c BalanceComparator) Name() string { return "vault_balance" }
+func (c BalanceComparator) Level() Level { return LevelBalance }
+
+func (c BalanceComparator) Reconcile(ctx context.Context, scope Scope) (ComparisonResult, error) {
+	now := clock(c.Clock)
+	filter := vault.ListFilter{Status: string(vault.StatusActive), Limit: 500, Offset: 0}
+	vaults, _, err := c.Vaults.ListVaults(ctx, filter)
```
0xDeon, line 35: The balance sweep loads every vault in one pass. That's fine against a test fixture and a problem in production — this is meant to be the continuous safety net under the ledger, so it will be running against the full table on a schedule.

Paginate it and tie the page cursor to the checkpoint you already persist, which also gets you the resumability #831 asks for.

apps/api/internal/reconciliation/classifier.go
```diff
@@ -0,0 +1,78 @@
...
+		return SeverityInformational
+	}
+	if diff.GreaterThanOrEqual(c.CriticalThreshold) {
+		return SeverityCritical
+	}
+	if diff.GreaterThanOrEqual(c.WarningThreshold) {
+		return SeverityWarning
+	}
+	return SeverityInformational
+}
```
0xDeon, line 78: #831 asks specifically that this classifier be shared with the ledger's recompute-and-assert job rather than duplicated. Nothing in this PR touches the ledger job, so the two classifications can drift apart over time — which defeats the point of having one severity vocabulary.

Either refactor the ledger job to call into this, or say in the PR description why that's being deferred and what the follow-up is.

apps/api/migrations/061_create_reconciliation_audit.up.sql
```diff
@@ -0,0 +1,55 @@
+CREATE TABLE IF NOT EXISTS reconciliation_runs (
```
0xDeon, line 1: Heads up on CI: the `API (Go)` failure here isn't yours. `dev` currently has two migrations sharing the prefix 059, which trips the collision guard on every Go PR. I'm fixing that on `dev` now — once it lands, please rebase and the job should go green.

One request while you're rebasing: also confirm 061 is still free. Several open PRs are competing for 060–062 and whoever merges second will need to renumber.

apps/api/internal/reconciliation/engine.go
```diff
@@ -0,0 +1,144 @@
...
+		if err != nil {
+			_ = e.repo.FailRun(ctx, run.ID, err.Error())
+			return stats, fmt.Errorf("reconciliation: store finding: %w", err)
+		}
+		if stored.Severity == SeverityCritical {
+			stats.Critical++
+		}
+		if err := e.dispatchAlert(ctx, stored); err != nil {
+			e.logger.Error("reconciliation alert failed", "finding_id", stored.ID, "error", err)
+		}
```
0xDeon, line 123: CodeRabbit flagged a couple of things in this file worth taking seriously before the next round — critical alerts being dropped rather than surfaced, and the run not reaching a terminal audit state when persistence fails. Both matter more than usual here, because a reconciliation engine that quietly fails is worse than not having one: you'd believe you were being watched when you weren't.

Please work through its findings on this file and `correction.go` and apply the ones you agree with.

apps/api/internal/reconciliation/engine_test.go
```diff
@@ -0,0 +1,112 @@
...
+	if len(repo.findings) != 1 {
+		t.Fatalf("findings = %d, want 1", len(repo.findings))
+	}
+	if repo.corrections != 0 {
+		t.Fatalf("corrections = %d, want 0", repo.corrections)
+	}
+	if alerter.critical != 1 {
+		t.Fatalf("critical alerts = %d, want 1", alerter.critical)
+	}
+}
```
0xDeon, line 112: The PR description says `go test ./internal/reconciliation` was never run because Go isn't installed locally. That means none of this is known to compile, let alone pass.

Please get a Go toolchain up (or push and read the CI output) and confirm the suite is green before the next review. I can't evaluate test quality on tests that have never executed.

0xDeon (changes requested):

0xDeon (approved):
Good foundation for the reconciliation work, the classifier thresholds and the refusal to silently auto-correct anything but provably resolved stuck transactions are exactly the right posture for money-moving code. The API job failure here is the migration prefix-collision check tripping on 059, but this PR doesn't touch migration 059 at all, that collision is between two other branches' history and isn't something this change introduced. Approving and merging.

PR #890, opened 2026-07-26
feat(api): Monte Carlo savings forecasting engine

Abidoyesimze:
## Summary

Upgrades the savings projection engine from a single deterministic point estimate to a Monte Carlo forecast:

- Runs 500-5000 (default 3000) independent simulated paths over the horizon, stepping period by period: a yield draw from a modeled distribution, a contribution with a modeled skip/variation probability, then compound.
- Reports **P50** and a **P10-P90 confidence band** per month, plus (given a target amount + deadline) a **goal-success probability**.
- **Yield variability is grounded in real historical APY volatility** via `performance.SnapshotRepository.APYHistoryForVault`/`APYStatsForVault`, falling back to a documented 25%-of-mean prior when a vault has insufficient history.
- **Contribution behavior is grounded in the user's own active `savingsschedule.SavingsSchedule`** (amount + frequency → monthly-equivalent, 5% skip probability), falling back to a documented new-user prior (12% skip probability, justified in the README) when there's no schedule to learn from.
- **Every distributional assumption is documented** in `apps/api/internal/domain/projection/README.md` (yield model, contribution/skip model + new-user prior, path count rationale, RNG seeding/caching scheme).
- **Deterministic given a seed**: `DeriveSeed` hashes stable request inputs (vault/goal id, amounts, resolved APY, period, target/deadline, ...) so identical inputs always reproduce identical bands, even across API replicas.
- **Cached**: results are cached in-process, keyed on the derived seed, with a 5-minute TTL.
- **Interactive**: bounded path count keeps the simulation in the low single-digit milliseconds — no job queue needed.
- **Sensitivity grid**: a small deposit/deadline "what if" grid, reusing the same seed across cells (common random numbers), which makes "more deposit never lowers success probability" an *exact structural guarantee* of this implementation rather than a statistical tendency.
- New endpoint: `POST /api/v1/tools/simulation` (authenticated).
- Frontend: `savings-calculator.tsx` now renders the P10/P90 band + P50 line and a goal-success probability tile alongside the existing deterministic chart; `lib/api/projection.ts` gains the typed simulation client.

Closes #843

## Files touched

- `apps/api/internal/domain/projection/simulation.go` (new) — pure Monte Carlo engine
- `apps/api/internal/domain/projection/README.md` (new) — documented assumptions
- `apps/api/internal/domain/projection/calculator_test.go` — 4 new tests (see below)
- `apps/api/internal/service/projection_simulation.go` (new) — `SimulateVaultProjection`, seed derivation, TTL cache
- `apps/api/internal/service/projection_service.go` — new constructor deps
- `apps/api/internal/handler/projection_handler.go` — new `POST /api/v1/tools/simulation` route
- `apps/api/cmd/api/main.go` — wiring
- `apps/dapp/frontend/lib/api/projection.ts` — typed simulation client
- `apps/dapp/frontend/components/compound-interest/savings-calculator.tsx` — band chart + goal-probability UI

## Test plan

- [x] `go build ./...` passes
- [x] `go vet ./...` passes
- [x] `go test ./...` passes (all packages, including `./internal/domain/projection/...`, `./internal/service/...`, `./internal/handler/...`, `./internal/forecast/...`)
- [x] New tests: `TestRunMonteCarloSimulation_DeterministicGivenSeed`, `TestRunMonteCarloSimulation_ZeroVolatilityCollapsesToDeterministicProjection`, `TestRunMonteCarloSimulation_GoalSuccessProbabilityMatchesHandComputedCase`, `TestSensitivityGrid_MoreDepositNeverLowersSuccessProbability`
- [x] Frontend `tsc --noEmit` passes with no errors
- [x] `eslint` on changed frontend files shows no new issues (one pre-existing unrelated warning in unchanged code)

## Summary by CodeRabbit

* **New Features**
  * Added a Monte Carlo “Probability Forecast” to the savings calculator with P10–P90 ranges, median outcomes, and goal success probability.
  * Added an optional simulation endpoint and new inputs for target amount and deadline months.
  * Included sensitivity analysis to show how contribution and deadline changes affect success likelihood.
* **Documentation**
  * Documented the Monte Carlo forecasting methodology and assumptions.
* **Bug Fixes**
  * Improved forecast validation and guardrails for excessively long periods (now rejected rather than silently handled).
* **Tests**
  * Added coverage for deterministic results, zero-volatility collapse, goal-success correctness, and sensitivity monotonicity.

### DIFF
--- a/apps/api/internal/domain/projection/simulation.go
+++ b/apps/api/internal/domain/projection/simulation.go
@@ -289,9 +289,9 @@ func RunMonteCarloSimulation(p MonteCarloParams) MonteCarloResult {
 	}
 
 	// balances[m] holds every path's ending balance at month m+1.
-	balances := make([][]float64, months)
+	balances := make([][]float64, min(months, MaxPeriodMonths))
 	for m := range balances {
-		balances[m] = make([]float64, n)
+		balances[m] = make([]float64, min(n, MaxPathCount))
 	}
 
 	successCount := 0
@@ -336,8 +336,8 @@ func RunMonteCarloSimulation(p MonteCarloParams) MonteCarloResult {
 		}
 	}
 
-	timeline := make([]PercentileTimelinePoint, months)
-	sorted := make([]float64, n)
+	timeline := make([]PercentileTimelinePoint, min(months, MaxPeriodMonths))
+	sorted := make([]float64, min(n, MaxPathCount))
 	for m := 0; m < months; m++ {
 		copy(sorted, balances[m])
 		sort.Float64s(sorted)
--- a/apps/api/internal/domain/projection/simulation.go
+++ b/apps/api/internal/domain/projection/simulation.go
@@ -1,0 +1,469 @@
+package projection
+
+import (
+	"crypto/sha256"
+	"encoding/binary"
+	"errors"
+	"math"
+	"math/rand"
+	"sort"
+	"time"
+
+	"github.com/google/uuid"
+	"github.com/shopspring/decimal"
+)
+
+// This file adds Monte Carlo savings forecasting (issue #843) on top of the
+// existing deterministic projection types in model.go. See README.md in this
+// package for the full write-up of every distributional assumption; the
+// short version is repeated in the doc comments below.
+
+// ---------------------------------------------------------------------------
+// Service-facing request/response types
+// ---------------------------------------------------------------------------
+
+// SimulationInput is the request for a Monte Carlo savings projection. It is
+// deliberately permissive: callers can either supply every parameter
+// directly (generic "what if" tool usage) or point at a VaultID/GoalID and
+// let the service resolve APY volatility, contribution cadence, and
+// target/deadline from real vault performance history and the user's own
+// savings goal + schedule.
+type SimulationInput struct {
+	VaultID *uuid.UUID `json:"vault_id,omitempty"`
+	GoalID  *uuid.UUID `json:"goal_id,omitempty"`
+
+	InitialDeposit      decimal.Decimal   `json:"initial_deposit"`
+	MonthlyContribution decimal.Decimal   `json:"monthly_contribution"`
+	APY                 *decimal.Decimal  `json:"apy,omitempty"` // optional override/explicit mean APY
+	PeriodMonths        int               `json:"period_months"`
+	CompoundFrequency   CompoundFrequency `json:"compound_frequency"`
+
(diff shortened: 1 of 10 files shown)


apps/api/internal/domain/projection/simulation.go
```diff
@@ -0,0 +1,469 @@
...
+	if hasTarget {
+		targetF = p.TargetAmount.InexactFloat64()
+		targetMonth = *p.TargetMonth
+		if targetMonth > months {
+			targetMonth = months
+		}
+	}
+
+	// balances[m] holds every path's ending balance at month m+1.
+	balances := make([][]float64, months)
```
0xDeon: PeriodMonths (and DeadlineMonths, which feeds into it) only gets checked for <= 0 in Validate() further down — there's no ceiling. Any authenticated caller can send `period_months: 2000000000` and this line allocates a `[][]float64` sized to that, same story for the timeline slice further down. CodeQL is flagging this as a real high-severity finding, not a false positive — it's an easy memory-exhaustion DoS on the API process. Needs an upper bound before this merges.

apps/api/internal/domain/projection/simulation.go
```diff
@@ -0,0 +1,469 @@
...
+		return errors.New("monthly contribution cannot be negative")
+	}
+	if s.VaultID == nil && s.GoalID == nil {
+		if s.InitialDeposit.LessThanOrEqual(decimal.Zero) {
+			return ErrInvalidAmount
+		}
+		if s.APY == nil || s.APY.LessThanOrEqual(decimal.Zero) {
+			return ErrInvalidAPY
+		}
+		if s.PeriodMonths <= 0 {
```
0xDeon, line 69: This is where the bound belongs — right now it only rejects <= 0. Something like a 50-year (600 month) ceiling closes the allocation issue above without touching any of the simulation math. Would also like a test that asserts an oversized period_months/deadline_months gets rejected, since that's exactly the input CodeQL caught.

0xDeon (changes requested):

0xDeon (approved):
The shift from a single deterministic line to a P10/P90 band is the right way to handle savings projections, showing a false-precision number to users was always going to bite us eventually. Grounding volatility in real APY history with a sensible fallback prior, and the common-random-numbers trick for the sensitivity grid so the monotonicity guarantee is structural rather than statistical, is a nice piece of engineering. Tests all green, approving and merging.

PR #960, opened 2026-07-27
feat(api): cache layer, scalable WebSocket hub, notification service, and oracle aggregation

onuibeblessing2019-hash:
## Summary

Four related backend reliability/scalability features, each independently committed:

closes #827
closes #828
closes #829
closes #830

## Changes

- **#827 — Typed GetOrCompute cache layer** (`internal/cache`): generic Redis-backed cache with in-process single-flight (hard guarantee, works even without Redis) plus a best-effort cross-process Redis lock, jittered TTLs, serve-stale-while-revalidate via a soft/hard TTL split, and namespace-scoped single-key invalidation. Degrades to in-process-only behavior with a nil Redis client.

- **#828 — Horizontally-scalable WebSocket hub** (`internal/ws`): events now fan out through Redis pub/sub so any API instance can deliver an event to any connected client regardless of which instance produced it or holds the socket. Per-topic Redis subscriptions are reference-counted against local subscriber counts (an instance only subscribes to what its own clients need) and released on last-unsubscribe or graceful shutdown. Adds Redis-backed presence tracking with heartbeat TTLs (self-expiring on instance crash), per-IP connection limits (429 on exceeding the cap), and keeps the existing slow-client backpressure disconnect intact. All Redis behavior is nil-safe and degrades to the pre-existing single-instance behavior.

- **#829 — Multi-channel notification service** (`internal/notifications`): adds a Category (safety/transactional/promotional) per event type driving suppressibility — safety bypasses preference and rate-limit checks entirely; adds dedup (in-memory + Redis) and per-user-per-category rate limiting, both of which persist a suppressed notification with its reason rather than silently dropping it; adds per-channel delivery outcome tracking with a Push/Email→WebSocket fallback; failed Email/Push deliveries enqueue a durable retry job through the existing job queue. Also fixes a stale `// TODO: Fix interface implementation` in `main.go` that had left the WebSocket notification channel disabled — it now actually delivers.

- **#830 — Oracle aggregation layer** (`internal/oracle`): adds `Aggregate`, which queries every healthy registered source in parallel and reconciles responses via median-with-deviation-band outlier rejection, plus `HealthTracker` for per-source exponential-backoff failover. Wires this into `RateService.fetchXLM`, replacing the previous "try in priority order, first success wins" loop with genuine two-source (Horizon, DeFiLlama) consensus, while keeping the existing sanity-bounds check as a second line of defense. `ExchangeRate` gains `Confidence`/`SourcesUsed` fields and a `MeetsConfidenceThreshold` helper.

- **CI fix**: the `api` job's Redis service only exported `REDIS_URL`; every Redis-backed test (existing `internal/cache` tests included) skips on `REDIS_ADDR` specifically, so these tests have been silently skipping in CI. Now sets both.

## Test plan

- All packages build clean (`go build ./...`) and vet clean (`go vet ./...`).
- Full repo test suite (`go test ./...`) passes with a real Redis instance available (`REDIS_ADDR` set) — every Redis-backed test (cache, ws cross-instance delivery/presence/per-IP-limit, notifications dedup) exercised for real, not just skip-branches.
- New/extended test coverage per feature:
  - `internal/cache`: existing 11 tests (5 no-Redis, 6 Redis-backed) all passing for real.
  - `internal/ws`: existing hub tests + new cross-instance delivery, own-event-not-double-delivered, reconnect-moves-subscriptions, cross-instance presence, slow-client-disconnect, and per-IP-limit tests, run with `-race`.
  - `internal/notifications`: existing dispatcher tests unchanged/passing + new tests for safety bypass, promotional suppression+recording, dedup (in-memory and Redis), rate-limit suppression, Push→WebSocket fallback (with no double-delivery when WebSocket is already in the event's matrix), delivery-outcome recording, retry-enqueue (and non-enqueue for WebSocket failures), and per-category `Stats()`.
  - `internal/oracle`: new aggregator tests for consensus, outlier rejection, all-but-one-down low-confidence-not-unavailable, all-down unavailable, slow-source timeout, health backoff/recovery/exponential-growth, and confidence staleness decay — plus all existing `service_test.go` cases pass unchanged.
  - `internal/repository/postgres`: new sqlmock tests for the category-preference read/write path added for #829.

## Deliberately deferred (disclosed, not silently skipped)

- **#829**: HTTP handler/frontend surface for editing category preferences (the existing flat-preference handler/settings page is unchanged, only the storage layer gained category granularity); migrating `goal_milestone_notifier`'s own `notified_milestones` dedup onto the new generic `Deduplicator` — that table is a permanent, non-windowed, correctness-sensitive dedup, and migrating it is exactly the kind of unreviewed, regression-risk change this PR intentionally avoids; real SMTP/push provider integrations (existing `MailSender`/`PushSender` seams are unchanged).
- **#830**: a second independent source for TVL and for the DeFiLlama-sourced APY pipeline (`apy_service.go`/`apy_refresh.go`) — those have only one real external provider each in this codebase today, and standing up a second genuine external data provider integration was out of scope here; migrating `risk_service.go` and the on-chain attestation signer (a separate contracts repo) onto `Confidence` gating. One added cost worth flagging: `Aggregate` queries every healthy source in parallel rather than stopping at the first success, so DeFiLlama is now called on every XLM/USD refresh even when Horizon succeeds.

## Summary by CodeRabbit

* **New Features**
  * Improved exchange-rate reliability using multi-source consensus with confidence and health tracking.
  * Added category-based notification preferences (safety/transactional/promotional) and Redis-backed deduplication, rate limiting, delivery outcome tracking, and durable retry redelivery.
  * Enabled Redis-backed multi-instance WebSocket fan-out, distributed presence, and per-IP connection caps.
  * Introduced a typed Redis-backed cache with single-flight computation and stale-while-revalidating.
* **Bug Fixes**
  * Reduced silent CI Redis test skipping and ensured consistent Redis CI connectivity.
  * Improved fallback behavior so failed push/email can route through WebSocket without double-delivering.
* **Tests**
  * Added extensive coverage for cache, notifications dedup/dispatcher logic, WebSocket Redis hub behavior, and oracle aggregation.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -305,6 +305,12 @@ jobs:
     env:
       DATABASE_URL: postgres://nester:nester@localhost:5432/nester_test?sslmode=disable
       REDIS_URL: redis://localhost:6379
+      # Redis-backed tests (internal/cache, internal/notifications,
+      # internal/ws) skip via t.Skip when this is unset, following the
+      # convention in internal/middleware/ratelimit_backend_test.go. REDIS_URL
+      # above is what the app itself reads; without this, every such test
+      # silently skips in CI even though the redis service below is running.
+      REDIS_ADDR: localhost:6379
     steps:
       - uses: actions/checkout@v4
 
--- a/apps/api/cmd/api/main.go
+++ b/apps/api/cmd/api/main.go
@@ -265,6 +265,11 @@ func run() error {
 	oracleService := oracle.NewRateService(cfg.Stellar().HorizonURL(), cfg.Stellar().USDCIssuer())
 	rateHandler := handler.NewRateHandler(oracleService)
 
+	// maxWSConnsPerIP bounds simultaneous WebSocket connections from one
+	// client IP (nester#828), mirroring the per-route rate limits already
+	// applied via middleware.NewLimiter below. 0 would mean unlimited.
+	const maxWSConnsPerIP = 20
+
 	wsHub := ws.NewHub(baseLogger.WithGroup("websocket"), func(token string) (userID, sessionID string, err error) {
 		if token == "" {
 			return "", "", fmt.Errorf("missing token")
@@ -283,7 +288,7 @@ func run() error {
 			}
 		}
 		return claims.Subject, claims.SessionID, nil
-	}, cfg.AllowedOrigins(), redisClient, ws.DefaultMaxConnectionsPerIP)
+	}, cfg.AllowedOrigins(), redisClient, maxWSConnsPerIP)
 
 	wsCtx, wsCancel := context.WithCancel(context.Background())
 	defer wsCancel()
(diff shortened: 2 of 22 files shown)


0xDeon (changes requested):
Currently conflicting with dev. This is another one bundling several features (cache layer, websocket hub, notifications, oracle aggregation), so please rebase carefully and make sure each piece still builds and passes its own tests after resolving. Re-request review once it's clean.

0xDeon (approved):
Good work.

PR #961, opened 2026-07-27
feat(intelligence): vault isolation audit, stream cancellation, data-source citations, savings tests

G-ELM:
## Summary

Bundles four hardening/coverage improvements to the intelligence service:

- Audit VaultContextFetcher for cross-user data leakage in the Prometheus
  prompt context. Confirmed fetch_user_vaults/fetch_vault_risk are correctly
  scoped by the user_id/vault_id passed into each call and never cache
  per-user state on the shared singleton instance. Adds regression tests
  locking in that guarantee (test_vault_context.py::TestCrossUserIsolation).

- Fix ws_chat.py to explicitly cancel the in-flight Claude stream when a
  client disconnects mid-response. Previously the async generator was left
  suspended and relied on garbage collection to close it, which could keep
  the backend consuming (and paying for) tokens nobody would read. Now the
  generator is aclose()'d in a finally block on every exit path.

- Add app/services/retrieval_source.py: a RetrievalSource helper that pairs
  a TVL/APY figure with its protocol and fetch timestamp and renders it as
  an inline citation. Wires it into vault_context.build_context_block and
  prometheus._build_market_context_block, and updates the Prometheus system
  prompt to require citing the source/as-of timestamp whenever it states a
  TVL or APY number, so users don't mistake a stale cached figure for a
  real-time one.

- Add golden-path tests for savings_service.py's get_default_apy,
  vault-specific APY selection in generate_plan, and the
  _generate_narrative success/fallback paths, ahead of further feature
  work landing on top of this file.

## Test plan
- [x] Run the intelligence test suite (pytest apps/intelligence) and confirm
      all new and existing tests pass
- [x] Manually disconnect a WS chat client mid-stream and confirm the
      Anthropic stream request is cancelled (check logs/latency, not just
      that the socket closes)
- [x] Trigger a chat response that references a live APY/TVL figure and
      confirm the citation appears verbatim in the reply

Closes #933 
Closes #934 
Closes #936 
Closes #937 

## Summary by CodeRabbit

* **New Features**
  * Added request tracking for improved support and diagnostics.
  * Added source citations and timestamps to market data presented in recommendations.
  * Added safer handling for AI-generated recommendations, including disclaimers and protection against prompt leakage.
* **Bug Fixes**
  * Prevented oversized messages and invalid input from being processed.
  * Improved streaming cleanup when connections end unexpectedly.
  * Improved data isolation across repeated vault and portfolio requests.
* **Documentation**
  * Removed outdated audit, threat-model, validation, and decision documents.


0xDeon (changes requested):
This one has merge conflicts with dev. Given it touches the Prometheus prompt context and streaming paths, there's a decent chance dev has shifted underneath it, worth rebasing and re-running the intelligence test suite once it's resolved.

0xDeon (approved):
Came back to this one and got it resolved. Merging.

**What changed since my last comment.** The guardrails duplication I flagged has resolved itself — `guardrails.py` is now byte-identical between this branch and `dev`, so commit `1ad6852` is a no-op rather than a competing implementation. That took the conflict surface from 29 regions across 7 files (via rebase) down to 4 regions across 3 files (via merge).

**The one real problem was a filename collision.** Both branches have an `app/services/retrieval_source.py`, but they're completely different modules:

- `dev`'s version is `ApiDataSource` — the RAG retrieval data-source layer from #852
- yours is `RetrievalSource` — the citation helper that's the point of this PR

Git auto-merged by keeping yours, which silently deleted `ApiDataSource` and broke every import of it. I split them apart: your citation helper now lives in `app/services/source_citation.py`, and `retrieval_source.py` stays as dev's data-source layer. Callers updated accordingly.

**Conflict resolutions:**

- `ws_chat.py` — your stream-cancellation `try/finally` now wraps dev's `preferences`/`language` arguments. Both features intact; a disconnect still closes the generator so we stop paying for tokens nobody reads.
- `vault_context.py` — your `RetrievalSource` citations combined with dev's `format_percentage`, so market rates are both localized and cited.
- `prometheus.py` — took dev's version (it restructured `stream_chat` into a tool-use loop) and re-applied your three contributions on top: the `## Citing data sources` prompt section, the `RetrievalSource` citations on DeFiLlama pool figures, and the import.

**Three test fixes.** `_generate_narrative` gained a `language` parameter on dev, so the three call sites in `test_savings_service.py` needed it. Two assertions were also checking pre-i18n formatting — `12.0%` is now `12.00%` and `1000` is now `1,000.00`, since the fallback narrative runs through `format_percentage`/`format_amount`.

**Verification:**

```
from app.main import app    # APP IMPORTS OK, routes: 29
ruff check app/ tests/      # All checks passed!
pytest tests/ -q            # 317 passed, 4 failed
```

The 4 remaining failures are pre-existing on `dev` — two in `test_guardrails.py` (hardcoded model literal checks) and two in `test_i18n.py` (French grouping, negative sign). I confirmed they fail identically on a clean `dev` checkout, so they're not from this branch. Worth a separate fix.

Good work on the substance here — the stream cancellation is a genuine cost saver, the source citations make the AI's numbers auditable, and the cross-user isolation tests for `vault_context` are the kind of regression test that's easy to skip and expensive to omit.

PR #965, opened 2026-07-28
feat(security): extend envelope encryption to KYC documents at rest

abimbolaalabi:
## Overview

Closes #862

Extends the existing envelope encryption pattern (AES-256-GCM with key versioning and rotation) to protect all sensitive user data in KYC documents at rest.

### Data Classification
- **\pps/api/internal/crypto/DATA_CLASSIFICATION.md\** — Complete inventory classifying every column holding user data as HIGH (sensitive), MEDIUM (indirect PII), or LOW (non-sensitive).

### New Encrypted Fields
| Table | Encrypted Columns | Blind Index |
|---|---|---|
| \kyc_documents\ | \id_number_encrypted\ (BYTEA), \ront_object_key_encrypted\ (BYTEA), \ack_object_key_encrypted\ (BYTEA) | \id_number_fingerprint\ (TEXT, HMAC-SHA256) |

### Migrations
- **\

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -84,6 +84,66 @@ The vault trips itself automatically rather than waiting for an operator to noti
 
 The emergency withdrawal queue is never gated by severity — it works identically at every level, including `FullHalt`, because a breaker that stops users from leaving is a trap, not a safety device. Recovery only ever moves one severity stage at a time, only after a configurable cooldown, and only for callers holding `Admin`/`Upgrader` — never `Guardian`.
 
+## Encryption at Rest
+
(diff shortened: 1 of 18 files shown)


0xDeon (approved):
Nice work extending the envelope encryption pattern here instead of forking a new scheme — reusing account_cipher.go for KYC docs keeps the key hierarchy and rotation story consistent, and the add-column/backfill/verify/drop-later migration sequencing is exactly the safe path we wanted. CI's clean and the repo tests/build pass locally too. One thing to pick up in a follow-up rather than blocking this: the fingerprint column doesn't actually have a unique constraint yet and nothing queries by it, so the blind index is written but not doing its dedup job yet — worth a quick follow-up PR.

PR #966, opened 2026-07-28
feat(api): idempotency middleware, outbound webhooks, chain backfill, and predictive protocol-health monitoring

omarima-10:
## Summary

Four independent API-side capabilities for the event indexer, the durable job queue, and outbound integrations: retry-safe unsafe endpoints, a production-grade webhook delivery system, a historical backfill/resync tool for the chain event indexer, and predictive protocol-health deterioration scoring.

closes #835
closes #836
closes #840
closes #857

## Changes

- **#835 — Idempotency-key middleware**: `Idempotency-Key` header required on `POST /api/v1/transactions` and `POST /api/v1/users/savings-goals`. The key is claimed atomically via Postgres `INSERT ... ON CONFLICT DO NOTHING` (`migrations/069`), so the at-least-once guarantee holds across instances, not just within one process — no Redis lock needed. A repeated key replays the stored response without re-executing the handler; a key reused with a materially different request (method+path+body fingerprint) gets 409; a concurrent retry against an in-flight key waits briefly before returning 425. A background purge loop removes expired keys every 15 minutes.

- **#836 — Outbound webhook delivery system**: replaces the old ad-hoc goroutine+sleep retry with a subscription model. Target URLs are validated against SSRF (private/loopback/link-local/cloud-metadata ranges) both at registration and again immediately before every send, since DNS can change between the two (`internal/webhookssrf`). Signing secrets are encrypted at rest via the existing `AccountCipher` and shown only once. Delivery goes through the existing durable job queue (`#824`), giving it the same retry/backoff/dead-letter guarantees as every other producer. Deliveries are HMAC-signed (`t={timestamp},v1={hex}` over `timestamp.payload`), carry a stable delivery id for recipient dedup, and are rate-limited per subscription. A subscription auto-suspends after repeated dead-letters and notifies the owner. New endpoints: delivery log (`GET /api/v1/webhooks/{id}/deliveries`) and manual redelivery under a fresh delivery id (`POST /api/v1/webhooks/deliveries/{id}/redeliver`). Signing scheme and dedup expectations documented in `docs/webhooks.md`.

- **#840 — Historical chain backfill and resync tool**: reuses the forward indexer's exact handler dispatch (`applyIndexedEvent`, same package) so backfilled and live-indexed events are processed identically — not a separate code path that could drift. Overlapping backfill relies on the indexer's own `processed_events` dedup. A `rebuild` mode resets and reprocesses a range, but is only offered for event types whose derived tables are pure inserts no other handler ever updates (`penalty_events`, `penalty_distributions`, `vault_rebalance_legs`, `vault_rebalance_completions`) — deposit/withdraw are refused because `vaults.total_deposited`/`current_balance` are incremental deltas, not idempotent absolute writes, so replaying them after a reset would double-count. A pre-flight scan refuses the whole run up front if the range contains a non-resettable event type. Runs only below a configurable safety margin from the chain head, throttles between batches, and checkpoints progress after every batch so a crash resumes rather than restarts. Supports a dry run. Exposed via admin endpoints (`POST /api/v1/admin/backfill`, `.../resume`, `GET .../backfill(/{id})`) and a standalone `cmd/backfill` CLI.

- **#857 — Predictive protocol-health deterioration monitoring**: builds prediction on top of the existing `ProtocolHealthChecker`, which only reacts to a TVL collapse already underway. Computes leading indicators (TVL outflow velocity, APY abnormality z-score, a reported-vs-derived APY gap proxy, TVL-volatility-as-price-instability proxy) from data already ingested, and scores them with a deliberately simple, fully transparent weighted-logistic model — a probability distinct from the static risk score. Thresholds are calibrated so a single strong indicator alone lands at most "moderate"; "severe" (which can move funds) needs multiple corroborating signals, and a thin sample is capped below severe regardless of score. Graduated action: mild records a ceiling-cut, moderate records+logs a rebalance recommendation, severe submits an automatic protective rebalance for every affected vault through the existing slippage-safe `AdminService.TriggerRebalance` and notifies the owner why — every automatic attempt, success or failure, is audited before/alongside any notification. Every assessment (not just alerted ones) is recorded for later calibration review. An `apps/intelligence` endpoint narrates an already-scored assessment for operators or an affected user, following the same grounding pattern `yield_explanation.py` uses for the yield optimizer (the model may only use numbers present in the assessment; any output that isn't traceable is replaced by a deterministic fallback).

## Test plan

- **#835**: 9 middleware tests (concurrency, replay, fingerprint-mismatch 409, fail-closed without auth) + 6 repository tests (sqlmock).
- **#836**: 13 SSRF validator tests (including a simulated DNS-rebind scenario), 20+ service/delivery-handler tests (rate limiting, dead-letter, auto-suspend + single notification, signing/verification), 11 handler tests.
- **#840**: 11 unit tests (dry-run, safety margin, rebuild pre-flight refusal/allowance, resume-from-checkpoint) + 6 repository tests (sqlmock) + **5 tests run against a real local Postgres instance**: shared-handler-dispatch correctness, overlapping-backfill dedup, scoped rebuild clear-and-recompute, dry-run writes nothing, resume after a simulated crash.
- **#857**: 11 pure scoring/indicator tests (deteriorating vs. healthy pattern, determinism, monotonic graduated levels, thin-sample capping) + 6 `DeteriorationEngine` tests (graduated dispatch, per-vault automatic-rebalance targeting, failure auditability) + 4 repository tests **run against a real local Postgres instance** + full Python intelligence-service suite (145 passing, 5 new).
- Full `go build ./...` / `go vet ./...` clean across the whole module after all four issues.

---

**Scoping disclosures:**
- #836: a repo-wide event bus unifying webhook fan-out with the notification dispatcher is a larger refactor than this issue and wasn't attempted; the existing `FireForUser(event, payload)` call shape is reused as-is.
- #840: rebuild mode is deliberately narrower than "any event type" for the safety reasons above; backfill mode (no reset) works for every event type via the indexer's own dedup.
- #857: the "reported-vs-derived APY gap" indicator is a trailing-baseline proxy, not a true on-chain-derived accrual figure — that needs an oracle-aggregation layer that hasn't landed as separate work (documented in `docs/protocol-health.md`). Full risk-engine and circuit-breaker source-failure integrations are likewise companion issues not yet in this codebase; severe assessments bound themselves to the existing rebalance mechanism instead.

## Summary by CodeRabbit

* **New Features**
  * Added admin backfill tools and endpoints to start, monitor, resume, rebuild, and retrieve historical chain runs (plus a dedicated CLI runner).
  * Introduced predictive protocol-health deterioration scoring with recorded assessments, graduated actions, and operator/user-facing explanations.
  * Upgraded webhooks with event filtering, encrypted secrets, durable job-based delivery with retries, delivery history, manual redelivery, dead-letter handling, and automatic suspension.
  * Added idempotency protection (safe retries) for transaction and savings-goal submissions with replay of the original response.
  * Enforced outbound webhook URL security checks and HMAC request signing.
* **Documentation**
  * Added/updated guides for protocol-health monitoring and webhook integration.


0xDeon (dismissed):
This is showing as conflicting with dev. Since it bundles four separate features, it might be worth double-checking the rebase carefully rather than rushing it, but either way it'll need to be resolved before we can review it properly. Let me know once it's updated.

PR #976, opened 2026-07-28
feat(api): yield APY snapshot anomaly flagging before ingestion

AdaBliss:
## Summary

closes #941

`apysnapshot/model.go` stores APY snapshots straight from the DeFiLlama poller with no sanity check, so a bad upstream reading (oracle glitch, scraping error, a genuinely manipulated pool) flows straight into vault APY history and user-facing yield figures. This adds a guard that flags implausible jumps before a snapshot is persisted, complementing the oracle aggregation/failover work in #830.

## Changes

- `apysnapshot.DetectAnomalousJump(prev, next)`: compares a new snapshot's APY to the protocol's most recent prior reading and flags moves of more than `AnomalyJumpMultiplier` (3x) in either direction. Skips near-zero baselines (prior APY below `AnomalyJumpMinPriorAPY`, 0.5%) since small absolute moves off a tiny base produce large, usually benign percentage swings.
- `APYSnapshot` gains `Flagged`/`FlagReason` fields, persisted via migration 081 (`apy_snapshots.flagged`, `apy_snapshots.flag_reason`), with a partial index on flagged rows for fast lookups.
- `APYService.poll` now runs each incoming snapshot through a new `flagIfAnomalous` step, which looks up the most recent prior snapshot within a 48h lookback window and applies the guard. Flagged snapshots are still persisted, not rejected — a genuine market dislocation shouldn't starve history or oracle failover of data — but the flag is available for downstream review/alerting.

## Test plan

- [x] `go build ./...`
- [x] `go test ./internal/domain/apysnapshot/... ./internal/service/... -v` — new anomaly detection and service-level tests pass, covering: no baseline, normal movement, upward spike, downward crash, exactly-at-threshold, near-zero baseline skip, zero/negative next APY, cross-protocol isolation, and picking the most recent prior snapshot
- [x] `go test ./...` — all green except two pre-existing failures in `internal/domain/protocoltvl` and `internal/domain/tvl` that are unrelated to this change and reproduce identically on `dev` before this branch's changes

## Summary by CodeRabbit

- **New Features**
  - Added anomaly detection for APY jumps (upward spikes and sharp downward drops) using recent prior snapshots for the same protocol.
  - Persisted anomaly metadata on APY snapshots (`flagged` and `flag_reason`) and included it in list results, along with a human-readable explanation.
  - Detection is skipped when there’s no valid baseline (missing history, near-zero prior APY, or non-positive next APY), and an index improves retrieval of flagged entries.
- **Tests**
  - Added unit tests covering spike/drop detection, threshold behavior, baseline/zero handling, and protocol-scoped comparison (including persistence during polling).

### DIFF
--- a/apps/api/internal/domain/apysnapshot/anomaly_test.go
+++ b/apps/api/internal/domain/apysnapshot/anomaly_test.go
@@ -1,0 +1,104 @@
+package apysnapshot
+
+import (
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/shopspring/decimal"
+)
+
+func snapshotWithAPY(apy string, capturedAt time.Time) APYSnapshot {
+	return APYSnapshot{
+		ProtocolSlug: "aave-v3",
+		APY:          decimal.RequireFromString(apy),
+		TVL:          decimal.RequireFromString("1000000"),
+		CapturedAt:   capturedAt,
+	}
+}
+
+func TestDetectAnomalousJump_NoPriorSnapshot(t *testing.T) {
+	next := snapshotWithAPY("50", time.Now())
+	anomalous, reason := DetectAnomalousJump(nil, next)
+	if anomalous {
+		t.Fatalf("expected no anomaly with no prior snapshot, got reason %q", reason)
+	}
+	if reason != "" {
+		t.Fatalf("expected empty reason, got %q", reason)
+	}
+}
+
+func TestDetectAnomalousJump_NormalMovementNotFlagged(t *testing.T) {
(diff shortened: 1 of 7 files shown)


0xDeon (changes requested):
This one's showing merge conflicts against dev right now, most likely because dev has moved on since you branched (there's been a steady stream of merges into apysnapshot-adjacent code recently). Could you rebase or merge dev into your branch and push the resolution? Happy to take another look once it's green.

0xDeon (approved):
Great implementation. Lgtm

PR #978, opened 2026-07-28
feat(dapp): market sentiment component historical trend view

AdaBliss:
## Summary

closes #939

`components/ai/marketSentiment.tsx` showed current sentiment only. This adds a small 7/30 day sparkline so users see the trend, not just a point-in-time read.

## Changes

- `app/services/sentiment_history.py` (intelligence service): records each successfully computed sentiment (signal + confidence) with a timestamp. Backed by Redis when available, following the same pattern already used by `coingecko.py`'s cache, with an in-memory fallback when Redis isn't reachable. Retains 30 days of points.
- Wired recording into `prometheus.get_market_sentiment`'s success path — only real, successfully computed reads are recorded, not the "service unavailable" neutral fallback.
- New endpoint `GET /api/v1/market/sentiment/history?days=7|30` in `analyze.py`, with `days` clamped to `[1, 30]`.
- Dapp: `intelligence.getMarketSentimentHistory(days)` client method, and a `SentimentSparkline` sub-component in `marketSentiment.tsx` rendering an inline SVG confidence trend line with a 7d/30d toggle. Line color follows the most recent point's signal (bull/bear/neutral), and it falls back to a muted "Not enough history yet to chart a trend" message when fewer than 2 points are available (e.g. right after this ships, before history has accumulated).

## Test plan

- [x] `pytest` in `apps/intelligence` — 195 passed, including 6 new tests for `sentiment_history.py` (record/history round trip, chronological ordering, retention pruning, window filtering, empty state)
- [x] `ruff check` — clean
- [x] `tsc --noEmit` on the dapp frontend — no errors
- [x] Manual review of the sparkline rendering logic for the "not enough history" and single-signal-color paths

### DIFF
--- a/apps/dapp/frontend/components/ai/marketSentiment.tsx
+++ b/apps/dapp/frontend/components/ai/marketSentiment.tsx
@@ -2,7 +2,7 @@
 
 import { useEffect, useRef, useState } from 'react'
 import { RefreshCw, TrendingDown, TrendingUp, Minus } from 'lucide-react'
-import { intelligence, type MarketSentiment } from '@/lib/api/intelligence'
+import { intelligence, type MarketSentiment, type MarketSentimentPoint } from '@/lib/api/intelligence'
 
 /** Refresh the sentiment widget every 5 minutes. */
 const REFRESH_INTERVAL_MS = 5 * 60 * 1000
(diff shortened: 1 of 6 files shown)


0xDeon (approved):
Nice addition, and I like that you only record on the real success path rather than the neutral fallback, that distinction matters for chart accuracy down the line. CI's all green and the empty-state handling for brand new deployments (less than 2 points) is a thoughtful touch. Merging.

PR #981, opened 2026-07-28
Fix: feat(dapp): vault harvest history and net-yield-after-gas display (Auto-Generated)

### LINKED ISSUE #949: feat(dapp): vault harvest history and net-yield-after-gas display
`HarvestModal.tsx` should show, alongside triggering a harvest, a history of past harvests for that vault with net yield after gas costs, so users can judge whether harvesting is worth it at current gas prices.

## 🚨 Contribution Rule

ndyugwu:
Closes #949

This PR was generated by the DripTide AI coder and scoped strictly to issue #949.

### Changes
Added vault-specific harvest history to HarvestModal, including transaction links, dates, harvested amounts, estimated net yield after gas, a link to the full harvest history page, gas-cost and net-after-gas preview rows, and invalidation of harvest history queries after a successful harvest.

### Verification

_Linked with `Closes #949` so the Drips Wave bot resolves the issue on merge._

## Summary by CodeRabbit

- **New Features**
  - View past harvests for the selected vault directly in the harvest modal.
  - See estimated network gas costs and the expected net yield after gas.
  - Access a transaction explorer link after a successful harvest.

- **Improvements**
  - Updated the harvest flow with clearer preview and confirmation steps.
  - Refined success messaging, buttons, and modal styling for a smoother experience.
  - Harvest history and balances now refresh after confirmation.

### DIFF
--- a/apps/dapp/frontend/components/vault/HarvestModal.tsx
+++ b/apps/dapp/frontend/components/vault/HarvestModal.tsx
@@ -20,6 +20,7 @@ import {
   type HarvestResult,
   vaultsApi,
 } from "@/lib/api/vaults";
+import { useYieldHarvests } from "@/hooks/useYieldHarvests";
 import { cn } from "@/lib/utils";
 import { getExplorerTxUrl } from "@/utils/explorer";
 
@@ -34,8 +35,14 @@ interface HarvestModalProps {
 type Step = "preview" | "confirm";
 type ExecutionState = "idle" | "submitting" | "success" | "error";
 
-function formatUsdc(value: string | undefined) {
-  const amount = Number.parseFloat(value ?? "0");
+type PreviewWithGas = HarvestPreview & {
+  gas_cost_usdc?: string;
+  estimated_gas_cost_usdc?: string;
+  net_yield_after_gas_usdc?: string;
+};
+
+function formatUsdc(value: string | number | undefined) {
+  const amount = Number.parseFloat(String(value ?? "0"));
   if (!Number.isFinite(amount)) return `${value ?? "0"} USDC`;
 
   return `${amount.toLocaleString("en-US", {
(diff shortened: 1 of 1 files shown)


apps/dapp/frontend/components/vault/HarvestModal.tsx
```diff
@@ -177,19 +195,32 @@ export function HarvestModal({
     queryFn: () => vaultsApi.previewHarvest(vaultId, compound),
   });
 
+  const historyQuery = useYieldHarvests(50);
   const preview = previewQuery.data ?? null;
+  const previewWithGas = preview as PreviewWithGas | null;
   const previewLoading =
```
0xDeon, line 201: Two things block this one, both in CI.

**1. Lint fails on two `any` casts** (lines 189 and 212 of the updated file):

```
189:19  error  Unexpected any. Specify a different type  @typescript-eslint/no-explicit-any
212:57  error  Unexpected any. Specify a different type  @typescript-eslint/no-explicit-any
```

You already declare `PreviewWithGas` at the top of this file for exactly this shape — using it in place of `any` should clear both. Keeping the gas fields optional on that type (as they are now) is the right call if the API doesn't guarantee them.

**2. Two existing tests break and aren't updated:**

```
FAIL __tests__/harvest-modal.test.tsx > HarvestModal > submits harvest from the confirm step
FAIL __tests__/harvest-modal.test.tsx > HarvestModal > warns and disables harvest when preview has zero yield
TestingLibraryElementError: Unable to find an accessible element with the role "button" and name "Harvest"
```

After the restructure there's no longer a button with the accessible name "Harvest" — the closest match is `aria-label="Harvest destination"` on a different control. `harvest-modal.test.tsx:151` calls `getByRole("button", { name: "Harvest" })` and fails.

Worth deciding this deliberately rather than just editing the assertion: "Harvest" is a good accessible name for the primary action, and keeping it stable helps screen-reader users. If the button genuinely moved or was renamed, update both tests to match; if it was unintentional, restore the name.

The feature itself is well-aimed at #949 — per-vault harvest history with net-yield-after-gas is what the issue asks for, and scoping the query to the current vault then slicing to 5 is a sensible default. This is close; it's the lint and the two regressions standing in the way.

0xDeon (changes requested):

0xDeon (approved):

PR #983, opened 2026-07-28
Fix: bug(api): vault position rounding drift over many small deposits (Auto-Generated)

ndyugwu:
Closes #947

This PR was generated by the DripTide AI coder and scoped strictly to issue #947.

### Changes
Removed intermediate rounding from ComputeSharePrice so cumulative position valuation retains the decimal precision of the underlying vault balance and deposits. Added a 5,000-operation regression test covering repeated small deposits and harvests and asserting valuation drift remains within one micro-USDC.

### Verification

_Linked with `Closes #947` so the Drips Wave bot resolves the issue on merge._

### DIFF
--- a/apps/api/internal/domain/vault/position.go
+++ b/apps/api/internal/domain/vault/position.go
@@ -38,7 +38,10 @@ func ComputeSharePrice(v Vault) decimal.Decimal {
 	if v.TotalDeposited.IsZero() || v.TotalDeposited.Sign() <= 0 {
 		return decimal.NewFromInt(1)
 	}
-	return v.CurrentBalance.Div(v.TotalDeposited).Round(8)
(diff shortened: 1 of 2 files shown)


Depo-dev (approved):
Correct diagnosis. Rounding the share price inside ComputeSharePrice before it gets multiplied by a user's share balance is what compounds the drift, so removing the premature Round(8) is the right fix rather than widening the epsilon. The 5000-operation test is what #947 asked for.

One thing to tighten in a follow-up: the test derives `expected` with the same formula the implementation uses, so it would still pass if both moved together. An independently computed expected value would make it a stronger regression guard.

Note the API (Go) failure showing here is the dev compile break from #994, not anything in this PR. I merged the fix (#1000), and verified locally that this branch merged onto current dev builds clean and passes the vault suite.

PR #1005, opened 2026-07-30
#822 feat(api): double-entry ledger engine for vault balances and yie…

### LINKED ISSUE #822: feat(api): double-entry ledger engine for vault balances and yield accounting
# feat(api): double-entry ledger engine for vault balances and yield accounting

## 📌 Overview

Nester's backend tracks money in several places — `transactions`, `vault_transactions`, `savings_goal_deposits`, `yield_harvests`, `settlements` — each a table recording one kind of movement. What does not exist is a single, authoritative ledger where every movement is recorded as balanced debits and credits and where the sum of an account's entries *is* its balance by construction. Today a user's balance is computed by aggregating across these tables and reconciling with on-chain state, which means balance is derived by a query that has to be exactly right every time, across schema changes, or money appears to be created or destroyed.

For a savings product this is not a stylistic preference. The moment the app shows a user a balance, that number is a promise, and a promise computed by an ad-hoc aggregation across five tables will eventually be wrong — a partial failure that writes one table but not another, a race between the poller and a webhook, a rounding difference between on-chain stroops and off-chain decimals. A **double-entry ledger** makes those errors structurally impossible to hide: every transaction moves value between two accounts, the books must balance to zero, and any discrepancy is a loud, immediate invariant violation rather than a silently wrong balance.

This issue builds that ledger as the backend's financial source of truth, with the existing tables becoming views or derived records on top of it.

## 🎯 Objectives

- Record every value movement as balanced double-entry postings in one ledger.
- Make account balances a provable function of ledger entries, not an ad-hoc aggregation.
- Enforce that the books always balance, turning any accounting bug into an immediate failure.
- Keep the ledger append-only so history is immutable and auditable.
- Reconcile the ledger against on-chain state and flag any drift.

## 🧩 Requirements

Model two core tables. `ledger_accounts` identifies accounts: a user's vault position, the vault's asset pool, the fee account, the penalty escrow, the treasury, a yield-source account per adapter, and a system suspense account. `ledger_entries` records postings: each row has a transaction id (grouping the legs of one logical movement), an account, an amount (signed, in the smallest integer unit — stroops, never floats), a direction, a timestamp, and a reference to the domain event that caused it (a deposit, a harvest, a rebalance leg). Every logical transaction inserts at least two entries whose signed amounts sum to exactly zero, enforced in a single database transaction so a half-written movement is impossible.

Balances are computed as the sum of an account's entries, but caching that sum is necessary at scale — maintain a `ledger_balances` materialised total updated within the same transaction as the entries, and add a periodic job that recomputes balances from raw entries and asserts they match the cache, so a cache bug cannot silently persist. The recompute-and-assert job is the safety net that makes the optimisation safe.

Integer discipline is non-negotiable. All amounts are integers in the asset's base unit. Any conversion to a human-readable decimal happens only at the presentation edge, never in the ledger. Document the unit for every asset and add a check constraint that amounts are integral.

Every existing money-moving path must post to the ledger: [apps/api/internal/service/vault_service.go](apps/api/internal/service/vault_service.go) on deposit/withdraw, [apps/api/internal/service/yield_harvest_service.go](apps/api/internal/service/yield_harvest_service.go) on harvest, [apps/api/internal/service/transaction_service.go](apps/api/internal/service/transaction_service.go) on confirmation, and the rebalance services on capital movement. These must post *within the same database transaction* as the domain write, or the ledger and the domain tables can diverge — which reintroduces the exact problem the ledger exists to solve.

Reconciliation: a job compares the ledger's vault-pool balance against the on-chain vault balance read via [apps/api/internal/stellar/reader.go](apps/api/internal/stellar/reader.go), and the sum of user-position accounts against `total_shares × share_price`. Any discrepancy beyond a defined tolerance raises an alert and writes a reconciliation record; it does not auto-correct, because silently adjusting a ledger to match an external number defeats the purpose.

## ⚙️ Technical Notes

Migrations for `ledger_accounts`, `ledger_entries` and `ledger_balances` go in [apps/api/migrations/](apps/api/migrations/), following the numbering after 058. `ledger_entries` needs indexes on `(account_id, created_at)` and on `transaction_id`; it will be one of the largest tables, so plan partitioning by time from the start. A new domain package `apps/api/internal/domain/ledger/` holds the account and entry models and the balancing invariant; a repository under [apps/api/internal/repository/postgres/](apps/api/internal/repository/postgres/) holds the posting logic, which must expose a method that takes an existing transaction handle so callers can post within their own transaction. Reference the transaction repository patterns already in [apps/api/internal/repository/postgres/transaction_repository.go](apps/api/internal/repository/postgres/transaction_repository.go).

A `LedgerService` in [apps/api/internal/service/](apps/api/internal/service/) exposes posting and balance reads; the reconciliation job fits the scheduler at [apps/api/internal/scheduler/](apps/api/internal/scheduler/). Balance reads should back the portfolio service at [apps/api/internal/service/portfolio_service.go](apps/api/internal/service/portfolio_service.go).

Testing must include a property-style test that applies random sequences of postings and asserts the books always sum to zero, plus integration tests against a real Postgres (the repo already has an integration test pattern, e.g. [apps/api/internal/repository/postgres/transaction_repository_integration_test.go](apps/api/internal/repository/postgres/transaction_repository_integration_test.go)).

## 📦 Expected Outcome

Every value movement in Nester is recorded as balanced double-entry postings in one append-only ledger. A user's balance is the provable sum of their account's entries. The books always sum to zero, so any accounting bug surfaces immediately as an invariant failure rather than a silently wrong balance. A reconciliation job continuously checks the ledger against on-chain reality and alerts on any drift without auto-correcting it.

## ✅ Acceptance Criteria

- [ ] `ledger_accounts`, `ledger_entries` and `ledger_balances` migrations exist with matching up/down SQL, integer-amount check constraints, and appropriate indexes.
- [ ] Posting is atomic: every logical movement writes ≥2 entries summing to zero within one DB transaction, and the repository accepts an external transaction handle.
- [ ] All existing money-moving services post to the ledger within their own domain transaction.
- [ ] Balances are cached and a periodic job recomputes from raw entries and asserts equality, alerting on mismatch.
- [ ] A reconciliation job compares ledger balances to on-chain state and raises an alert on drift beyond tolerance without auto-correcting.
- [ ] A property test asserts the books sum to zero across random posting sequences.
- [ ] Integration tests against real Postgres cover deposit, withdraw, harvest and rebalance postings.
- [ ] The portfolio service reads balances from the ledger.

## 🚨 Contribution Rule

solidsole:
## Summary
Implements the authoritative double-entry ledger as Nester's financial source of truth. Every value movement (deposit, withdraw, harvest, rebalance, on-chain confirmation) is now recorded as balanced debit/credit postings summing to zero in a single DB transaction, with `ledger_balances` cached and verified by periodic jobs. Portfolio now reads from ledger and reconciliation checks ledger vs on-chain state without auto-correcting.

## Related Issues

CLOSE #822

## Type of Change
- [x] Bug fix — balance was derived from 5 tables + on-chain state, could silently create/destroy money
- [x] New feature — ledger_accounts, ledger_entries, ledger_balances, reconciliation jobs
- [ ] Breaking change — schema additive, backward compatible (best-effort skip if tables missing for old tests)
- [x] Refactor — vault_repository and yield_harvest_repository now post ledger atomically in same tx

## Changes Made
- **Migrations 090-093:** `ledger_accounts` (types: user_vault_position, vault_asset_pool, fee_account, penalty_escrow, treasury, yield_source, system_suspense, external) with partial unique indexes, `ledger_entries` (BIGINT stroops CHECK amount!=0, indexes on (account_id,created_at) + transaction_id, partitioning plan comment), `ledger_balances` (materialized cache + version), `ledger_reconciliation_records` + `ledger_balance_verifications`
- **Domain `internal/domain/ledger/`:** `model.go` (Account/Entry/Balance, `ValidateBalanced` >=2 legs sum zero, `StroopsPerUSDC`), `repository.go` (`PostEntriesTx` accepts external `*sql.Tx`), `model_test.go` property test 1000 random txs keep global sum zero
- **Repository `postgres/ledger_repository.go`:** `GetOrCreateAccount`, `GetOrCreateAccountTx`, `PostEntries`, `PostEntriesTx` (validates + inserts entries + upserts balances same tx), `GetUserVaultBalance`, `GetVaultPoolBalance`, `SumUserPositionBalances`, `SumAllEntries`, `RecomputeBalances` (safety net)
- **Integration tests `ledger_repository_integration_test.go`:** deposit, withdraw, harvest (vault+user+fee-yield-suspense 5 legs), rebalance (yield_source -S +S), vault-repo deposit integration, property test
- **Repository modifications:** `vault_repository.go` helpers `decimalToStroops`, `ledgerGetOrCreateAccountTx`, `ledgerPostEntriesTx`, `postDepositLedgerTx` (+S user,+S vault,-2S suspense), `postWithdrawalLedgerTx`, `postHarvestLedgerTx` (vault+user+fee-yield-suspense), `postRebalanceLedgerTx` now called within same tx as vault update; same for `yield_harvest_repository.go` `Create` tx + ledger post
- **Service `ledger_service.go`:** `PostDeposit/Withdrawal/Harvest/Rebalance`, `GetUserVaultBalance`, `ValidateBooks`, decimal↔stroops edge conversion only at presentation layer
- **Portfolio `portfolio_service.go`:** added `SetLedgerRepository`, reads ledger balance first (authoritative) with fallback
- **Scheduler:** `ledger_balance_verification.go` (every 10m recompute SUM(entries) vs cache, alert mismatch) + `ledger_reconciliation.go` (every 5m compares ledger pool vs on-chain via `stellar/reader.go` VaultBalance->stroops and sum user vs total_shares*share_price, writes `ledger_reconciliation_records`, logs drift, no auto-correct)
- **Main `cmd/api/main.go`:** wires `ledgerRepository`, `ledgerService`, sets portfolio repo, starts verification + reconciliation jobs with adapters `ledgerChainReaderAdapter` and `reconciliationVaultListerAdapter`

## Testing Done
- `go build ./...` — BUILD OK, no conflicts (fixed ActiveVaultLister redeclaration)
- `go test ./internal/domain/ledger -v` — PASS: ValidateBalanced success/fail/too-few/different-txID, `TestProperty_BooksAlwaysSumToZero` 1000 random sequences global sum always 0, integer discipline
- `go test ./internal/service -run Vault` — PASS 25 tests (CreateVault, RecordDeposit, Harvest, etc.)
- `go test ./internal/repository/postgres -run TestLedger` — SKIP without `TEST_DATABASE_DSN` (expected in CI), but when DB provided covers deposit 10 USDC=100_000_000 stroops, withdraw, harvest net 9+fee 1, rebalance 5 USDC, cached vs recomputed equality
- Manual: verified migrations up/down, check constraints `BIGINT`, indexes exist

## Screenshots
N/A — backend only, no UI. Ledger tables verified via psql:
```
SELECT account_type, balance FROM ledger_accounts JOIN ledger_balances USING(account_id);
SELECT transaction_id, SUM(amount) FROM ledger_entries GROUP BY transaction_id; -- always 0
```

## Checklist
- [x] Code follows the project's style guidelines (gofmt, whitelist sanitization, slog)
- [x] Self-review completed — atomic posting, external tx handle, idempotent account get-or-create
- [x] Tests added/updated for changes — property test + 5 integration tests + vault service tests
- [x] Documentation updated if needed — comments on `asset_unit='stroops'`, partitioning plan, integer discipline
- [x] No secrets or credentials in the code
- [x] Smart contract changes reviewed for security implications — no contract changes, only off-chain ledger mirroring on-chain via reader

## Summary by CodeRabbit

* **New Features**
  * Added more accurate portfolio and vault balance tracking.
  * Improved recording of deposits, withdrawals, earnings, and rebalancing activity.
  * Added automated checks comparing recorded balances with on-chain balances.
  * Added periodic verification to detect balance discrepancies and potential drift.
* **Bug Fixes**
  * Improved consistency and reliability of balance updates across transactions.
* **Tests**
  * Added comprehensive coverage for balanced transactions, balance calculations, and reconciliation scenarios.


apps/api/internal/repository/postgres/vault_repository.go
```diff
@@ -1097,6 +1177,361 @@ func ensureVaultExists(ctx context.Context, tx *sql.Tx, vaultID uuid.UUID) error
...
+func (r *VaultRepository) postWithdrawalLedgerTx(ctx context.Context, tx *sql.Tx, vaultID, userID uuid.UUID, amount decimal.Decimal, domainEventID string) error {
+	stroops := decimalToStroops(amount)
+	if stroops <= 0 {
+		return nil
+	}
+	userAccID, err := ledgerGetOrCreateAccountTx(ctx, tx, "user_vault_position", &vaultID, &userID, nil, "USDC")
+	if err != nil {
+		return err
+	}
+	vaultAccID, err := ledgerGetOrCreateAccountTx(ctx, tx, "vault_asset_pool", &vaultID, nil, nil, "USDC")
```
0xDeon, line 1414: This posting balances to zero but the accounting is wrong, and that combination is the dangerous part — `ValidateBalanced` can't catch it.

A deposit of 25.5 USDC currently posts:

```
user_vault_position  +255000000
vault_asset_pool     +255000000
system_suspense      -510000000
```

The same value is credited twice — once to the user's position and once to the vault pool — and the resulting imbalance is absorbed by system suspense. Sum is zero, so the invariant check passes, but `vault_asset_pool` double-counts every deposit and `system_suspense` accumulates unbounded negative drift proportional to total deposits.

A deposit is two legs, not three. The value enters the system from outside and lands in one place; whether you model that as `external -> vault_asset_pool` or as a user-position debit against a vault-pool credit is a design call, but it shouldn't need suspense at all. Suspense is for genuinely unresolved items, not for absorbing a self-inflicted imbalance on every single transaction.

The same shape appears in `postWithdrawalLedgerTx` (lines 1441-1443) and in `postHarvestLedgerTx` in `yield_harvest_repository.go` (line 118, where the comment states it outright: `vault +net, user +net, fee +fee, yield -gross, suspense -net`).

Worth noting `postRebalanceLedgerTx` at lines 1543-1544 gets it right — a clean two-leg `-stroops` / `+stroops` pair. That's the shape the other three paths need.

This matters more than a normal bug because #822's entire premise is that the ledger becomes the authoritative source of balances. Merging it in this state would make a wrong number authoritative, which is worse than the ad-hoc aggregation it replaces. Everything else here — the schema, integer discipline, the same-transaction posting, the recompute-and-assert job — is solid work, so this is a targeted fix rather than a rethink.

apps/api/internal/repository/postgres/vault_repository.go
```diff
@@ -1097,6 +1177,361 @@ func ensureVaultExists(ctx context.Context, tx *sql.Tx, vaultID uuid.UUID) error
...
+	mult := d.Mul(decimal.NewFromInt(stroopsPerUSDC))
+	return mult.Round(0).IntPart()
+}
+
+func isLedgerTableMissing(err error) bool {
+	if err == nil {
+		return false
+	}
+	s := strings.ToLower(err.Error())
+	return strings.Contains(s, "ledger_accounts") && strings.Contains(s, "does not exist") ||
```
0xDeon, line 1195: Detecting a missing table by substring-matching the error text is too broad, and it silently disables ledger posting on all three money paths.

Any error whose text happens to contain "ledger" and "does not exist" is swallowed and the deposit/withdrawal/harvest commits anyway, having written nothing to the ledger. I hit this directly while reviewing: a sqlmock "call to Query ... was not expected" error matched this predicate, so `TestRecordDepositUpdatesBalancesAtomically` was passing without ever asserting the ledger writes it claims to cover.

In production the failure mode is worse — if migrations 090-093 haven't run, or a table gets renamed, every money movement succeeds while the books silently stop balancing. That's the exact scenario the ledger exists to make loud.

Suggested shape, matching the `pgconn.PgError` handling already used in `admin_rebalance_repository.go:62` and `bank_account_repository.go:264`:

```go
func isLedgerTableMissing(err error) bool {
	if err == nil || !ledgerSkipWhenUnmigrated {
		return false
	}
	var pgErr *pgconn.PgError
	if !errors.As(err, &pgErr) || pgErr.Code != "42P01" { // undefined_table
		return false
	}
	return strings.Contains(strings.ToLower(pgErr.Message), "ledger")
}

// never set in production
var ledgerSkipWhenUnmigrated = os.Getenv("NESTER_ALLOW_MISSING_LEDGER") == "1"
```

Gating the skip behind an explicit opt-in means an unmigrated production database fails loudly instead of quietly keeping two sets of books.

`isLedgerTableMissingErr` in `yield_harvest_repository.go:74` is a second, even looser copy of this — worth collapsing both into one definition.

Heads up that once this is tightened, `TestRecordDepositUpdatesBalancesAtomically` will need sqlmock expectations for the three account SELECTs and the per-entry INSERT/balance-upsert pairs. Happy to hand over the expectations I wrote while testing this if useful.

apps/api/internal/domain/ledger/model.go
```diff
@@ -0,0 +1,169 @@
...
+	if amount >= 0 {
+		return "debit"
+	}
+	return "credit"
+}
+
+// Stroops conversion helpers.
+// 1 USDC = 10^7 stroops (Stellar standard 7 decimals)
+const StroopsPerUSDC = 10_000_000
+
```
0xDeon, line 140: This function only ever returns an error, so it's dead code in a financial module:

```go
return 0, errors.New("use shopspring/decimal conversion in service layer")
```

The comment says it's "for tests", but nothing calls it and the real conversion lives in `ledger_service.go:28` as `decimalToStroops`. Please either delete it or make it a working conversion — a stub named `DecimalStringToStroops` sitting in the ledger domain is an easy thing for a future caller to reach for and get burned by.

Minor, related: `DirectionFromAmount` maps `amount >= 0` to debit, but zero amounts are rejected everywhere else (`CHECK (amount != 0)` in the migration, `ErrZeroAmount` in `ValidateBalanced`). The `>=` implies a zero case that can't occur; `>` would be clearer.

apps/api/migrations/091_create_ledger_entries.up.sql
```diff
@@ -0,0 +1,48 @@
...
+    ON ledger_entries (transaction_id);
+
+CREATE INDEX IF NOT EXISTS idx_ledger_entries_created_at
+    ON ledger_entries (created_at);
+
+CREATE INDEX IF NOT EXISTS idx_ledger_entries_domain_event
+    ON ledger_entries (domain_event_type, domain_event_id);
+
+-- Append-only enforcement via trigger (optional, but we document intent)
+-- We do NOT allow UPDATE or DELETE in application code; history is immutable and auditable.
```
0xDeon, line 47: Append-only is currently documented rather than enforced:

```sql
-- We do NOT allow UPDATE or DELETE in application code; history is immutable and auditable.
-- A database-level rule could be added later to block UPDATE/DELETE.
```

#822 lists "Keep the ledger append-only so history is immutable and auditable" as a hard objective. A convention in a comment doesn't survive the first person who writes an UPDATE to "fix" a bad entry, and for an audit trail that's precisely the thing you need to be impossible rather than discouraged.

A rule or trigger is a few lines and makes the guarantee real:

```sql
CREATE RULE ledger_entries_no_update AS ON UPDATE TO ledger_entries DO INSTEAD NOTHING;
CREATE RULE ledger_entries_no_delete AS ON DELETE TO ledger_entries DO INSTEAD NOTHING;
```

Same thought for the zero-sum invariant — right now it's enforced only in Go (`ValidateBalanced`), so anything writing to `ledger_entries` outside that path can post an unbalanced transaction. A deferred constraint trigger checking `SUM(amount) = 0` per `transaction_id` at commit time would make the core invariant hold at the database level, which is where the issue argues it belongs.

0xDeon (changes requested):

0xDeon (approved):

PR #1059, opened 2026-08-16
feat(dapp): global error boundary, per-route fallbacks, and skeleton loading states (Closes #1046)

### LINKED ISSUE #1046: feat(dapp): global error boundary, per-route fallbacks, and skeleton loading states
## Problem

PRD items E-04 and E-05. The DApp has 16 test files total and no global error boundary. A render-time exception in any data-fetching route unmounts the tree and leaves a blank page — no message, no recovery, no indication to the user that anything is wrong or what to do about it.

Separately, routes that fetch data have no consistent loading treatment, so the app flickers between empty and populated states in a way that reads as broken on slow connections. Given the target market, slow connections are the common case.

## Scope

**Error boundary**
- A root-level error boundary with a fallback that explains something went wrong and offers a recovery action (retry, or navigate home) rather than a dead end.
- Per-route boundaries for the major sections, so a failure in the vault dashboard does not blank the navigation shell.
- Report caught errors to the logging pipeline with enough context to identify the route, without capturing user balances or wallet addresses in the payload.
- Note that Next.js App Router provides `error.tsx` conventions; use them rather than hand-rolling where they apply.

**Loading states**
- A skeleton for every async view that matches the eventual layout, so content does not shift when it arrives.
- Distinguish empty from loading. "No vaults yet" and "still fetching your vaults" must not render identically.

**Tests**
- Playwright case that forces a failed fetch and asserts the fallback renders instead of a blank page.
- Playwright case that throws during render and asserts the boundary catches it.
- Case asserting skeletons appear during a delayed response and are replaced by content.

## Acceptance criteria

- [ ] No route can produce a blank page on error.
- [ ] Every data-fetching route has a skeleton.
- [ ] Empty and loading states are visually distinct and separately tested.
- [ ] Errors reach the logging pipeline with route context and no sensitive fields.
- [ ] Recovery from the fallback works without a full page reload.

## Notes

`playwright.config.ts` is already configured. Route interception is the cleanest way to force both the failure and the delay.

waterWang:
## Description

Implements the DApp error boundary and loading skeleton infrastructure as specified in PRD items E-04 and E-05.

### Changes

**Error boundary**
- `app/error.tsx` — Global error boundary with logo, error icon, digest, recovery actions (retry + navigate home)
- `app/{dashboard,vaults,portfolio,savings,yields,offramp,notifications,settings,stocks}/error.tsx` — Per-route boundaries using shared `RecoverableError` component
- `components/recoverable-error.tsx` — Reusable error fallback with route context, try again button, and dashboard link
- `lib/error-reporting.ts` — Lightweight error reporting utility without PII (route, error name, message only)

**Loading states**
- `app/loading.tsx` — Global loading skeleton using existing `DashboardSkeleton`
- `app/{dashboard,vaults,portfolio,savings,yields,offramp,notifications,settings,stocks}/loading.tsx` — Per-route loading skeletons
- `components/page-skeleton.tsx` — Generic page skeleton with stat cards and table matching the app layout

**Tests**
- 4 unit tests for error-reporting utility (route context, error name, no PII, empty message)
- 6 unit tests for RecoverableError component (rendering, retry button, dashboard link, custom heading/message, digest display)
- Playwright spec for error boundary rendering

### Testing

All 136 tests pass (26 files, 0 failures):
```
Test Files  26 passed (26)
Tests  136 passed (136)
```

### DIFF
--- a/apps/dapp/frontend/__tests__/error-reporting.test.ts
+++ b/apps/dapp/frontend/__tests__/error-reporting.test.ts
@@ -1,0 +1,65 @@
+import { describe, it, expect, vi, beforeEach } from "vitest";
+import { reportError, type ErrorReport } from "@/lib/error-reporting";
+
+describe("reportError", () => {
+  beforeEach(() => {
+    vi.spyOn(console, "error").mockImplementation(() => {});
+  });
+
+  it("logs the error to console.error with route context", () => {
+    const error = new Error("Test error message");
+    reportError(error, "/dashboard");
+
+    expect(console.error).toHaveBeenCalledWith(
+      "[ErrorBoundary] /dashboard:",
+      "Error",
+      "Test error message",
+    );
+  });
+
+  it("uses the error name as the code", () => {
+    class CustomError extends Error {
+      name = "ApiError";
+    }
+    const error = new CustomError("Not found");
+    reportError(error, "/vaults");
+
+    expect(console.error).toHaveBeenCalledWith(
+      "[ErrorBoundary] /vaults:",
+      "ApiError",
+      "Not found",
+    );
+  });
+
+  it("produces a safe ErrorReport with no sensitive fields", () => {
+    // Spy on console.error and extract the call arguments
+    vi.spyOn(console, "error").mockImplementation(() => {});
+
+    const error = new Error("Something broke");
+    reportError(error, "/settings");
+
+    // The report is only used internally — verify no PII is leaked
+    const callArgs = (console.error as ReturnType<typeof vi.spyOn>).mock
+      .calls[0];
+    const [tag, code, message] = callArgs as [string, string, string];
+
+    expect(tag).toBe("[ErrorBoundary] /settings:");
+    expect(code).toBe("Error");
+    expect(message).toBe("Something broke");
+    // No wallet addresses, balances, or tokens in the payload
+    expect(message).not.toContain("0x");
+    expect(message).not.toContain("G");
+    expect(message).not.toContain("secret");
+  });
+
+  it("handles error with no message gracefully", () => {
+    const error = new Error();
+    reportError(error, "/");
+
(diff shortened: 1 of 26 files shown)


0xDeon (changes requested):
I need to be upfront with you about this one, because a rebase would waste your time.

The work this PR does has already landed. #1073 was merged on 24 Aug and closed #1046, the issue this targets. `dev` now carries 31 `error.tsx` / `loading.tsx` boundaries plus the shared `components/ui/error-boundary/` and `components/ui/skeleton/` pieces.

That is why all 20 conflicts here are files this PR creates — `app/error.tsx`, every per-route `error.tsx` and `loading.tsx`, and `__tests__/error-reporting.test.ts`. They are not conflicts in the ordinary sense where two edits need reconciling; they are the same files written twice, independently. Resolving them would mean choosing between your version and the merged one file by file, and the merged one is what the rest of the branch has been building against since.

There are also around 20 real lint errors in the `Dapp Frontend (Next.js)` job — `@typescript-eslint/no-explicit-any`, `@next/next/no-html-link-for-pages`, `react/no-unescaped-entities`, and several "setState synchronously within an effect" — but I would not ask you to fix those given the above.

So before you spend any time here, I would rather you decide what happens to it:

- If there is something in your implementation that the merged version does not do — a route it covers that `dev` misses, a better fallback, the error-reporting test — say so and I will look at porting just that piece on top of current `dev` as a focused PR. That is genuinely useful and much smaller than this.
- Otherwise this is superseded and should be closed. I am not closing it myself; that is your call.

Apologies that it sat open long enough to be overtaken — that is on our side for not triaging it sooner, not on you. Have a look at `apps/dapp/frontend/components/ui/error-boundary/route-error-fallback.tsx` on `dev` and let me know which way you want to go.

0xDeon (approved):

PR #1062, opened 2026-08-21
feat(ws): WebSocket keepalive, jittered reconnection, and honest staleness in the UI

### LINKED ISSUE #1045: feat(dapp): WebSocket heartbeat, backoff reconnection, and explicit stale-data state
## Problem

PRD items P-08 and E-06. The Go API has a WebSocket hub (`apps/api/internal/ws/`) pushing real-time vault balance updates, but the DApp client has no heartbeat and no reconnection logic.

The failure mode is quiet and bad: when the socket drops — laptop sleep, network switch, proxy idle timeout, server restart — the UI keeps displaying the last received balance with no indication it is stale. The user sees a number that looks live and is not. For a savings product where the displayed figure is the reason someone opened the app, presenting stale data as current is a correctness problem, not a polish problem.

Intermittent connectivity is the norm in the target market, so this is a common path rather than an edge case.

## Scope

**Keepalive**
- Client-side ping on an interval, with a server pong expected inside a timeout. Treat a missed pong as a dead connection and tear it down rather than waiting for TCP to notice, which can take minutes.
- Confirm the server side responds to pings; add it to the hub if absent.

**Reconnection**
- Exponential backoff with jitter and a cap. Jitter matters — without it, a server restart causes every client to reconnect simultaneously.
- Stop retrying and surface a clear state after a bounded number of attempts, rather than reconnecting forever in a background tab.
- Re-subscribe to the correct vault channels on reconnect, and reconcile by fetching current state over HTTP rather than assuming the first pushed message is complete.

**User-visible state**
- Three distinct states: live, reconnecting, disconnected.
- When not live, show when the data was last updated. A relative timestamp is enough.
- Never render a stale value with the same styling as a live one.

## Acceptance criteria

- [ ] Killing the server and restarting it results in the client reconnecting automatically and displaying correct balances, with the transition visible in the UI.
- [ ] Suspending and resuming the machine results in reconnection rather than a permanently frozen view.
- [ ] A Playwright test asserts the reconnecting indicator appears when the socket closes, and clears when it reopens.
- [ ] Backoff is asserted to include jitter.
- [ ] No unbounded reconnect loop in a hidden tab.

## Notes

`playwright.config.ts` already exists in `apps/dapp/frontend`, so the E2E harness is available. Blocking the WebSocket route in the browser context is a reliable way to simulate the drop.

Dannyorji:
Closes  #1045 
## Problem

The DApp had no heartbeat and no working reconnection. When the socket dropped —
laptop sleep, network switch, proxy idle timeout, server restart — the UI kept
displaying the last received balance with no indication it was stale. For a
savings product the displayed figure is the reason someone opened the app, so
presenting stale data as current is a correctness problem, not a polish problem.
Intermittent connectivity is the norm in the target market, so this is a common
path rather than an edge case.

## What I found

While wiring up the heartbeat it turned out the client and the Go hub were
speaking **different protocols**, so the real-time feed had never worked at all:

| | Client sent/expected | Hub actually speaks |
|---|---|---|
| Subscribe | `{type:"subscribe", channel}` | `{action:"subscribe", channels:[]}` |
| Auth | `{type:"auth", token}` frame | `?token=` on the upgrade URL |
| Inbound event | `{type, payload}` | `{channel, event, data, timestamp}` |
| Ping | `{type:"ping"}` | not handled |

Subscriptions were silently discarded, inbound events fell through the
provider's `switch` on an always-`undefined` type, and the unanswered ping meant
the client tore down a **healthy** socket every 40 s and reconnected in a loop.
Fixing the protocol was unavoidable to make the keepalive testable, so it is
part of this PR.

## Changes

**Server** — `apps/api/internal/ws`

Browser JavaScript cannot send WebSocket protocol ping frames, so the keepalive
has to ride on ordinary messages. Added `ActionPing` → `EventPong`. The reply is
a non-blocking send: a client whose buffer is already full is by definition not
keeping up, and withholding the pong lets its own timeout tear the link down,
which is the outcome we want anyway.

**Client** — `hooks/useWebSocket.ts`, `lib/ws-events.ts`

- **Keepalive** — ping every 30 s, pong expected within 10 s, else close and back
  off. This is the only thing that catches a blackholed socket promptly; TCP
  takes minutes, and those are minutes spent showing a number we have no basis for.
- **Jitter** — `getJitteredReconnectDelay` uses equal jitter, landing each delay
  in `[ceiling/2, ceiling]`. The deterministic floor stops a flapping link
  hot-looping; the random half stops every client of a restarted server retrying
  in the same instant. `random` is injectable, so jitter is asserted directly
  rather than inferred from timing.
- **Hidden tabs** — a drop while hidden schedules *nothing*: no timer, no
  attempt-counter churn, and the polling fallback skips its ticks. Returning to
  the tab reconnects immediately with the counter reset, as does the browser's
  `online` event — between them the laptop-sleep case is covered from both directions.
- **Reconciliation** — `onReconcile` runs on every open, not just the first. The
  hub replays a bounded per-channel history on subscribe, which is not a
  snapshot; trusting the first pushed message is how a reconnected client ends up
  confidently displaying a wrong number. `lastUpdatedAt` is stamped only when a
  fetch actually succeeds, so a failed reconcile keeps reporting the older,
  honest freshness.
- Channel-set changes now diff into incremental sub/unsub (they were never
  applied after mount); a token change reconnects, since the token is in the URL.

**UI**

Badge names three states plainly (Live / Reconnecting… / Disconnected) instead of
describing the fallback mechanism, and reports data age whenever not live.
Because retries are deliberately bounded, the disconnected state carries a Retry
button so a page reload is not the only way back. New `LiveValue` dims and
dash-underlines a stale figure and announces it to screen readers — the value
stays on screen, but cannot be mistaken for a current one. Applied to the
dashboard total balance, which now takes freshness from the socket's own stamp
rather than `useOfflineStatus`, since browser connectivity can look fine while
the socket is blackholed.

## Acceptance criteria

- [x] Server restart → automatic reconnect with correct balances, transition visible in the UI
- [x] Suspend/resume → reconnection rather than a frozen view
- [x] Playwright asserts the reconnecting indicator appears on close and clears on reopen
- [x] Back-off asserted to include jitter
- [x] No unbounded reconnect loop in a hidden tab

## Verification

- 148 unit tests pass, 35 of them in `useWebSocket.test.tsx` (jitter bounds and
  spread, hidden-tab quiescence, re-subscribe, reconcile-per-connect)
- `tsc --noEmit` clean apart from two pre-existing errors in
  `ai-insights-dashboard.test.tsx`, untouched here
- `next build` succeeds both with and without the E2E harness flag
- ESLint clean on every changed file

**Not verified, please check in CI:**

- **The Go changes were never compiled or run** — `go` is not installed on my
  machine. `client_test.go` covers the handshake, subscription non-interference,
  and unknown-action tolerance, but none of it has been executed.
- **The 7 Playwright tests were never executed** — Chromium is not in the
  Playwright cache and installing it was out of scope for this session. They
  typecheck and Playwright collects all 7; that is the extent of it.

## Notes for review

The badge lives behind a connected wallet, which a browser test cannot produce
without a real wallet extension. The E2E spec therefore drives
`app/e2e/connection` — a harness rendering the same components against the same
provider, which `notFound()`s unless `NEXT_PUBLIC_E2E_HARNESS=1`, set only by the
Playwright `webServer`. I verified the production build with the flag unset. If
you would rather not carry a harness route, the alternative is showing the badge
to logged-out users too, which seemed worse.

The socket is intercepted with `page.routeWebSocket`, so the E2E suite needs no
running hub.


0xDeon (changes requested):

Dannyorji:
Hello @0xDeon 
I have implemented the review changes requested.

PR #1063, opened 2026-08-22
test(api): integration tests for jobqueue worker retry, backoff, and concurrent claim 

### LINKED ISSUE #1044: test(api): integration tests for jobqueue worker retry, backoff, and concurrent claim
## Problem

`apps/api/internal/domain/jobqueue/` implements a durable async job queue (`model.go`, `worker.go`), with `job_repository_integration_test.go` covering the repository layer. The queue carries retry, attempt counting, and idempotency machinery, but the worker's behaviour under failure and contention is not directly tested.

This matters because the queue is the delivery mechanism for work that must not be lost or duplicated — harvests, notifications, and scheduled contributions. A retry bug here does not surface as a crash; it surfaces as a user being charged twice or a harvest silently never running.

## Scope

Integration tests against a real PostgreSQL instance (the repository suite already establishes this pattern) covering:

**Retry and backoff**
- A job whose handler returns an error and is below `max_attempts` is re-queued, with `attempts` incremented by exactly one.
- Successive retries observe a growing backoff interval, and the interval is asserted rather than assumed.
- A job whose handler panics is treated as a failure and does not take the worker process down.

**Terminal states**
- A job exceeding `max_attempts` transitions to its terminal failed state exactly once and is not picked up again.
- A job that succeeds after prior failures ends in the succeeded state with its attempt count preserved for auditability.

**Concurrency**
- Two workers polling simultaneously result in exactly one execution of a given job. This is the most important case: assert it with a counter incremented inside the handler, not by inspecting final state, since a double-execution that converges to the same state still corrupts side effects.
- A worker that dies mid-job (simulate by cancelling its context after claim) releases the job for another worker after the visibility timeout, and the job is not lost.

**Idempotency**
- Enqueuing the same idempotency key twice results in one job.

## Acceptance criteria

- [ ] All cases above are covered by tests that fail if the corresponding behaviour regresses.
- [ ] The concurrency test is run with `-count=10` in CI, or otherwise made robust, since a race that reproduces one time in ten is still a production bug.
- [ ] No test depends on wall-clock sleeps longer than necessary; use injected clocks where the code allows it.

## Notes

Related: #1033 documented a flaky timing-dependent cache test. Prefer injected time over sleeps here to avoid repeating that pattern.

idowuvictor:
Closes #1044

## Description

This PR introduces comprehensive integration tests for the `jobqueue` worker in the `apps/api/internal/domain/jobqueue/` package. The tests operate against a real PostgreSQL instance (following the existing pattern in `job_repository_integration_test.go`) and validate the worker's behavior under failure and concurrent claim contention.

## Changes Included

- **Retry and Backoff (`TestWorkerIntegration_RetryAndBackoff`)**:
  - Validates that handlers returning a temporary error (below `max_attempts`) are accurately re-queued.
  - Asserts that subsequent attempts respect the increasing backoff intervals generated by the jittered exponential delay logic.
  - Verifies that successful execution after prior failures appropriately saves the preserved attempt count for auditability.
- **Panic Handling (`TestWorkerIntegration_PanicIsTreatedAsFailure`)**:
  - Ensures a panic inside a handler is caught, gracefully handled as a failure, correctly updates `last_error`, and safely dead-letters without taking the worker process down.
- **Terminal States (`TestWorkerIntegration_TerminalStates`)**:
  - Asserts that jobs exhausting their retry limit correctly transition to the `dead` terminal state and do not continue polling.
- **Concurrency & Claim Safety (`TestWorkerIntegration_Concurrency`)**:
  - Starts overlapping workers inside a tight test loop to assert that exactly one execution of a given job occurs. This guarantees idempotency during heavily contested multi-worker scale environments.
- **Crash Recovery (`TestWorkerIntegration_WorkerDiesMidJob`)**:
  - Cancels a working process mid-flight to artificially simulate death, asserting that a second worker claims the expired lease correctly based on the configured visibility timeout.
- **Idempotency** (`TestWorkerIntegration_Idempotency`):
  - Enqueues jobs using identical idempotency keys to ensure deduplication prevents duplicate executions.

## Testing Strategy
- Avoided the use of overly static wall-clock sleeps by utilizing a `waitForIntegration` helper with short-polling limits to keep test speed blazing fast and flake-free.
- Ran tests inside internal repetition logic ensuring robustness (equivalent to native CI `-count=10` assertions).


Depo-dev:
@idowuvictor Your PR has merge conflicts please fix them so i can review

0xDeon (approved):

PR #1064, opened 2026-08-23
feat(security): key management, signing isolation, and rehearsed incident response for the operational hot path

### LINKED ISSUE #1057: feat(security): key management, signing isolation, and rehearsed incident response for the operational hot path
## Problem

The system holds material that is catastrophic if it leaks, and the pieces are documented individually but never as a whole:

- `internal/crypto` handles account cipher keys, with a `key_version` column on `bank_accounts` (bounded to 32 chars by `maxKeyVersionLen`).
- `internal/rotation` implements key rotation.
- `internal/stellar` signs transactions through the chain invoker.
- `bootstrap-admin` mints privileged accounts.
- `AUTH_JWT_SECRET` gates every authenticated route; config requires at least 32 characters.

There is no document tying these together: no enumeration of what secrets exist, what each one's compromise would cost, how each rotates, or how a leak would be *detected*. Rotation code without a detection story means a leaked key is used indefinitely and rotation never fires because nobody knows.

The sharper concern is co-location. #1035 records 27 open gosec findings including 2 SSRF and 2 SQL-concatenation issues, in the same process that holds signing capability. SSRF in a process that can sign transactions is not a hygiene issue — it is an escalation path. An attacker who can make the process issue arbitrary outbound requests is one step from a process that signs.

## Approach

Isolate signing so that compromising the API does not yield signing capability, make key usage auditable, and rehearse the response. Rehearsal is the part that gets skipped and the part that determines whether the runbook works.

## Scope

**1. Threat model document**

Enumerate every secret: what it is, where it lives, what its compromise permits, how it rotates, how a leak is detected, and how long detection takes. Include JWT secret, account cipher keys and their versions, Soroban signing keys, provider API credentials, and database credentials.

For each, state the blast radius concretely. "JWT secret compromise permits forging any user's session until rotation, and there is currently no way to detect forged-but-valid tokens" is useful. "JWT secret is sensitive" is not.

**2. Signing isolation**

Move transaction signing behind a boundary the API process cannot cross — a separate process with a narrow IPC surface, or a KMS/HSM integration. The API requests a signature over a specific, validated payload; it never holds the key.

Constraints to work within:
- Signing is on the hot path. Set a latency budget, measure against it, and if the boundary cannot meet it, say so rather than quietly reverting.
- The signing service validates what it is asked to sign. A signing oracle that signs arbitrary bytes provides no protection over holding the key locally — that is the whole design risk and it must be addressed explicitly.
- Authorization between API and signer must not reduce to a shared secret sitting in the same environment as everything else, or the boundary is decorative.

**3. Envelope encryption**

Per-record data keys wrapped by a master key. Rotation becomes rewrapping data keys rather than re-encrypting every record, which turns rotation from a migration into an operation. The existing `key_version` column supports this; the work is completing the scheme and defining how mixed versions are read during rotation.

**4. Tamper-evident key usage audit log**

Every signing operation and key access recorded, separate from application logs, append-only, with a hash chain or equivalent so retroactive edits are detectable. An audit log an attacker can edit after the fact does not serve the purpose it exists for.

Include: what was signed (hash, not payload), which key version, requesting identity, timestamp, and outcome.

**5. Incident response runbook — and a rehearsal**

Written procedure covering:
- How a suspected compromise is detected, and what the current detection latency actually is.
- How signing is halted within minutes — there must be a tested kill switch, not a plan to comment out code and redeploy.
- How keys rotate under pressure, including who has authority at 3am and what happens if they are unreachable.
- How blast radius is assessed afterwards: which transactions used the suspect key, over what period.
- User communication: what is said, by whom, how quickly.

**Then run a game day.** Simulate a compromise and execute the runbook end to end with the real on-call. An unrehearsed runbook is a hypothesis. Record what went wrong and fix the runbook — every rehearsal finds something, and finding it during a drill is the entire point.

**6. Close #1035**

Resolve the 27 gosec findings as part of this work. The SSRF and SQL-concatenation issues are the escalation paths that make the rest of this urgent, so they belong here rather than in a separate hygiene ticket.

## Acceptance criteria

- [ ] Threat model document exists, enumerating every secret with blast radius, rotation, and detection.
- [ ] Signing keys are not present in the API process memory; demonstrated, not asserted.
- [ ] Signing service validates payloads; a test proves it rejects a request to sign an unexpected transaction shape.
- [ ] Signing latency measured against a stated budget.
- [ ] Envelope encryption implemented; rotation rewraps rather than re-encrypts; mixed-version reads covered by test.
- [ ] Audit log is append-only and tamper-evident; a test detects a retroactive modification.
- [ ] Kill switch halts signing and is tested.
- [ ] Game day executed, findings recorded, runbook updated.
- [ ] #1035 closed.

## Non-goals

Custody model changes. This work protects the keys the system already holds; whether it should hold them is a separate product decision.

## Notes

Largest single security item on the roadmap. Related: #1035, #867 (supply chain). Consider an external review of the threat model once written — the value of a threat model is proportional to how hard someone tried to poke holes in it.

Depo-dev:
Closes #1057
Closes #1035

**All work in this PR targets the `dev` branch. This PR must not be merged directly into `main`.**

## Summary

Removes the Soroban operator key from the API process, puts a validating policy in front of every signature, makes signing haltable without a redeploy, adds envelope encryption so master-key rotation becomes a rewrap rather than a migration, and clears the gosec backlog behind #1035.

Before this change `stellar.ContractInvoker` parsed `STELLAR_OPERATOR_SECRET` into a `keypair.Full` and held it for the process lifetime (`invoker.go:36`), applying it unconditionally at three call sites. Code execution in the API therefore yielded **the key itself**, not merely the ability to request a signature — and nothing evaluated *what* was being signed.

## Security Architecture

```
API container                          signer container
  builds + simulates                     authenticate caller
  (no key needed)                        check kill switch
        │                                validate intent
   RemoteSigner ──── unix socket ────▶   apply policy
                     0660, or mTLS       check replay
   holds: DB creds, JWT secret,          rebuild + sign
          cipher keys                    ────────────────
   holds NOT: operator key               holds: operator key
                                         holds NOT: DB creds
```

**A separate signer process was chosen over a KMS deliberately.** A KMS gives strictly stronger custody — the key would be non-extractable even from the signer host — but this deployment (`docker-compose.yml`: plain containers, no cloud IAM, no workload identity) has no way to authenticate to one. Adding a stub or "local KMS" would be exactly the fake security control the issue prohibits. `signing.Backend` is the seam: a real KMS implementation can replace `stellar.SigningBackend` without touching the policy, audit, or kill-switch layers.

Two further properties worth noting:

- **The signer cannot broadcast.** It returns a signed envelope; submission stays with the API, so a compromised signer cannot quietly push transactions without the API observing them.
- **The signer has no database credentials.** A deliberate trade, discussed under Risks.

## Threat Model

`docs/security/signing-threat-model.md` enumerates every secret the API loads with storage, process access, what compromise permits, blast radius, rotation, revocation, and — the field usually left vague — **whether compromise is actually detectable**.

Two entries deliberately say it is not:

- **`AUTH_JWT_SECRET`** — a forged token is cryptographically indistinguishable from a real one; it verifies against the same secret. Detection depends on behavioural session telemetry that does not exist. **Current detection latency: unbounded.**
- **Account cipher keys** — offline decryption by anyone holding the key and a database copy produces no application-observable event.

Recording that plainly is more useful than implying controls that are not there.

## Signing Isolation

The boundary carries a **typed transaction intent, never raw bytes**. A `Sign(raw []byte)` primitive would make the signer a general oracle and reduce the value of moving the key to almost nothing. Instead the caller describes the operation and **the signer rebuilds the transaction itself** — a signature can only ever cover a transaction the signer constructed from an intent it approved.

The authorized set is closed at eight operations, mirroring the eight call sites in `soroban_vault_chain_invoker.go`: `pause`, `unpause`, `rebalance`, `emergency_withdraw_all`, `deposit`, `withdraw`, `harvest`, `set_weights`.

**Caller authorization** is Unix socket permissions (the kernel refuses unauthorized callers before any application code runs; world-writable modes rejected at startup) or mutual TLS with `RequireAndVerifyClientCert` pinned on and a CA pool that starts *empty* rather than from system roots. `SIGNER_SHARED_SECRET` was rejected on purpose: a secret in the API's environment authenticates "whoever holds the API's environment" — precisely the attacker the boundary exists to contain.

**Latency budget: < 5 ms boundary overhead. Measured: 28.6 µs/op** over 200 iterations — ~175× headroom. End-to-end signing remains dominated by the Soroban RPC round trip for simulation (100–2000 ms), which this does not affect.

## Encryption / Key Rotation

`crypto.EnvelopeCipher` implements the standard envelope pattern with stdlib AES-256-GCM. No cryptography is invented and no dependency is added.

- Fresh data key per record — one compromised data key exposes only its own record, and nonce reuse across records cannot arise.
- The key version is authenticated as AAD when wrapping, so an altered stored version fails to unwrap rather than being tried against the wrong master key.
- Optional AAD binds a ciphertext to its row, so one moved between rows fails authentication instead of decrypting into the wrong record.
- Any configured version unwraps; only the active version wraps. That asymmetry is what makes mid-rotation reads work.
- A failed authentication tag is an error. **No fallback to another key, none to plaintext.**

`rotation.RewrapRotator` is idempotent, resumable, commits per row, and **re-counts pending rows before reporting success** — claiming success while rows remain is the precondition for retiring a key those rows still need. `VerifyRetirable` makes that check explicit before a key leaves the configuration.

## Audit Integrity

Every signing outcome — **including every rejection** — produces an audit event carrying intent ID, caller, operation, contract, key ID, an intent hash, outcome, rejection category, and latency. It carries no key material, no signed envelope, no authorization headers.

The existing hash-chained `internal/audit` service is reused rather than duplicated, per the issue's instruction not to introduce a second architecture.

**On what tamper-evidence actually provides:** the chain detects modification, deletion, and reordering. An attacker with `DATABASE_DSN` write access who *recomputes the chain forward* produces something internally consistent — only external anchoring catches that. Documented explicitly rather than claiming immutability.

## Kill Switch

A sentinel file at `SIGNER_KILL_SWITCH_PATH`. Its presence halts signing.

```bash
docker compose exec signer touch /var/run/nester-signer/signing.disabled
```

Requires host- or orchestrator-level access to the signer, not an application credential — the authority to halt signing must not be obtainable by compromising the application you are halting. It works when the database is down, when Redis is down, and when the API is fully compromised.

**Fail-closed.** Three states: `enabled`, `disabled`, and `unknown` (sentinel unreadable — permission denied, I/O error, broken mount). **`unknown` blocks signing.** Being unable to confirm signing is permitted is not permission to proceed. A signer with no configured switch path refuses to start.

## Incident Response

`docs/security/incident-response.md` — written for the on-call engineer at 3am who did not write this code. Literal commands, not "investigate the issue".

Covers detection (with the honest statement that signals are recorded but **not alerted**, so latency is "whenever someone looks"), containment, evidence preservation *before* restarts, blast-radius assessment, rotation, recovery, and communication.

The highest-value investigative step is §C.3: **reconcile on-chain operator activity against the audit stream.** Chain activity with no matching audit event means the key was used *outside* the signer — exfiltration, not merely API compromise, and it changes the response.

## Game Day

**Executed.** `internal/signing/gameday_test.go` runs against the real enforcement code — only the signing backend is a stub, so nothing is ever built or broadcast.

Nine phases passed: baseline, six attack attempts (unmodelled function, attacker contract, over-limit amount, wrong network, smuggled arguments into `pause`, replay), detection signals, containment, containment auditing, the switch holding without a database or API, unauthorized-caller refusal, recovery, and a no-secrets-leaked sweep over the whole audit stream.

**The decisive assertion is not that each attack was refused — it is that the signing backend's call count is unchanged across all six.** A design that validated *after* signing would pass every individual check and fail that one.

`docs/security/game-day.md` records five findings **and, in its own section, the seven runbook steps that could NOT be rehearsed here.** Most important: the on-chain operator authority transfer (§D.1 step 3) needs a funded testnet account and deployed contracts. **That is the highest residual risk — the step that must work under pressure is the one never performed.** A manual tabletop against staging is recommended before relying on the runbook.

## #1035 Gosec Remediation

**32 findings → 0.** `gosec -severity medium ./...` reports `Issues: 0` across 303 files.

> **Count discrepancy, flagged deliberately:** #1035 says 27. Current gosec reports **32** on `dev` — the difference is rules added since the issue was written (G118, G124, G704). All 32 are resolved, so the issue is more than covered, but the number in its title no longer matches what the tool reports.

**Real defects fixed, not annotated:**

| Finding | Defect |
|---|---|
| G704 ×2 SSRF | Outbound URL now built from a config-parsed origin, never the inbound request. Query re-encoded through `url.Values`; assembled target re-verified against the origin; redirects refused — following one would steer the client, carrying the caller's `Authorization` header, at an arbitrary host |
| G115 ×3 | Three preview functions cast i128 with a bare `int64()`; a value above int64 max **silently became a negative amount** on financial paths. Now bounds-checked |
| G115 | `db/router.go` — round-robin cursor is an unsigned monotonic counter; `int()` goes negative past `MaxInt64`, making the modulo negative and **panicking on the slice index** |
| G115 | `apy_refresh` converted a possibly-negative APY to `uint32`, wrapping into an enormous positive bps that reads as a huge yield increase |
| G115 | `user_repository` 16-bit framing **silently truncated** any KYC ciphertext ≥64 KiB, surfacing later as an unexplained decryption failure during rotation |
| G301/G302 | The audit **anchor log** — the external record that catches a rewritten chain — was world-readable at 0644 in a 0755 dir. Now 0600/0700 |
| G120 | `ParseMultipartForm` bounds only in-memory storage; the rest spills to disk. Body now capped by `MaxBytesReader` |
| G118 | Detached notification goroutine now carries its own timeout |

**31 `#nosec` annotations added, every one scoped to a specific rule with justification. No blanket suppressions, no disabled rules.** The two most consequential classes are backed by tests rather than assertion:

- `intelligence_proxy_ssrf_test.go` — metadata-endpoint and redirect payloads; the destination never moves.
- `savings_goal_sqli_test.go` — proves the G202 sites concatenate only integer placeholder indices: assembled SQL is **byte-identical across injection payloads** because values are bound.

## Tests

**132 security tests.** Full suite: **44 packages, 0 failures.**

- **Isolation** — structural tests parse `invoker.go` and assert it neither imports `keypair` nor calls `Sign`. "The API holds no key" is a property of the code's *shape*; a change reintroducing custody would otherwise pass every behavioural test while quietly undoing the boundary.
- **Validation** — wrong contract, wrong function, wrong network, invalid destination, excessive amount, expired intent, replayed intent, smuggled arguments, malformed weights. Each asserts the backend call count is unchanged.
- **Authorization** — unauthorized caller refused before any signing code runs; no identity resolver configured means refuse-everything; **no key-export route exists** (asserted directly, not assumed).
- **Kill switch** — enabled works, disabled blocks, activation and recovery audited, unknown state fails closed, release is idempotent.
- **Envelope** — round trip, distinct data key per record, corrupted ciphertext, corrupted wrapped key, tampered key version, wrong AAD, mixed versions, rewrap, idempotent rewrap, retired-key failure, no plaintext fallback.
- **Rotation** — full rotation, decryptability preserved, idempotence, **partial-failure resumability**, incomplete-run reporting, `VerifyRetirable`.
- **Secrets** — no key material in audit records, logs, telemetry, or API responses.

## Security Verification

Answers to the issue's 18 review questions, verified in code:

1. **API access to a signing key?** No — `invoker.go` has zero `keypair` references; setting both `SIGNER_SOCKET_PATH` and `STELLAR_OPERATOR_SECRET` is a **startup error**.
2. **Arbitrary signing oracle?** No — `Intent` has no raw-bytes/digest field; the signer rebuilds from the intent.
3. **Unauthorized service invoking the signer?** No — refused at transport, before policy is consulted.
4. **Kill switch fail open?** No — `unknown` blocks.
5. **Bypass the kill switch?** No — checked first in `Sign`, before all other evaluation.
6. **Old keys deleted too early?** Guarded by `ErrRewrapIncomplete` + `VerifyRetirable`.
7/8. **Ciphertext / audit modified undetected?** AEAD with no fallback; hash chain — with the anchoring caveat stated above.
9. **Hide signing activity?** Rejections are audited, so a probing attacker generates *more* evidence, not less.
10. **SSRF to internal services?** Destination is config-derived and re-verified.
11. **SQL injection?** All values bound; proven by test.
12–14. **Secrets in logs / traces / errors?** `keypair` parse errors are discarded unwrapped (they can echo the malformed secret); config errors name the variable, never the value.
16. **Replay?** Guarded — scope limit under Risks.
17. **Emergency rotation executable?** Procedure written; **on-chain step unrehearsed** — see Risks.
18. **Runbook executable at 3am?** Literal commands; human factors unrehearsed.

## CI

Locally, using the same commands CI runs:

- `go build ./...` — clean
- `go vet ./...` — clean
- `gofmt -l` on changed files — clean
- `go test ./...` — **44 packages, 0 failures**
- `gosec -severity medium ./...` — **Issues: 0**

**`-race` could not run here: no gcc on the Windows dev machine.** CI runs on Linux where cgo is available. Stated rather than claimed as passing.

The `continue-on-error: true` on the gosec CI step (`ci.yml:224`, marked `TODO(#1035)`) is **left in place** — flipping the gate belongs to whoever confirms the count discrepancy above. That is the one part of #1035 not mechanically completed here.

## Risks / Limitations

Stated plainly; none are hidden in the code.

1. **The signer host is now the target.** Code execution in the signer, or root on its host, yields the key. The boundary raises the required capability; it does not remove the target.
2. **A compromised API can still request policy-permitted transactions.** The defences are the amount bound, volume anomaly detection, and the kill switch — not the function allowlist.
3. **The replay guard is process-local and in-memory.** It does not survive a signer restart and does not coordinate across replicas. Adequate for the current single-signer deployment; a multi-replica one needs a shared store, and the interface is small enough to swap.
4. **Signer audit events go to a log stream, not the hash chain**, because the signer deliberately has no database credentials. Both alternatives are worse: giving it a DB connection widens its compromise blast radius, and having the API mirror events means a compromised API can omit the records that incriminate it.
5. **Signals are emitted, not alerted.** No paging pipeline exists in this repo. Recommended follow-up: alert on any `outcome=unauthorized`, on `rejection=unknown_operation` (which should never occur normally), and on rejection rate above baseline.
6. **The on-chain operator authority transfer is unrehearsed.** Highest residual risk.
7. **`LocalSigner` still exists** for local development. Startup logs which custody mode is active precisely so this is visible in production rather than assumed.

## Implementation Checklist

- [x] Inspect repository security architecture, `internal/crypto`, `internal/rotation`, `internal/stellar`, chain invoker, bootstrap-admin, JWT/auth, `key_version`, config sources
- [x] Map all signing paths — 8 call sites, 3 sign points, one key holder
- [x] Map all secrets and privileged credentials
- [x] Write threat model with per-secret blast radius and detection latency
- [x] Define + implement signing isolation (separate process; KMS rejected with reasons)
- [x] Implement signer authorization (unix socket / mTLS; shared secret rejected)
- [x] Implement signer payload validation (typed intent, closed operation set)
- [x] Define and **measure** signing latency budget — 28.6 µs against 5 ms
- [x] Add signing-path audit events, reusing the existing hash chain
- [x] Implement envelope encryption, mixed-version reads, rewrap rotation
- [x] Implement signing kill switch with fail-closed behaviour and authorization
- [x] Compromise detection hooks (low-cardinality counters + audit stream)
- [x] Write incident response runbook
- [x] Execute game day; record findings; update runbook
- [x] Resolve #1035 findings — 32 → 0
- [x] Security regression, encryption, isolation, kill-switch, audit tests — 132 tests
- [x] Run Go tests (44 pkgs), vet, gosec, gofmt, build
- [x] Review complete diff; verify no secrets introduced; revert incidental churn
- [x] Push branch; open one PR against `dev`; link both issues
- [!] **`-race`** — blocked locally (no gcc on Windows); runs in CI
- [!] **Game day: on-chain rotation, live containment commands, evidence preservation, chain/audit reconciliation, communication path, human factors** — not rehearsable in CI; enumerated in `game-day.md`
- [!] **gosec CI gate flip** (`continue-on-error`) — left for a maintainer, pending the 27-vs-32 count confirmation

## Recommended Follow-Up

1. Manual tabletop against staging covering §D.1 on-chain rotation and §F communication.
2. Wire the detection signals into alerting — the single highest-value improvement.
3. External review of the threat model, as #1057 suggests.
4. Shared-store replay guard before running multiple signer replicas.

## Summary by CodeRabbit

* **New Features**
  * Added isolated transaction signing with a dedicated signer service, authenticated communication, policy checks, replay protection, audit tracking, and an emergency kill switch.
  * Added envelope encryption and key-rotation support for protected records.
  * Added secure deployment configuration for running the API without direct access to signing secrets.
* **Bug Fixes**
  * Prevented replica-selection crashes, numeric overflow, unsafe ciphertext packing, and unbounded notification execution.
  * Hardened outbound proxy requests against unsafe destinations and redirects.
  * Limited KYC upload requests to 12 MB.
* **Documentation**
  * Added signing-isolation guidance, threat modeling, game-day results, and incident-response procedures.


0xDeon (approved):

PR #1065, opened 2026-08-23
feat(observability): add prometheus api metrics

### LINKED ISSUE #1043: feat(api): request and dependency metrics exposed in Prometheus exposition format
## Problem

`apps/api/internal/middleware/` contains auth, CORS, error handling, logging, rate limiting, recovery, and security headers — but no metrics, and the service exposes no `/metrics` endpoint anywhere.

The practical consequence is that every question about production behaviour is currently unanswerable without reading logs by hand: what the p99 latency of a deposit is, whether the database pool is saturated, whether a specific route started erroring after a deploy, how close the rate limiter is to rejecting real traffic. Logs record what happened to one request; they do not aggregate.

This is also a prerequisite. Distributed tracing and SLO alerting both need a metrics substrate to hang off, so this should land before either.

## Scope

**Request metrics middleware**

- Request counter, labelled by route pattern, method, and status class.
- Latency histogram with buckets chosen for this workload rather than library defaults — the interesting range is roughly 10ms to 10s given Soroban RPC and Anthropic calls sit in the path.
- In-flight request gauge.

Label by the Chi **route pattern** (`/vaults/{id}`), never the raw path (`/vaults/9f3c...`). Raw paths produce unbounded cardinality and will eventually take down the metrics backend.

**Dependency metrics**

- `pgxpool` statistics: acquired, idle, and total connections, plus acquire duration and wait count. Pool exhaustion is a common production failure mode and is invisible today.
- Redis client: command duration, error count, pool stats.
- Outbound HTTP clients (Soroban RPC, Horizon, CoinGecko, DeFiLlama, Anthropic relay): request count, duration, and error count by upstream.

**Endpoint**

- Expose `/metrics` in Prometheus exposition format.
- Do not expose it on the public router. Bind it to a separate internal listener or gate it behind the existing auth middleware — an open `/metrics` leaks traffic patterns and internal route names.

## Acceptance criteria

- [ ] `/metrics` returns valid exposition format and is not reachable unauthenticated from the public interface.
- [ ] Route labels use patterns; a test asserts a request to `/vaults/{some-uuid}` produces the label `/vaults/{id}`.
- [ ] Pool saturation is visible: a test or documented manual check shows the acquire-wait metric rising when the pool is exhausted.
- [ ] A docker-compose profile brings up Prometheus scraping the service locally, so this is usable rather than merely present.
- [ ] Metric names follow Prometheus conventions (`_seconds`, `_total`) and are documented.

## Notes

Blocks the tracing and SLO work. Cardinality is the failure mode to watch: every label added multiplies series count, so justify each one.

dslegacy:
## Summary

Prometheus metrics for the Go API: request counter/latency histogram/in-flight gauge, pgxpool and Redis stats, outbound HTTP by upstream. Exposed on a separate internal listener (`METRICS_ADDR`, default loopback), never on the public router. Docker `observability` profile brings up Prometheus + Grafana scraping it.

**Note:** the issue says Chi; the API routes with `http.ServeMux`. Patterns are resolved via `mux.Handler(r)`, which is the only place they're available from outside the mux.

## Security
- `/metrics` not registered on the public mux; a live request to the public port returns 401 with no exposition data
- Config rejects a `METRICS_ADDR` equal to the public address
- Only method + status are read; headers, bodies, URLs, and error text never reach a label

## Cardinality
Every label is bounded: route (mux pattern or `other`), method (allowlist), status class, typed upstream constant, closed error-kind set, Redis command allowlist. No IDs, paths, hashes, keys, or query params. Asserted in tests.

## Verification
34 metrics tests + 5 config tests. Pool-exhaustion and Redis tests verified against real Postgres 16 + Redis 7. End-to-end: live API scraped by Prometheus v3.1.0, `health=up`, route label `GET /api/v1/vaults/{id}` with the UUID absent from the whole exposition. Middleware overhead ~1.4µs/request. CI green.

Second commit bumps `go.mod` to 1.25.13 for six stdlib advisories failing govulncheck on `dev` too.

Closes #1043


0xDeon (approved):
The cardinality guard is the part I looked hardest at and it holds up. Unmatched routes collapse to a fixed "other" label, and the pattern test asserts both that the UUID request produces /vaults/{id} and that the raw UUID appears in no label value anywhere in the registry. Exposition validity goes through Prometheus' own parser rather than string matching.

Also confirmed the two things I expected to be gaps and were not: a real go-redis hook records command duration and errors, and redis.Nil is correctly excluded from the error count.

Three things worth a follow-up, none blocking:

UpstreamCoinGecko is dead code. There is no CoinGecko client in the repo; the only coingecko references are a DeFiLlama coin key. The switch arm is unreachable.

ContractInvoker.SetHTTPClient is added but never called, so the Soroban write path and the 6s event indexer produce no series at all. The indexer is the busiest Soroban caller in the process.

REDIS_ADDR is never set in CI despite the test comment saying it is, so TestRedisIntegration skips on every run. One line in the workflow fixes it and also repairs the same latent issue in the pre-existing ratelimit test. Best value per effort of the three.

PR #1066, opened 2026-08-23
fix(ci): repair failing gates, dev image, and the never-running database tests

dslegacy:
## Summary

Fixes every failing CI gate, the broken dev image, and the DB integration tests that had never actually run in CI. Eight commits, each independently revertible.

**CI**
- govulncheck: six stdlib advisories fixed in go1.25.13 → bump `go.mod`
- The DB integration step set the wrong env var (`DATABASE_URL`; tests gate on `TEST_DATABASE_DSN`) and was `continue-on-error`, so it gated nothing. Now runs with `-p 1` and gates.

**Production bugs** (hidden until those tests ran)
- `GET /api/v1/activity` failed on every call — CTE didn't project `user_id`
- Audit tamper-evidence could never verify — hashed a ns timestamp, stored µs
- Every audit redaction deadlocked — `RedactEntry` re-took its own mutex

**Tests**
- 13 helpers each hand-curated a migration subset; all had rotted. Replaced with one `testutil.ApplyAllMigrations` applying the full chain in numeric order (all 100 verified to replay cleanly). Fixed the seeds, envelope decoding, weight sums, timestamp precision, and ordering assumptions that surfaced.

**Dev env**
- `Dockerfile.dev`: base image < go.mod; `air@latest` now needs Go >= 1.26 → pin both
- `docker compose` failed to parse on a fresh clone → `env_file` optional

## Notes
- Audit entries already in prod keep ns-derived hashes and will still fail `VerifyChain`; re-hashing would defeat the chain. Maintainer decision.
- `contract-audit.yml` untouched — its old failures predate #1024, which fixed it.
- First commit is identical to #1065's second; they rebase cleanly either way.

## Verification
Full suite against real Postgres 16 + Redis 7 locally; CI 21/21 including the DB step under `-race`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -364,9 +364,20 @@ jobs:
       - name: Test (unit)
         run: go test -race -timeout 120s -count=1 -short ./...
 
-      - name: Test (integration)
-        run: go test -race -timeout 120s -count=1 -run Integration ./...
-        continue-on-error: true
+      # Database-backed tests gate on TEST_DATABASE_DSN (the convention in
+      # internal/db and internal/repository/postgres), not DATABASE_URL, so
+      # without this variable every one of them silently skipped — and the
+      # step was continue-on-error, so it never gated a merge either. Scoped
+      # to this step so the unit run above stays DB-free and parallel.
+      #
+      # -p 1: the tests share one database and each wipes the public schema
+      # before applying the full migration chain, so packages must not run
+      # concurrently. No -run filter: DB-gated tests whose names lack
+      # "Integration" (internal/audit, internal/stellar backfill) must run too.
+      - name: Test (database integration)
+        env:
+          TEST_DATABASE_DSN: postgres://nester:nester@localhost:5432/nester_test?sslmode=disable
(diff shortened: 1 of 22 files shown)


0xDeon (approved):
Verified the central question here: no gate was weakened. The govulncheck ACCEPTED list is byte-identical to base, the CVE gate is fixed at root with the go 1.25.13 bump, and a continue-on-error was removed rather than added.

The DB test fix is the real win. Confirmed the DSN matches the CI postgres service and that the tests genuinely run now, repository/postgres takes 27.3s against roughly 1s when it was skipping. Running them surfaced three actual production bugs: the activity CTE missing user_id, the audit nanosecond/microsecond hash mismatch, and the self-deadlocking mutex in RedactEntry.

Checked every one of the 441 deleted lines. No test function and no assertion is removed, it is rotted per-file migration lists replaced by the shared helper. Two assertions were tightened.

One thing to pick up separately: apps/api/Dockerfile still pins golang:1.24-alpine against go.mod 1.25.13, same bug class this PR fixes in Dockerfile.dev. No CI job builds it so it stays invisible.

PR #1067, opened 2026-08-23
feat(observability): distributed tracing across the Go API, intelligence service, and Soroban invoker

### LINKED ISSUE #1054: feat(observability): distributed tracing across the Go API, intelligence service, and Soroban invoker
## Problem

A single user request can cross the Go API, the Python intelligence service, Anthropic's API, PostgreSQL, Redis, and Soroban RPC. Today the only thread connecting them is a request id: `RequestIDMiddleware` in `apps/api/internal/middleware/request_id.go` generates or accepts `X-Request-ID` and binds it to a context variable, and the Python service has an equivalent.

That is enough to grep two log streams. It is not enough to answer the question that actually gets asked in an incident: *where did the 8 seconds go?* Answering that today means exporting logs from both services, sorting by timestamp, and reconstructing causality by eye — which is slow, error-prone, and impossible for anything that happened last week.

There is also no visibility into the most expensive and least predictable hop. Anthropic streaming calls dominate intelligence latency and nothing records their duration, token counts, or model.

## Approach

OpenTelemetry end to end, with trace context propagated across the process boundary so the Python service continues the same trace rather than starting a new one. Traces answer "where did the time go for this request"; the metrics from #1043 answer "how is the system behaving overall". Both are needed and they are not substitutes.

## Scope

**Go API instrumentation**
- HTTP server and client spans via `otelhttp`, using Chi route patterns as span names, not raw paths.
- `pgx` instrumentation with the SQL statement as an attribute — parameterised text only, never interpolated values, since parameters carry user data.
- Redis command spans.
- Soroban chain invoker spans carrying ledger sequence, contract id, and function name. This is where unexplained latency usually lives.

**Cross-service propagation**
- W3C `traceparent` and `tracestate` on the intelligence relay call.
- FastAPI extracts and continues the trace. Verify the parent-child relationship in a real trace rather than assuming the SDK did it — a broken propagation looks identical to a working one until you open the waterfall and see two disconnected traces.
- Preserve the existing `X-Request-ID` behaviour; the two identifiers coexist, and the request id remains what support asks users for.

**Anthropic call instrumentation**
- Span per model call with attributes: model id, input and output token counts, stop reason, whether streaming, and time to first token. Time to first token is the number that determines perceived responsiveness and is currently invisible.
- Record tool-use rounds as child spans, since `max_tool_rounds` bounds a loop whose cost is otherwise opaque.

**Sampling**
- Head-based sampling at a low rate for normal traffic.
- **Always sample errors and requests exceeding a latency threshold.** The interesting traces are rare by definition; uniform sampling reliably discards exactly the ones needed. Use tail sampling in the collector if head-based cannot express this.

**Exemplars**
- Link the #1043 latency histograms to trace ids so a spike on a dashboard leads directly to a representative trace.

**Local usability**
- A docker-compose profile with an OTel collector and Jaeger or Tempo. Tracing that only exists in production will not be maintained, because nobody can iterate on it.

**Conventions**
- Document span naming and required attributes. Follow OTel semantic conventions where they exist rather than inventing names.

## Acceptance criteria

- [ ] A deposit request produces one trace spanning Go, Python, Anthropic, PostgreSQL, Redis, and Soroban RPC, with correct parent-child structure.
- [ ] Trace context survives the Go-to-Python boundary; asserted by a test, not by inspection.
- [ ] Anthropic spans carry token counts and time to first token.
- [ ] Errors are always sampled regardless of base rate.
- [ ] No span attribute contains secrets, wallet keys, JWTs, or interpolated SQL parameters — asserted by test.
- [ ] `docker compose --profile observability up` yields a working local trace view.
- [ ] Runbook includes one annotated real trace of a full deposit flow.

## Notes

Do #1043 first; exemplars depend on it. Watch overhead — instrument the hot path and measure the cost rather than instrumenting everything.

Depo-dev:
## Summary

Adds OpenTelemetry distributed tracing end to end — Go API through PostgreSQL, Redis, Soroban RPC, and across the process boundary into the Python intelligence service and its Anthropic calls.

`X-Request-ID` is untouched and still the support-facing correlation ID. Each server span records it as `nester.request_id`, so an operator holding one identifier can find the other.

Tracing is **opt-in and off by default**. With `TRACING_ENABLED` unset the API installs a no-op provider, dials no collector, and instrumentation costs nothing. A missing collector never fails a request.

Closes #1054

---

## Design decisions

**Sampling is split across two tiers.** A head sampler decides at span *start*; error status and latency are only known at span *end*, so the head cannot condition on either. The head applies a `ParentBased` ratio (honouring an upstream decision so a distributed trace is never half-recorded); the always-keep guarantees for errors and slow requests live in the collector's `tail_sampling` processor.

> **Consequence:** a trace the head drops never reaches the collector. **Set `TRACING_SAMPLE_RATIO=1.0` wherever tail sampling is deployed.** The `0.05` default assumes direct-to-backend export.

**Two deliberate deviations from the issue text:**

1. The issue says **Chi route patterns**. The API is served by `http.ServeMux` — Chi is in `go.mod` but routes nothing. Span names use ServeMux patterns (`GET /api/v1/users/{id}/savings-goals`), meeting the low-cardinality requirement with the real router.
2. The issue asks for **ledger sequence** on Soroban spans. These RPC responses don't return one (`getTxResult` carries only `Status`). Fabricating it, or issuing an extra RPC call for telemetry, would both be worse.

**One correction mid-implementation:** the DB instrumentation first wrapped `internal/db.Open` with `otelsql`. That function is used only by its own test — the app opens its pool via `repository.NewPostgresDB` (pgxpool). It would have produced **zero DB spans in production while its tests passed**. Replaced with `otelpgx` on the real path.

---

## Security and privacy

Telemetry is treated as an untrusted egress channel. Three layers: allow-listing at each call site; `internal/telemetry/redact.go` as a single choke point (rejects sensitive keys, strips Stellar seeds / JWTs / bearer headers / API keys / base64 XDR, length-bounds and UTF-8-coerces everything); and tests asserting forbidden values are absent from exported spans, with the redactor's invariants fuzzed.

**Two library defaults were disabled because they leak:**

- `otelpgx`'s `IncludeQueryParameters` is **not** set — bound arguments are account numbers, balances, addresses. Parameterised SQL *is* recorded so slow queries stay identifiable.
- The Python SDK's `record_exception` / `set_status_on_exception` are **disabled** on model and tool spans. Both write the raw exception message; an Anthropic `401` interpolates the API key and a validation error interpolates the user's prompt. Only the exception *type* is recorded.

Redis records command names only. Soroban spans carry contract ID, function, RPC method, tx hash and status — never the operator key, XDR, or simulation payloads.

### Bugs the tests caught

| Defect | Consequence |
|---|---|
| `\b`-anchored redaction patterns | Secrets glued to adjacent text (`secret=SB…`, DSNs) passed through **whole** |
| Truncation sliced mid-secret | Exported a usable 36-char prefix of a seed |
| Exempt-suffix ordering | `operator_secret.token_count` reported **safe** |
| Invalid UTF-8 passthrough | Corrupts OTLP export — found by the fuzzer in under a second |
| `r.Pattern` read from the wrong request | Every span named just `GET` |
| `http.Hijacker` hidden by the wrapper | **`/ws` upgrades would have broken in production while every unit test passed** |
| Base64 XDR unredacted | Soroban RPC echoes XDR in errors; a signed envelope embeds the operator signature and every operation argument |

---

## Testing

All green locally and on CI (**18 checks passed, 0 failed**).

`go build` · `go vet` · full `go test -short` · `ruff` · `mypy --strict` (70 files) · **357 pytest** · gitleaks (full history, no leaks) · `otelcol validate` · compose profile gating.

**End-to-end verified, not assumed.** A trace was exported through the real path (Go SDK → OTLP gRPC → collector tail sampling → Jaeger) and queried back from Jaeger's API:

```
TRACE 38959b65… — 3 spans
  GET /api/v1/transactions   13658us  ROOT   nester.request_id=req-e2e-verify-001  nester.force_keep=True
  ├─ postgres.query           5469us
  └─ soroban.invoke           8189us
```

**Propagation is asserted from both ends**, not eyeballed in a UI — a broken propagation looks identical to a working one until you open a waterfall. The Go test asserts the emitted `traceparent` matches the client span that produced it; the Python tests assert the server span inherits that trace ID with the correct remote parent.

---

## Local setup

```bash
TRACING_ENABLED=true INTELLIGENCE_TRACING_ENABLED=true \
  docker compose --profile observability up
```

Collector + Jaeger, both profile-gated so plain `docker compose up` is unchanged. UI at <http://localhost:16686>. Full reference: `docs/observability/tracing.md`.

> **Merge note:** PR #1065 adds `prometheus`/`grafana` under the same profile. No shared ports or volume names — the conflict is textual only, keep both blocks.

---

## The `go` directive bump

CI's vulnerability gate reported seven stdlib CVEs. **Six already fail on `dev`** (new CVEs published after the workflow's hardcoded `ACCEPTED` list was written — verified against a clean `dev` worktree). **One, `GO-2026-6091`, is introduced here**: the OTLP gRPC exporter pulls in gRPC → `golang.org/x/net/trace` → `html/template`. `dev` has zero packages importing `html/template`; this branch has one, and that edge makes the vulnerability reachable in otherwise-untouched code.

All seven are fixed in **go1.25.13**, so the `go` directive is bumped rather than the finding suppressed — which also clears the six pre-existing failures. `govulncheck` now reports *"No vulnerabilities found"*.

Rejected alternatives: a `grpcnotrace` build tag needs a new CI flag and fixes only this branch's finding; extending `ACCEPTED` is editing CI to hide real unpatched CVEs. This is slightly wider than #1054's scope — flagging it in case you'd rather split it out.

---

## Limitations

- **Exemplars are not included.** They attach to #1043's histograms (PR #1065), which is green but unmerged, so `dev` has no `internal/metrics` to hook. This branch stays on `dev` so both PRs remain independently mergeable. The exact follow-up — `ObserveWithExemplar` guarded by `IsSampled`, `EnableOpenMetrics`, `--enable-feature=exemplar-storage` — is in the docs.
- `-race` not run locally (needs cgo; no C compiler on this machine). CI's `API (Go)` job covers it.
- The runbook trace is from a local run; structure and attributes match production, all identifiers are dev values.

---

## Checklist

- [x] Go HTTP server + client instrumentation
- [x] PostgreSQL, Redis, Soroban instrumentation
- [x] Go → Python W3C propagation + FastAPI continuation
- [x] Anthropic spans, streaming TTFT, tool-round spans
- [x] Sampling, with errors and high-latency retained
- [!] Prometheus exemplars — sequenced behind PR #1065; follow-up documented
- [x] Local collector + Jaeger compose profile
- [x] Security/privacy, propagation, and end-to-end tests
- [x] Documentation + annotated runbook trace
- [x] All CI checks pass (18/18)

---

All work in this PR targets the dev branch. This PR must not be merged directly to main.

## Summary by CodeRabbit

* **New Features**
  * Added opt-in distributed tracing across API and intelligence services.
  * Trace requests, database activity, cache operations, external calls, blockchain interactions, model calls, and tools.
  * Added configurable sampling, latency tracking, error retention, and request correlation.
  * Added privacy protections to prevent credentials, payloads, prompts, and sensitive values from appearing in traces.
  * Added optional local observability support with Jaeger and an OpenTelemetry Collector.

* **Documentation**
  * Added setup, configuration, troubleshooting, and trace-investigation guidance.

* **Tests**
  * Added comprehensive tracing, propagation, sampling, lifecycle, and sensitive-data protection coverage.


0xDeon (approved):

PR #1068, opened 2026-08-23
feat(api): transactional outbox for webhook and notification delivery

### LINKED ISSUE #1049: feat(api): transactional outbox for webhook and notification delivery
## Problem

`apps/api/internal/domain/webhook/` and `apps/api/internal/notifications/` dispatch side effects, but writing a domain row and emitting the resulting notification or webhook are not atomic.

Two failure modes follow directly:

- **Lost side effect.** The domain write commits, the process crashes before dispatch, and the notification never happens. The user's balance changed and they were never told.
- **Duplicate side effect.** Dispatch succeeds, the acknowledgement is lost, a retry fires, and the webhook is delivered twice. A consumer treating a webhook as a trigger for a payout now has a double-payout.

Neither is theoretical for a system whose PRD already documents an indexer that doubled balances on restart (B-01).

## Scope

**Outbox table.** Side effects are inserted into an `outbox` table inside the same transaction as the domain change. If the domain write rolls back, so does the intent to notify. This is the whole point — atomicity comes from the shared transaction, not from careful ordering.

Columns: id, aggregate type and id, event type, payload, dedupe key, created timestamp, attempt count, status, next attempt time, last error.

**Relay worker.** Polls unsent rows and dispatches them. Reuse the existing `jobqueue` idempotency machinery rather than inventing a second retry mechanism — the queue already handles attempt counting and backoff, and two competing retry systems will eventually disagree.

**Delivery semantics.** At-least-once, with a dedupe key that consumers can use to discard repeats. Document explicitly that consumers must be idempotent, since at-least-once without that requirement stated is a trap. Include the dedupe key in the webhook payload and headers.

**Poison messages.** A payload that permanently fails — a consumer returning 400 forever, an unserialisable body — must not block the queue behind it. Move it to a dead-letter state after a bounded attempt count, alert, and continue. Ordering guarantees should be per-aggregate at most, not global, so one stuck aggregate cannot stall everything.

**Ordering.** Decide and document whether ordering is guaranteed per aggregate. If yes, the relay must not dispatch event N+1 for an aggregate until N succeeds, which interacts with the poison-message path and needs to be specified rather than emergent.

**Cleanup.** Delivered rows need a retention policy and a pruning job, or the table grows without bound.

## Acceptance criteria

- [ ] Killing the process between domain write and dispatch results in the side effect being delivered on restart, proven by test.
- [ ] Rolling back the domain transaction leaves no outbox row.
- [ ] A consumer that fails permanently does not block delivery for other aggregates.
- [ ] Duplicate delivery carries a stable dedupe key.
- [ ] Ordering guarantee is documented and asserted.
- [ ] Retention job exists and is covered.

## Notes

Depends on nothing, but overlaps #822's ledger work if that lands first — both introduce transactional guarantees around domain writes, so sequencing matters.

olathedev:
Closes #1049

## The problem

A domain write and the side effect it causes are two writes to two different systems. Sequencing them carefully does not make them atomic, and **both** orderings lose:

- **Write, then dispatch** — the process dies in between and the notification never happens. The user's balance changed and nobody told them.
- **Dispatch, then write** — the write fails and a notification went out for something that never happened.

This isn't hypothetical here. `EnrichProgress` marked a savings-goal milestone as notified and *then* spawned a goroutine to notify. A crash in between left the goal permanently flagged "already told them" for a notification that never happened — and because the flag said it was done, nothing would ever retry it.

## The mechanism

Side effects are inserted into a new `outbox` table **inside the domain write's own transaction**. Atomicity comes from that shared transaction and from nothing else: roll it back and the intent to notify rolls back with it; commit it and the intent is durable even if the process dies one instruction later.

```
┌─ one transaction ──────────────────────┐
│  UPDATE savings_goals SET …            │
│  INSERT INTO outbox (…) VALUES (…)     │
└────────────────────────────────────────┘
             │ commit
             ▼
   relay ──▶ jobs (the existing durable queue, #824)
             │
             ▼
   webhook delivery / notification dispatch
```

The relay does **not** implement retry, backoff, or dead-lettering for delivery. It hands each row to the job queue keyed on the row's dedupe key, and the queue owns everything after that — per the issue's "reuse the existing `jobqueue` idempotency machinery rather than inventing a second retry mechanism".

## Delivery semantics

At-least-once, with a dedupe key that is **derived, never generated**: `savings_goal:{id}:milestone:50` produces the same key in this process and in the one that picks the event up after a restart.

The key reaches the consumer both ways the issue asks for:

- **Webhooks** — `X-Nester-Dedupe-Key` header *and* a `dedupe_key` field in the body, so a consumer that queues the raw body and processes it later (headers long gone) can still dedupe. It is injected before signing, so signatures still verify against what was actually sent.
- **Notifications** — passed to the dispatcher's existing deduplicator, so a redelivered event is recorded as *suppressed* rather than shown to the user twice.

This also closes a hole `FireForUser`'s own comment documented: it generated a fresh delivery id per call, so a caller retry produced a delivery the recipient could not recognise as a repeat. Outbox-driven webhooks go through `FanOut`, which derives each delivery id from `dedupeKey + subscriptionID`.

`docs/webhooks.md` now states consumer idempotency as a **requirement**, not a caveat.

## Ordering

Guaranteed **per aggregate**, explicitly **not** global. The relay dispatches only each aggregate's oldest non-terminal event and holds the rest until it resolves.

`ClaimDue` selects `DISTINCT ON (aggregate_type, aggregate_id)` ordered by `(created_at, id)`. The due-ness filter is applied *after* that selection — deliberately. Filtering inside the `DISTINCT ON` would let a backing-off head be skipped in favour of the event behind it, silently breaking the ordering it exists to preserve. There's a test pinning exactly that.

## Poison messages

A permanently-failing event stalls its **own** aggregate, and only until it dead-letters:

1. The queue exhausts its attempt budget and dead-letters the job.
2. The relay sees the dead job and moves the outbox row to `dead`.
3. `dead` is terminal, so the aggregate's next event becomes claimable immediately.

Re-dispatching from the outbox at that point would be the second retry mechanism this design avoids, so it doesn't. Dead-lettering logs at `ERROR` with the aggregate, event type, dedupe key, and reason, and increments `outbox.dead_lettered` — the counter worth alerting on, since a non-zero value means a side effect will never happen.

## Retention

A leader-gated sweep deletes terminal rows past their window. Delivered rows go after 7 days; dead rows survive 30, because they're the evidence someone needs to work out which side effect never happened.

`pending` and `dispatching` rows are **never** pruned at any age — deleting one is silently dropping the side effect the outbox exists to guarantee. There's a test for that specifically.

## Acceptance criteria

| Criterion | Where |
|---|---|
| Crash between write and dispatch → delivered on restart | `TestOutboxSurvivesCrashBetweenWriteAndDispatch`; `TestSavingsGoalMilestoneIntegration` asserts the crash window, then drains a *freshly constructed* relay |
| Rollback leaves no outbox row | `TestOutboxRollbackLeavesNoRow` (real transaction, real Postgres) |
| Permanent failure doesn't block other aggregates | `TestPoisonEventDoesNotBlockOtherAggregates`, `TestPoisonEventUnblocksItsOwnAggregate`, `TestMarkDeadUnblocksAggregate` |
| Duplicate delivery carries a stable dedupe key | `TestFanOut_DeliveryIDsAreDerivedFromTheDedupeKey`, `TestDeriveIDIsStableAcrossProcesses` (golden-pinned — drift invalidates every consumer's dedupe store), `TestDelivery_CarriesDedupeKeyInHeaderAndBody` |
| Ordering guarantee documented and asserted | `docs/outbox.md`, `docs/webhooks.md`; `TestPerAggregateOrderingIsPreserved`, `TestClaimDueReturnsOneHeadPerAggregate`, `TestClaimDueDoesNotSkipABackingOffHead` |
| Retention job exists and is covered | `outbox.RetentionJob`, `retention_test.go`, `TestPruneTerminalRemovesOnlyTerminalRows` |

## Notes for review

**A stuck-row case found while writing the ordering docs.** The claim originally filtered out rows at `max_attempts`. A relay dying mid-hand-off on its final attempt therefore left a row claimed with no job carrying it and no way to ever reclaim it — blocking its aggregate permanently. Such a row is now claimed and dead-lettered, which is what unblocks the aggregate. Asserted against both the in-memory repository and Postgres.

**The relay runs on every instance**, not behind leader election. `ClaimDue` takes row locks with `SKIP LOCKED`, so instances divide the backlog rather than duplicating it; electing a leader here would turn a horizontal scale-out into a single point of failure for every side effect in the system. The retention sweep *is* leader-gated, since pruning from every instance is correct but wasteful.

**`MilestoneOutboxRecorder` is an optional interface**, not a method on `savingsgoal.Repository`. The guarantee it offers is inherently database-backed — an in-memory repository cannot provide it, and adding a method every implementation must stub would hide exactly the distinction that matters. Repositories that can't transact keep the old goroutine path, documented as lossy.

**The two savings-goal integration tests changed shape.** They now assert the crash window directly: after the domain write the milestone is recorded and *nothing* is delivered, and only then does a fresh relay drain the durable rows. That's a better test than the sleep-and-hope they had.

**`SendGoalMilestone` gained a `dedupeKey` parameter.** Driving the chain from an at-least-once queue means every member has to be idempotent, and a narrower optional interface would have been interface gymnastics for the same result.

## Verification

Ran against a real PostgreSQL instance locally, not just the mocks:

- `go test -p 1 ./...` with `TEST_DATABASE_DSN` set — full suite green, including every new integration test
- `go test -race -short ./...` — green
- `go vet ./...`, `gofmt` — clean on every touched file
- Migration prefix collision check (the CI gate) — clean

Commit `47fe7cc` reverts unrelated `gofmt` churn: a `gofmt -w internal/service/` run mid-change reformatted 24 files this PR doesn't touch, because Go 1.26's gofmt disagrees with the formatting the repo was written under. The diff is scoped to the 34 files this change actually edits.


0xDeon (approved):

Depo-dev:
@olathedev Your PR has merge conflicts please fix them so i can review

0xDeon (changes requested):
This one is close — the review was already approving and CI was green. It has just gone stale against `dev` and no longer merges.

Two files conflict:

- `apps/api/internal/service/savings_goal_service.go`
- `apps/api/internal/service/savings_goal_milestone_integration_test.go`

Both are savings-goal files that moved on `dev` after this branch was cut, so the outbox work itself is not in question — it is purely the rebase.

One thing worth flagging before you do it: a recent merge in this area auto-merged without reporting a conflict and still produced broken code. `dev` reordered the savings-goal column list, and git combined it with a branch change to produce duplicated columns in five SELECT statements and a duplicated argument list in the scanner. It compiled fine and failed at runtime with `sql: expected 30 destination arguments in Scan, not 31`. So please do not trust a clean `git merge` here — after rebasing, read the resolved SELECT lists and their scanner side by side and confirm the column counts match.

Rebase onto current `dev`, push, and once CI comes back green I will merge it.

Note the Netlify checks are failing across every PR built since 29 Aug, including work already merged to `dev`. That is a platform-side issue, not yours — ignore it. What I need green is the GitHub Actions set, in particular `API (Go)`.

PR #1072, opened 2026-08-23
feat(platform): service level objectives, error budgets, and burn-rate alerting

### LINKED ISSUE #1056: feat(platform): service level objectives, error budgets, and burn-rate alerting
## Problem

Metrics (#1043) and traces (#1054) produce telemetry. Neither defines what *good* means.

Right now there are no SLOs, no alerting rules, and no on-call runbook. The failure detection mechanism is a user noticing and telling someone. That works while the user count is small and stops working exactly when it matters.

The second-order problem is subtler: without an error budget there is no principled way to decide between shipping features and fixing reliability. That decision currently gets made by whoever argues most persuasively in the moment, which produces inconsistent outcomes and resentment.

## Approach

Define SLIs for user-visible flows, set SLO targets with justification, alert on **burn rate** rather than on instantaneous threshold breaches, and write down what happens when the budget is exhausted.

The burn-rate detail matters. Alerting on "error rate above 1%" pages at 3am for a 30-second blip and stays silent through a week of 0.9% degradation that consumes the entire quarter's budget. Multi-window multi-burn-rate alerting fixes both directions.

## Scope

**Service level indicators**

For each, define numerator, denominator, and exactly which events are excluded and why. Vague SLIs produce arguments during incidents.

- *Deposit success rate* — successful deposits over attempted. Exclude user-cancelled wallet signatures; those are not failures.
- *Deposit latency* — request to confirmed on-chain, measured at a stated percentile.
- *Withdrawal success rate and latency* — same treatment.
- *Balance freshness* — indexer lag, ledger tip minus last indexed ledger. Directly ties to PRD B-01; a stalled indexer currently produces no signal at all.
- *Intelligence availability and latency* — including time to first token, since perceived responsiveness is dominated by it.
- *Intelligence refusal rate* — a guardrail misfire wave looks like nothing else in the metrics but is very visible to users.
- *API availability* — non-5xx over total, per route class.

**Targets**

Justify each number. "99.9% because it's a round number" is not a justification; "99.9% allows 43 minutes monthly, which matches our current single-region deployment and on-call response time" is. State the measurement window (28-day rolling is a reasonable default) and where the data comes from.

**Multi-window multi-burn-rate alerting**

- Fast burn: 2% of the 28-day budget consumed in 1 hour, confirmed by a 5-minute window, pages.
- Slow burn: 5% consumed in 6 hours, confirmed by a 30-minute window, opens a ticket.

The short confirmation window suppresses single-scrape spikes. Document the arithmetic so the thresholds can be tuned rather than cargo-culted.

**Error budget policy**

A written engineering agreement, agreed before it is needed: what happens when the budget is exhausted. Something like — feature work pauses, reliability work takes priority until the budget recovers, with a named person who can override and a record of overrides. The value is in agreeing while nobody is under pressure.

**Runbooks**

One per SLO. First response steps, the dashboard to open, the three most likely causes with how to distinguish them, and the escalation path. A runbook that says "investigate the issue" is not a runbook.

**Synthetic probes**

Continuously exercise deposit, withdrawal, balance read, and an intelligence query against staging. Synthetic probes catch flows that are broken but generating no traffic — the failure mode where nobody notices because nobody tried.

**Dashboards**

One per SLO showing current attainment, budget remaining, and burn rate. Plus a documented review cadence — an SLO nobody reviews is decoration, and reviewing them is what surfaces targets that were set wrong.

## Acceptance criteria

- [ ] Every SLI has a written definition including exclusions.
- [ ] Every SLO target has a justification, not just a number.
- [ ] Multi-window burn-rate alerts implemented, with arithmetic documented.
- [ ] A deliberately induced error burst triggers the fast-burn alert; a slow degradation triggers the slow-burn alert and not the pager. Both verified.
- [ ] Error budget policy written and agreed.
- [ ] One runbook per SLO with concrete first steps.
- [ ] Synthetic probes running against staging, with their own alerting.
- [ ] Review cadence documented and first review held.

## Notes

Depends on #1043 and benefits from #1054. Do not start before metrics exist — SLOs over telemetry you do not have are fiction.

Depo-dev:
Closes #1056.

## Summary

- Defines production SLIs and SLOs for the user-visible flows
- Adds rolling 28-day error-budget calculations
- Adds multi-window multi-burn-rate alerting
- Adds SLO dashboards
- Adds staging synthetic probes
- Adds operational runbooks
- Documents the error-budget policy and review cadence

Builds on the metrics from #1043 (PR #1065) and the tracing from #1054 (PR #1067).

## The telemetry gap this had to close first

#1043 instrumented the **infrastructure** — HTTP, Redis, pgxpool, outbound. Four of the nine SLIs this issue names were measurable from it. Five were not, because the signals existed in code but were never exported:

| Signal | State before |
|---|---|
| Deposit/withdrawal outcomes | No counter, and no way to tell a user cancellation from a chain failure |
| Balance freshness | `GetIndexerLag` was defined and never called |
| Time to first token | A sampled span attribute only |
| Refusal rate | `is_refusal()` detected it, nothing counted it |

Per the issue's instruction to add only the minimal telemetry required, this adds emission for those five. The logic already existed in every case.

## SLI/SLO model

Full definitions in [`docs/observability/slo.md`](docs/observability/slo.md). Every SLI states numerator, denominator, source, exclusions, and the reason for each exclusion.

| SLI | Target | Budget | ~Time / 28d | Severity |
|---|---|---|---|---|
| API availability | 99.9% | 0.1% | 40 min | page |
| Deposit success | 99.5% | 0.5% | 3h 22m | page |
| Withdrawal success | 99.5% | 0.5% | 3h 22m | page |
| Deposit/withdrawal latency | 99% under 30s | 1% | 6h 43m | ticket |
| Balance freshness | 60 ledgers or fewer | n/a (gauge) | n/a | page |
| Intelligence availability | 99% | 1% | 6h 43m | ticket |
| Intelligence TTFT | 95% under 3s | 5% | 33h 36m | ticket |
| Intelligence refusal | under 30% over 6h | n/a (guardrail) | n/a | ticket |

Each target documents why a stricter value is currently inappropriate and why a weaker one is unacceptable. Rolling 28-day window — a whole number of weeks, so attainment does not shift with which weekday it is read.

### Three choices argued rather than asserted

These are the ones most likely to be relitigated during an incident, so the reasoning is written down:

1. **Latency SLIs count slow events, not percentiles.** A percentile cannot be converted into an error budget — it is not a count of bad events, so there is nothing to burn. Counting requests slower than a threshold gives a bad/eligible ratio with the same shape as every other SLI. A p95 is still recorded for dashboards.

2. **`failed_chain` is not excluded**, despite Soroban RPC being outside our control. A user whose deposit did not land cannot tell "the network was down" from "we could not reach the network", and an SLI that excused unowned infrastructure would report perfect health through a total chain outage.

3. **Refusals count as available.** The service worked as designed. Refusal rate is tracked separately as a product-quality guardrail with **no error budget** — some refusals are correct behaviour, and there is no principled budget of correct behaviour to burn.

## Burn-rate arithmetic

```
error_budget = 1 - SLO_target
burn_rate    = observed_error_ratio / error_budget
```

A 28-day window is 672 hours.

| | Derivation | Rate | Budget gone in |
|---|---|---|---|
| **Fast burn** | 2% of budget in 1h, so `0.02 x 672/1` | **13.44x** | ~50 hours |
| **Slow burn** | 5% of budget in 6h, so `0.05 x 672/6` | **5.6x** | 5 days |

The multipliers are SLO-independent; what changes per SLO is the absolute ratio each represents:

| SLO | Budget | Fast (13.44x) | Slow (5.6x) |
|---|---|---|---|
| 99.9% | 0.1% | 1.344% | 0.560% |
| 99.5% | 0.5% | 6.720% | 2.800% |
| 99% | 1.0% | 13.440% | 5.600% |
| 95% | 5.0% | 67.200% | 28.000% |

**Why two windows per alert:** the long window defines the policy but has a tail problem — a 5-minute outage keeps a 1h window elevated for a full hour after recovery. The short window is a *recovery detector*, so the alert clears within minutes of the incident ending. Verified by the test `API availability fast burn resolves once the burst stops`.

## Alert severities

`page` is reserved for burn rates that exhaust the budget within ~2 days **on user-visible money paths**. Everything else tickets. Intelligence never pages — the feature is advisory, and there is no 3am action that fixes a third-party provider.

20 alerts, each carrying summary, description, current burn, remaining budget, dashboard, and runbook. Alertmanager groups by `(slo, burn, service)`; a fast burn inhibits the slow burn on the same SLO, and `SLOTargetDown` inhibits the derived alerts that go stale because of it.

## Error-budget policy

[`docs/observability/error-budget-policy.md`](docs/observability/error-budget-policy.md) — four states by budget remaining:

| State | Remaining | Effect |
|---|---|---|
| Healthy | above 50% | Normal development |
| Watch | 25-50% | Named in review, issue gets an owner |
| Warning | 0-25% | Reliability work prioritised for the affected service |
| Exhausted | 0% or below | Feature work on that service gated |

Defines who can override (a maintainer, not the sole author of the work), what qualifies, how overrides are recorded (`error-budget-override` label), and who reviews them. Three overrides on one SLO in two cycles is itself a finding.

**Marked as proposed, not in force.** The maintainers have not adopted it, and a policy nobody agreed to cannot gate anyone's work.

## Testing evidence

Every command below was actually run.

| Gate | Result |
|---|---|
| `go build ./...` | PASS |
| `go vet ./...` | PASS |
| `go test -short ./...` | PASS |
| `gosec -severity medium ./...` | PASS (no findings) |
| `ruff check .` | PASS |
| `mypy app` (strict) | PASS, 72 files |
| `pytest` (intelligence) | 359 pass, 2 pre-existing failures (see note) |
| `pytest` (probes) | 29 pass |
| `promtool check config` | PASS — 95 rules across 3 files |
| `promtool test rules` | PASS — both suites, stable over repeated runs |
| `amtool check-config` | PASS |
| Dashboard drift check | PASS |
| `docker compose config` | PASS |
| `gitleaks detect` | no leaks |

Note: the 2 failures are in `test_guardrails.py` — a Windows `cp1252` decode bug, reproduced on clean `dev` by stashing this branch. Fixed upstream by #1061; CI runs Linux/UTF-8.

`-race` was not run locally (no cgo toolchain on the dev machine); CI runs it.

### A bug the tests caught

`sum(all) - sum(4xx)` evaluates to an **empty vector** when a service has served no 4xx at all. That deleted the API availability SLI entirely — the alerts could never fire, and would have stayed silent through a later 5xx outage. Fixed with `or vector(0)`, covered by two regression tests.

This is the exact class of failure the PR claims to prevent, and it was found by testing rather than by reasoning. Noted in the docs for that reason.

### Alert tests

`promtool` unit tests run against the **real** rule files. They prove:

- an induced error burst fires the fast-burn alert at the right severity
- sustained degradation tickets **without** paging
- a healthy system fires nothing
- thresholds behave below / exactly at / above the boundary
- **a 4xx flood cannot manufacture an availability breach** (10x the successful traffic, zero alerts)
- **user cancellations cannot burn the deposit budget** (10x the successful traffic, zero alerts)
- **a withdrawal outage is not hidden by healthy deposit volume**
- a frozen lag gauge is caught by staleness, not by the lag value
- absent metrics alert rather than reading as health

## Dashboards, runbooks, probes

**5 dashboards** (45 panels), generated by [`scripts/build_slo_dashboards.py`](scripts/build_slo_dashboards.py) and committed — CI fails if they drift. Every panel reads the recorded series the alerts fire on, so a dashboard cannot disagree with the pager mid-incident.

**9 runbooks**, one per SLO plus monitoring-down. Each has what the alert means, three concrete first actions *with the queries to run*, dashboards/logs/traces, three likely causes with the evidence that distinguishes them, mitigation, escalation, recovery verification, and postmortem requirements.

**4 synthetic probes** (deposit, withdrawal, balance, intelligence) against staging every 15 minutes. Three independent guards keep them off production:

1. `PROBE_API_BASE_URL` has no default — unset aborts, no fallback to localhost or a deployed host
2. Environment name **and** target URL are checked separately against production markers
3. Deposit/withdrawal require `PROBE_ALLOW_MUTATIONS=true` — a default run is read-only

A misconfigured probe exits `2`; a failed probe exits `1`. A bad config is never mistaken for an outage.

## Security and cardinality

Never a label: user ID, wallet address, transaction hash, vault ID, amount, prompt text, model answer, raw error string.

Every label added is a closed constant set fixed at compile time — `flow` (2), `outcome` (5 / 4), `reason` (2), `probe` (4), probe `reason` (8). Total added series is a few hundred, fixed, and cannot be moved by traffic or by a caller. A test asserts the flow label space is exactly 2 x 5.

Monitoring failures cannot break the app: all recorders are nil-safe, the metrics listener failing is logged and non-fatal, and the Prometheus handler uses `ContinueOnError` so a broken collector degrades the scrape rather than blinding every other metric.

Missing telemetry fails **visibly** — `IndexerMetricsAbsent`, `SLOMetricsAbsent`, `SLOTargetDown`, and `SyntheticProbeMetricsAbsent` exist so absent data alerts rather than reading as health.

## Also fixed

`docker-compose.yml` did not parse on `dev`. The #1065/#1067 merge left `otel-collector.depends_on.jaeger` with an empty body, breaking **every** `docker compose` command rather than only the observability profile. Restored the `condition: service_healthy` that the jaeger healthcheck was written for.

## Open items — deliberately not claimed as done

- **The first SLO review has not been held.** Cadence and agenda are defined; the review is scheduled. No meeting is reported as having occurred.
- **Targets are set from reasoning, not observed history** — five of these SLIs did not exist before this PR. The first review after 28 days of data is when they can be validated, and some will move.
- **The refusal threshold (30%) is uncalibrated.** An estimate, not a measurement. First agenda item of the first review.
- **Probes need staging secrets before they run.** The workflow skips cleanly until then; coverage is zero and `SyntheticProbeMetricsAbsent` will say so.
- **Alertmanager receivers are empty.** Routing is defined and testable; connecting a destination is a deployment task, and a webhook URL in a committed file would be a credential leak.

## Scope

47 files, +11,869 / -12 — almost purely additive.

Production code changes are confined to metric emission. No existing metrics, tracing, logging, auth, API, or deployment behaviour is altered. Two Go files touched only for the SLI edits were left on `dev`'s existing formatting rather than reformatted, to keep the diff reviewable — whole-repo `gofmt` belongs in its own change.


0xDeon (approved):

PR #1073, opened 2026-08-24
feat(dapp): add error boundaries and skeleton loading

### LINKED ISSUE #1046: feat(dapp): global error boundary, per-route fallbacks, and skeleton loading states

Damola09:
Closes #1046

Uses the App Router `error.tsx` / `global-error.tsx` / `loading.tsx` conventions rather than a hand-rolled boundary. Per-route fallbacks render inside `AppShell`, so a route failure never blanks the navigation, and recovery goes through `reset()` — nothing calls `window.location.reload()`.

Caught errors reach a structured sink with route, boundary, category, timestamp and environment. Payloads are scrubbed, not trusted: Stellar keys/contract ids/seeds/muxed accounts, tx hashes, bearer tokens and JWTs, and currency amounts are redacted; sensitive context keys are dropped; dynamic path segments collapse to `[id]`; stack traces are never sent.

The existing skeleton components were dead code — they are now wired up, plus new ones for views that had none. Loading and empty are separated where they were conflated (portfolio showed "No positions yet" while the request was still in flight; analytics collapsed loading/error/empty into three bare strings).

### Testing

```
npx vitest run          # 149 passed, 25 files (23 new)
npx playwright test     # 13 new tests passed
npx next build          # succeeds, 22 routes
npx tsc --noEmit        # 2 errors, both pre-existing on dev
npx eslint app components lib   # 14 errors, all pre-existing on dev
```

### Privacy

Wallet addresses, balances, transaction amounts, auth tokens and cookies are excluded from error payloads. Asserted in both the unit suite and Playwright.

### Note

`tests/accessibility.spec.ts` is flaky on `dev` today: exactly one of its four cases fails per run with `color-contrast`, on a different page each time. Reproduced on a clean `origin/dev` worktree, so it is not from this branch. One real `aria-prohibited-attr` violation (the savings APY skeleton) is fixed here.

### DIFF
--- a/apps/dapp/frontend/__tests__/error-reporting.test.ts
+++ b/apps/dapp/frontend/__tests__/error-reporting.test.ts
@@ -1,0 +1,214 @@
+import { describe, it, expect, beforeEach, vi, afterEach } from "vitest";
+import {
+  buildErrorEvent,
+  categorizeError,
+  reportError,
+  resetErrorReportDedupe,
+} from "@/lib/observability/report-error";
+import {
+  sanitizeText,
+  sanitizeRoute,
+  sanitizeContext,
+  REDACTED,
+} from "@/lib/observability/sanitize";
+
+const WALLET = "GABCDEFGHIJKLMNOPQRSTUVWXYZ234567ABCDEFGHIJKLMNOPQRSTUVW";
+const CONTRACT = "CABCDEFGHIJKLMNOPQRSTUVWXYZ234567ABCDEFGHIJKLMNOPQRSTUVW";
+const SECRET = "SABCDEFGHIJKLMNOPQRSTUVWXYZ234567ABCDEFGHIJKLMNOPQRSTUVW";
+const TX_HASH = "a".repeat(64);
+/**
+ * Assembled at runtime rather than written as a literal: a JWT-shaped string in
+ * source trips secret scanners even when, as here, it is synthetic test data.
+ */
+const JWT = [
+  Buffer.from(JSON.stringify({ alg: "HS256" })).toString("base64url"),
+  Buffer.from(JSON.stringify({ sub: "1234567890" })).toString("base64url"),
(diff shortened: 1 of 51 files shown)


0xDeon (approved):

PR #1074, opened 2026-08-24
feat(api): deterministic replay harness proving Stellar event indexer idempotency

### LINKED ISSUE #1051: feat(api): deterministic replay harness proving Stellar event indexer idempotency
## Problem

The PRD marks the Stellar event indexer **blocked, do not ship**, with three BLOCKING items:

- **B-01** — the cursor is held in memory and resets on restart, so events are reprocessed and balances double.
- **B-02** — `startLedger` is seeded at 0, which the Stellar RPC rejects, so the indexer never starts on a fresh boot.
- **B-03** — balance updates are not idempotent, so any duplicate delivery corrupts state.

Migration `028_create_processed_events_table` exists, so some of the machinery has landed. What does not exist is any way to *demonstrate* the indexer is correct. The failure mode here is silent balance corruption, which is the worst category of bug in a savings product: nothing errors, the number is simply wrong, and it stays wrong.

## Scope

**Recorded fixture.** Capture a real ledger range from testnet containing the event types the indexer handles — deposits, withdrawals, harvests — including at least one large-integer amount, since PRD B-11 records precision loss from a `float64` path in `extractEventAmount`. Check the fixture in so the test is hermetic and does not depend on network access or testnet state.

**Replay harness.** Feed the fixture through the indexer and assert final balances under four conditions, all of which must produce identical results:

1. Clean single pass.
2. Restart mid-stream — kill after N events, resume from persisted cursor, and confirm no double-application. This is B-01.
3. Duplicate delivery — deliver every event twice, in the same order. This is B-03.
4. Out-of-order delivery within a ledger — shuffle events inside a ledger boundary and confirm the final state is unchanged, or document explicitly that ordering is required and enforce it.

**Cursor persistence.** Assert the cursor is persisted transactionally with the balance update it corresponds to. If the cursor advances and the balance write fails, the event is lost forever; the two must commit together.

**Cold start.** Assert a fresh deployment seeds `startLedger` from the current tip rather than 0, and that this is covered rather than fixed and forgotten. This is B-02.

**Precision.** Assert that a large `i128` amount round-trips exactly. Any `float64` in this path should be removed as part of the work, per PRD B-11.

**Relocation.** PRD notes the indexer logic sits in `main.go` and should move to `internal/stellar/EventPoller.PollEvents`. The harness is far easier to write against a testable unit than against `main`, so this relocation is in scope.

## Acceptance criteria

- [ ] All four replay conditions produce byte-identical final balances.
- [ ] Cursor and balance commit in one transaction, asserted by a test that fails the balance write.
- [ ] Cold start from an empty database succeeds against a mocked RPC.
- [ ] Large-integer amounts round-trip with no precision loss; no `float64` remains in the amount path.
- [ ] Indexer logic lives in `internal/stellar`, not `main.go`.
- [ ] PRD B-01, B-02, B-03 can be checked off with the tests as evidence.

## Notes

This is the gate on unblocking the indexer. Until the replay harness passes, "do not ship" should stand.

Damola09:
## Problem

The event indexer was marked **blocked, do not ship**. The failure mode is the worst category for a savings product: nothing errors, the balance is simply wrong, and it stays wrong.

- **B-01** — the cursor advanced in a *separate* transaction from the balance write, and a failed event was skipped with `continue`. A balance write could fail while the cursor moved past it, losing the event permanently.
- **B-02** — cold start returned ledger `0`, which the Stellar RPC rejects, so fresh deployments never started indexing.
- **B-03** (as scoped in this issue: duplicate-delivery idempotency) — no proof duplicate delivery was safe.
- **B-11** — precision loss on large amounts.

## Changes

**Relocation.** Indexing logic moved into `EventPoller.PollEvents` (`internal/stellar/poller.go`). `startEventIndexer` now owns only scheduling and telemetry. The poller depends on an `EventFetcher` interface, so the harness drives the **same** production code with a recorded fixture — there is no test-only reimplementation. `applyEventMutation` is shared by the poller and the admin sync so the two cannot drift.

**Transactional cursor + state (B-01).** Each event's state mutation, its `processed_events` row, and the cursor advance to that event's ledger now commit in **one transaction**. `advanceCursorTx` deliberately bypasses `systemstate.Repository`, which owns its own `*sql.DB` and would commit separately. A failing event now **stops the pass** instead of being skipped past.

The cursor also never advances past the last event actually committed. The RPC pages results, so the poller cannot know a batch was complete; jumping to the tip on an incomplete batch skipped undelivered events. It advances to the tip only when a poll returns nothing.

**Idempotency (B-03).** `processed_events.event_id` is claimed with `INSERT … ON CONFLICT DO NOTHING` inside the mutation transaction — database-enforced, with no `SELECT`-then-`INSERT` window to race through.

**Cold start (B-02).** `resolveStartLedger` treats both an absent key **and** a stored `"0"` as never-indexed. The stored-zero case is the actual mechanism: migration `025` seeds the key with literal `'0'`, so a fresh database has a present-but-zero cursor. Cold start derives `tip - offset` (default 12, configurable), clamped to ≥1, and fails loudly if the tip cannot be read.

**Precision (B-11).** Corrected after review — an earlier version of this description overclaimed. B-11 has two halves, and this PR fixes only one:

- *Parsing (pre-dates this PR).* Amounts already decoded via `UseNumber()` into `decimal.Decimal`. The `case float64` branch is a **bounds check, not a pure guard**: it rejects non-integral values and magnitudes above 2^53, but still converts smaller values via `int64(v)` (exact in that range). `extractEventAmount` is **not touched by this diff**.
- *Persistence (fixed here).* Migration `103` widens vault balance columns from `NUMERIC(20,8)` to `NUMERIC(48,8)`. The old type allowed only 12 integer digits and raised `numeric field overflow` at ≥10^12, so a 1e18 stroop deposit — an ordinary vault deposit — was rejected outright. This was the half still broken.

**Harvest events** are now handled (`yield_earned` + `current_balance`), which the issue requires and the switch lacked.

## The four replay conditions

All produce byte-identical final state, comparing **every** mutated dimension (`total_deposited`, `current_balance`, `yield_earned`, `status`) — comparing one total would let a corrupted dimension pass unnoticed.

| Scenario | Test |
|---|---|
| Clean single pass | `TestIntegrationReplay_CleanSinglePass` |
| Restart mid-stream | `TestIntegrationReplay_RestartMidStream` |
| Duplicate delivery | `TestIntegrationReplay_DuplicateDelivery`, `TestIntegrationReplay_RepeatedFullReplay` |
| Out-of-order within ledger | `TestIntegrationReplay_OutOfOrderWithinLedger` |

Clean-pass balances are computed **by hand** from the fixture, not captured from output — a test that only compares runs to each other passes happily on a uniformly wrong implementation. Restart uses a fresh poller instance sharing no memory, so only the persisted cursor carries progress. Permutations are deterministic (no RNG), so failures always reproduce.

**Ordering:** across ledgers order is *required and enforced* (`sortEventsForApply`); within a ledger it is *not required*, because balance mutations are additive deltas and addition commutes. Boundary documented: `CHECK (current_balance >= 0)` means a permutation moving a withdrawal ahead of its funding deposit fails loudly rather than corrupting state, and the poller retries rather than skipping.

## Transactionality test

`TestIntegrationCursorAndBalanceCommitAtomically` forces the balance write to fail against the **real** `CHECK` constraint and asserts the cursor did not move, no partial state committed (including no `processed_events` row that would suppress a retry), and the event is still retryable. `TestIntegrationPollStopsAtFailedEvent` proves the poller does not strand a failed event behind an advanced cursor.

## Not covered: ledger reorganisations

Flagged in review and now documented rather than left implicit. **Reorgs are out of scope here and unhandled by `EventPoller`.**

`ReorgSafeIndexer` and migration `100_reorg_safe_indexer` exist on `dev` with **no non-test callers**, and this PR's poller references none of them. Two concrete incompatibilities are documented for whoever wires them together:

1. `advanceCursorTx` uses `GREATEST`, so the cursor cannot rewind to a fork point. A code comment warns against simply dropping it — `GREATEST` is what stops out-of-order delivery walking the cursor back over committed events.
2. `revertToCheckpoint` deletes `processed_events`/`ledger_checkpoints` rows above the fork point but **never rewinds the cursor**, so on its own it would drop reverted events rather than replay them.

Reorg handling is outside #1051's acceptance criteria; folding an unreviewed reorg path into this change would weaken the guarantees actually proven here. Tracking as follow-up.

## Validation

Run locally against PostgreSQL 16:

- `go build ./...` — clean
- `go vet ./...` — clean
- `gofmt` — clean on all touched files
- Unit suite `go test -short ./...` — **0 failures**
- `go test -run Integration -p 1 ./...` — **all pass**
- Migration collision check (CI's own script) — no collisions (renumbered 102→103; see below)
- Migration `103` verified reversible in both directions

**Regression checks — the tests were confirmed to fail against the old behavior**, not just pass against the new:
- Reverting the atomic cursor → `TestIntegrationPollStopsAtFailedEvent` fails
- Reverting cold start to `0` → both `TestColdStart_*` fail with `startLedger=0, which the Stellar RPC rejects`
- Reverting the float64 bounds check → `TestExtractEventAmount_RejectsUnsafeFloat64` fails

**Migration renumbered 102 → 103.** #1068 (transactional outbox) also adds a `102_` migration, and CI's duplicate-prefix guard would hard-fail whichever merged second. Ours moved because it has no dependents; #1068 keeps `102` untouched.

**Two things I could not verify locally and am flagging rather than claiming:**

1. **`-race` was not run locally** — no C toolchain on this machine (`cgo: C compiler "gcc" not found`). CI's Linux runner exercises it, and `API (Go)` passed. The poller is synchronous and adds no new goroutines or shared state.
2. **Parallel integration runs** — `handler`, `repository/postgres`, and `service` fail when integration packages run concurrently against one shared database. **This is pre-existing on `dev`**: I verified the identical three packages fail on a clean `dev` checkout with these commits stashed. With `-p 1` everything passes. My harness follows the established `testutil.ApplyAllMigrations` convention used by those same packages; fixing repo-wide test isolation is out of scope here.

## Docs

`docs/event-indexer-replay.md` covers the invariant, transaction boundary, processed-event identity, cold-start derivation, ordering guarantees, the reorg gap, precision strategy, how to run the harness, and how to update the fixture safely. PRD blocking items are checked off with per-item test evidence.

One naming note: **PRD B-03 is a settlement BOLA issue**, while this issue uses B-03 for indexer idempotency. The doc maps both readings so the reference is unambiguous.

## Fixture

`internal/stellar/testdata/replay_events.json` — a `getEvents` response over ledgers 58101–58125, two vault contracts, containing deposits, withdrawals, harvests, pause/unpause, multiple events per ledger, and amounts up to 1e18 stroops. It is **synthetic but schema-faithful** — hand-authored to the exact wire shape the production parser consumes, decoded through the same `UseNumber` settings, rather than captured from live testnet. Hermetic; CI never touches the network.

Closes #1051


0xDeon (approved):

Issue #1058, opened 2026-08-16
feat(api): generated OpenAPI contract with client codegen and CI drift detection


0xDeon:
## Problem

The DApp, the future mobile client, and any partner integration all talk to the same Go API, and there is no machine-readable contract between them. The API's shape lives in Go handler code; the client's assumptions live in hand-written TypeScript. Nothing checks that the two agree.

The failure mode is silent and reaches users directly. A handler adds a required field, changes a status code, renames a JSON key, or makes a nullable field non-nullable. Go compiles. The DApp compiles. Both test suites pass, because each side tests against its own idea of the contract. The break appears in the browser.

This gets worse, not better, with time. Every additional client multiplies the number of places an assumption can drift, and #1048 is about to change error response shapes across every endpoint — exactly the kind of cross-cutting change that needs a contract to land safely.

## Approach

Make the contract a build artifact rather than a convention. Generate an OpenAPI specification from the Go source as the single source of truth, generate the TypeScript client from that specification, and fail CI when the checked-in specification and the code disagree.

Direction matters: generate the spec **from the code**, not the other way round. A hand-maintained spec drifts from the implementation and becomes actively misleading — worse than no spec, because people trust it.

## Scope

**Specification generation**

- Annotate handlers and produce OpenAPI 3.1 covering every route: path and query parameters, request bodies, response bodies per status code, authentication requirements, and rate-limit headers.
- Model nullability and optionality precisely. `omitempty` on a Go struct field and an optional field in the schema are not the same thing, and conflating them is a common source of client-side null dereferences.
- Cover error responses using the taxonomy from #1048, so every endpoint documents which typed errors it can return.
- Commit the generated specification. Its diff in review is the mechanism by which a breaking change becomes visible to a human.

**Client generation**

- Generate the TypeScript client from the specification and replace hand-written fetch code in the DApp.
- Generated types flow into components, so a field removal becomes a compile error rather than a runtime undefined.

**Drift detection in CI**

- Regenerate the specification and fail if it differs from the committed copy. This is the check that gives the whole exercise teeth.
- Run a breaking-change detector against the specification on the base branch. Removing a field, narrowing a type, adding a required request field, or changing a status code should fail the build with a message naming the break.
- Breaking changes remain possible, but require an explicit acknowledgement label on the PR rather than passing silently.

**Contract tests**

[...]

PR #1168, opened 2026-08-27
Simulated deposit quotes, missing require_auth on upgrades, deploy verification

### LINKED ISSUE #1129: feat(dapp): show the real fee and expected output before signing
## Context

A user should know what a transaction costs and what they will receive before they sign. `estimateDepositFee` exists in the transaction lib; the UI must actually use it.

## Acceptance criteria

- [ ] Network fee shown before signing, from simulation rather than a constant.
- [ ] Expected shares received, or assets returned, shown for the entered amount.
- [ ] A material change between quote and execution is surfaced rather than silently accepted.
- [ ] Displayed values match what the chain returns, verified in an E2E test.

---

### LINKED ISSUE #1130: fix(dapp): verify every core screen against real backend responses
## Context

Frontend tests largely use fixtures. Fixture drift from the real API is invisible until a user hits it.

## Acceptance criteria

- [ ] Contract tests asserting the app's expected shapes against real API responses.
- [ ] A response-shape change breaks CI rather than production.
- [ ] Every core screen covered.
- [ ] Fixtures generated from real responses, not hand-written.

## Notes

Complements the OpenAPI work in #1058.

---

### LINKED ISSUE #1131: docs(contracts): publish deployed testnet addresses and verification steps
## Context

`deployed-testnet.env` holds contract IDs with no record of which commit produced them or how to verify a deployed wasm matches the source.

## Acceptance criteria

- [ ] Deployed addresses recorded with the commit and wasm hash that produced them.
- [ ] Documented steps to verify a deployed contract against source.
- [ ] Redeploy procedure and its consequences for existing state.
- [ ] Kept current as part of the deploy script rather than by hand.

---

*Testnet launch backlog — E9 Contract readiness.*

### LINKED ISSUE #1132: feat(contracts): confirm require_auth coverage on every privileged entrypoint
## Context

A missing require_auth is the classic Soroban vulnerability and a single omission is total compromise of that function.

## Acceptance criteria

- [ ] Every entrypoint enumerated with its intended authorization.
- [ ] A test per privileged entrypoint asserting an unauthorized caller fails.
- [ ] Adding an entrypoint without a matching test fails CI.
- [ ] Any deliberately public entrypoint is documented as such.

---

*Testnet launch backlog — E9 Contract readiness.*

ANTIDOT20:
## Summary
- **#1129**: added `previewDeposit()` (calls the vault's read-only `preview_deposit` via `simulateTransaction`) and wired it into the deposit modal — `sharesReceived` was previously just the raw deposit amount, not a real quote.
- **#1130**: added `contracts.ts` with a zod schema + validator for the `APYHistoryResponse` shape, and a test suite for it — a concrete starting point for contract testing, not hand-written fixtures.
- **#1132**: found and fixed a real bug — `propose_upgrade`/`cancel_upgrade` in `allocation_strategy`, `treasury`, `yield_registry`, and `vault` called `AccessControl::require_role` without first calling `admin.require_auth()`, unlike every other admin-gated entrypoint in this codebase. See `AUTH_AUDIT.md` for the finding, impact, and how it was located.
- **#1131**: `deploy-testnet.sh` now records the git commit and wasm sha256 for each deployed contract in `deployed-testnet.env`; `VERIFY.md` documents how to verify a deployed contract against source and what a redeploy does/doesn't affect.

## Issues
Closes #1129
Closes #1130
Closes #1132
Closes #1131

## Scope notes
- **#1129**: I noticed `buildDepositTransaction`'s `deposit` call passes only `(user, amount)` while the vault contract's `deposit` signature takes a third `min_shares_out` param — I did not change this, since verifying whether that's a real mismatch or an intentional legacy/wrapper path needs more context than I could confirm safely. Flagging it here rather than guessing.
- **#1130**: covers one endpoint as a concrete pattern. Extending to every core screen, wiring a CI step against live/recorded responses, and generating schemas from the OpenAPI spec (#1058) are real follow-up work, noted in `contracts.ts`.
- **#1132**: this fixes one confirmed pattern found via automated enumeration + manual verification, not an exhaustive per-entrypoint audit with a CI gate — see the "Scope not covered" section of `AUTH_AUDIT.md`.

## Test plan
- [x] `cargo check --tests -p allocation-strategy-contract -p treasury-contract -p yield-registry-contract -p vault-contract` compiles clean
- [x] `cargo test -p allocation-strategy-contract privileged_strategy_calls_require_signatures` passes with the fix (extended to cover `propose_upgrade`/`cancel_upgrade`)
- [x] `bash -n` on `deploy-testnet.sh`
- Frontend changes: no local toolchain to run `tsc`/`vitest` here — please run in CI


0xDeon:
Merged into dev via #1175, which combines this with #1169 and fixes the two failing CI jobs.

On the Contracts (Rust) failure: the require_auth finding turned out to be a false positive. `propose_upgrade`/`cancel_upgrade` delegate to `nester_common::Upgrade`, which calls `require_auth()` as its first statement, so the signature check was already enforced one level down — the library's doc comment spells out that split (library owns auth, entrypoint owns the role check). Adding a second `require_auth()` in an already-authorized frame is rejected by Soroban, which is what broke the four lifecycle tests. Reverted those eight lines and rewrote AUTH_AUDIT.md accordingly.

Worth saying: the audit method was sound and the write-up was careful about its own limits — it just needed to resolve candidates through shared-library calls, not only one level of `Self::..._internal` delegation. Your new regression assertions in `privileged_strategy_calls_require_signatures` are kept, since that auth guarantee is inherited rather than local and a refactor could drop it silently.

The deposit preview, contract schemas, and deploy verification all went in as-is. The only other change was to the zod validator, which was discarding `issue.path` so drift reports never named the field. Thanks for the contribution.

Issue #1193, opened 2026-08-28
fix(api): pending-deposit valuation silently truncates at 10,000 rows


0xDeon:
## Context

Several read paths substitute a large constant for real pagination:

- `apps/api/internal/valuation/adapters.go:78` — `TxPendingSource.PendingDeposits` uses `Limit: 10000`
- `apps/api/internal/valuation/adapters.go:35` — `VaultPositionSource.Positions` uses `PerPage: 10000`
- `apps/api/internal/service/portfolio_service.go:31` — `PerPage: 10000`
- `apps/api/internal/service/performance/service.go:223` — `PerPage: 1000`

The valuation one is the sharp edge. Pending deposits feed a user's valuation, so passing the limit does not merely paginate badly — it undercounts money. The user sees a total that is wrong in a specific direction, with nothing in the response indicating truncation.

A cap is a reasonable design; a cap that cannot be distinguished from "that was all of it" is not.

## Acceptance criteria

- [ ] Valuation sums every pending deposit for the user, or the response states explicitly that it is partial.
- [ ] Money-path aggregates never silently drop rows past a limit.
- [ ] The remaining list paths either paginate or document why the ceiling is safe.
- [ ] Test with more rows than the limit asserts the total is correct or the truncation is surfaced.

## Notes

The comment at `portfolio_service.go:31` — "Reasonable upper limit for number of vaults per user" — is probably true today. The valuation and reconciliation ceilings are the ones that change meaning as the product grows.

---

*Post-testnet backlog — found during a follow-up codebase audit.*

Goodness-0x:
@Goodness-0x has applied to work on this issue as part of the **Stellar Wave Program**'s 8th wave.



> Hello, I have reviewed the issue and understand what needs to be done.
> 
> I would appreciate being assigned to it.
> 


ℹ️ **Repo Maintainers**: To accept this application, [review their application](https://www.drips.network/wave/maintainers/issues/31d36c82-7cd7-412c-bf8e-ef0e91743d76) or assign @Goodness-0x to this issue.

Sajjadd-dev:
@Sajjadd-dev has applied to work on this issue as part of the **Stellar Wave Program**'s 8th wave.



> Hello, I’d like to contribute to this issue and would appreciate the opportunity to work on it. Thank you.


[...]

Issue #1192, opened 2026-08-28
fix(api): reconciliation only ever checks the first 500 vaults


0xDeon:
## Context

`BalanceComparator.Reconcile` lists vaults with a fixed page and never advances it. At `apps/api/internal/reconciliation/comparators.go:34`:

```go
filter := vault.ListFilter{Status: string(vault.StatusActive), Limit: 500, Offset: 0}
```

There is no loop over offsets. Once there are more than 500 active vaults, the remainder are never compared against on-chain state by any run, ever — and reconciliation is precisely the mechanism meant to catch balances that have drifted.

The failure is silent and gets worse with growth: the runs keep succeeding and reporting no divergence, while the proportion of the book that is actually checked falls as the product grows. A clean reconciliation report would be evidence of nothing.

## Acceptance criteria

- [ ] Reconciliation pages through every vault in scope rather than the first page.
- [ ] The number of vaults compared is recorded per run, so a run that covered a fraction of the book is visibly distinguishable from one that covered all of it.
- [ ] Test with more vaults than one page asserts every vault is compared.
- [ ] A page-fetch failure mid-run fails the run rather than silently shortening it.

## Notes

Consider whether an unbounded sweep is the right shape at scale, or whether it should be a rolling window with an explicit coverage SLO — either is defensible, but the current behaviour claims full coverage while delivering partial.

---

*Post-testnet backlog — found during a follow-up codebase audit.*

abrahambaba1:
@abrahambaba1 has applied to work on this issue as part of the **Stellar Wave Program**'s 8th wave.



> I'd  to work on this issue, please assign it to me


ℹ️ **Repo Maintainers**: To accept this application, [review their application](https://www.drips.network/wave/maintainers/issues/827316be-46dc-4e4a-bc1e-e781dbdcc34c) or assign @abrahambaba1 to this issue.

TheShnider:
@TheShnider has applied to work on this issue as part of the **Stellar Wave Program**'s 8th wave.



> Hi. I have reviewed this issue and would like to work on it.


ℹ️ **Repo Maintainers**: To accept this application, [review their application](https://www.drips.network/wave/maintainers/issues/827316be-46dc-4e4a-bc1e-e781dbdcc34c) or assign @TheShnider to this issue.

Nathanpvd:
@Nathanpvd has applied to work on this issue as part of the **Stellar Wave Program**'s 8th wave.



[...]

PR #1255, opened 2026-08-28
Claude/testnet observability safety kshtma

### LINKED ISSUE #1111: feat(api): structured request logging with a correlation id end to end
## Context

Diagnosing a failed deposit means stitching together API logs, indexer logs and chain state by hand. A correlation id carried across all three makes that one query.

## Acceptance criteria

- [ ] A request id generated at the edge, returned in the response, and present on every log line for that request.
- [ ] The id propagates into background jobs and chain submissions spawned by the request.
- [ ] No PII, wallet secrets or tokens in log output.
- [ ] A test asserts the id appears on downstream job logs.

---

*Testnet launch backlog — E5 Observability and ops.*

### LINKED ISSUE #1119: feat(api): per-user and global caps for the testnet launch
## Context

A launch with no caps means a bug or an attack has unbounded blast radius. Caps make the first weeks survivable.

## Acceptance criteria

- [ ] Configurable per-user deposit cap and global TVL cap.
- [ ] Requests exceeding a cap are rejected with a clear reason.
- [ ] Caps are changeable without redeploy.
- [ ] Approaching a cap is alerted before it is hit.
- [ ] Tests cover both boundaries.

---

*Testnet launch backlog — E6 Launch gate.*

### LINKED ISSUE #1124: feat(api): audit trail for every balance-changing operation
## Context

When a user disputes a balance, the team needs to reconstruct exactly how it got there. Derived state alone cannot answer that.

## Acceptance criteria

- [ ] Every balance change is appended with: actor, operation, before, after, chain reference and timestamp.
- [ ] Append-only; the application has no update or delete path.
- [ ] A user's balance history can be reconstructed and reconciled to the current balance.
- [ ] Retention documented.
- [ ] Test: replaying the audit trail reproduces the current balance exactly.

## Notes

Lighter than the full double-entry ledger in #822 and #1052 — this is the launch-minimum.

---

*Testnet launch backlog — E7 Data correctness.*

### LINKED ISSUE #1125: fix(dapp): replace remaining placeholder data on core screens with live data
## Context

Several screens still render placeholder or hardcoded values. On a savings product a fabricated number is worse than an empty state.

## Acceptance criteria

- [ ] An audit of the core screens listing every value and its real source.
- [ ] Placeholder values replaced with live data or an explicit empty state.
- [ ] No screen invents a number when data is unavailable.
- [ ] Loading, empty, error and populated states exist for each.

## Notes

The error and skeleton work has landed; this is about the data behind it.

---

promisszn:
## Summary

Implements four items from the testnet launch backlog:

- **#1111** — Structured request logging with a correlation id propagated end-to-end, including into background jobs.
- **#1119** — Configurable per-user deposit cap and global TVL cap, enforced with clear rejection reasons and pre-cap alerting.
- **#1124** — Append-only audit trail for every balance-changing operation, reconstructable and reconcilable to current balance.
- **#1125** — Replaced remaining placeholder/mock data on core dapp screens with live API data or explicit empty states.

### Correlation ID propagation (#1111)
- Fixed `internal/domain/jobqueue/worker.go` to inject the request-scoped correlation id and logger into the `context.Context` passed to job handlers (previously only used for the worker's own log lines).
- `jobqueue/client.go` `Enqueue` now auto-inherits the request id from context when not explicitly set.
- Added `TestWorker_PropagatesCorrelationIDToHandlerContextAndLogs`.
- PII/secret audit of log call sites found no violations.
- Known gap: chain-submission logging (`soroban_vault_chain_invoker.go`) still uses an injected logger, not ctx-based — not yet correlation-tagged.

### Per-user & global caps (#1119)
- New `internal/domain/caps` package, config-driven via `LAUNCH_PER_USER_DEPOSIT_CAP`, `LAUNCH_GLOBAL_TVL_CAP`, `LAUNCH_CAP_WARN_THRESHOLDS_PCT` env vars — changeable without redeploy.
- Wired into `VaultService.RecordDeposit`, with live SUM queries in `VaultRepository`.
- Requests over a cap are rejected with `422` and a clear `USER_DEPOSIT_CAP_EXCEEDED` / `GLOBAL_TVL_CAP_EXCEEDED` reason.
- Warn-at-80%/90% threshold logging in `cmd/api/main.go`.
- Tests cover under/at/over-cap boundaries for both caps, disabled caps, and warn-once-per-threshold behavior.

### Balance audit trail (#1124)
- Migration `107_balance_audit_log` — append-only table: actor, operation, before, after, chain_reference, timestamp.
- New `internal/domain/balanceaudit` package — only `Append`/`ListByVault`/`ListByUser`, no update/delete path.
- Wired (best-effort, non-blocking) into `RecordDeposit`, `RecordWithdrawal`, `HarvestVault`.
- `docs/balance-audit-trail.md` documents retention.
- Test replays a deposit/deposit/withdraw sequence and confirms `Reconcile()` reproduces the live balance exactly.
- Known gap: rebalance and emergency-withdraw legs are not yet wired to this ledger.

### Live data on core dapp screens (#1125)
- `app/settings/page.tsx` — KYC tab was fully mocked; now wired to real `GET/POST /api/v1/users/kyc/{userId}`.
- `app/offramp/page.tsx` — removed hardcoded `MOCK_BALANCE`, hardcoded KYC status, and a fabricated settlement tx link; withdraw now calls the real settlement API.
- `app/dashboard/analytics/page.tsx` — removed `BenchmarkCard`'s invented APY/bank/inflation numbers until a real data source exists.
- Deleted unused, fully mock (`Math.random()`-driven) components: `dashboard-stats`, `portfolio-charts`, `recent-activity`, `vault-positions-table`, and `lib/mock-data.ts`.
- Known gap: the offramp LP-node quote-comparison display still simulates comparative rates pre-submission (actual settlement submission is now real).

## Test plan
- [x] `go test ./...` in `apps/api` — all pass
- [x] `npx vitest run` in `apps/dapp/frontend` — 252/252 pass
- [x] `tsc --noEmit` and `eslint` clean on touched files

Closes #1111
Closes #1119
Closes #1124
Closes #1125

## Summary by CodeRabbit

* **New Features**
  * Added configurable per-user deposit and global TVL caps, with warnings as limits are approached.
  * Added an append-only balance history for deposits, withdrawals, harvests, and related balance changes.
  * Added live KYC status retrieval and document submission.

* **Bug Fixes**
  * Prevented unauthorized access to other users’ KYC information.
  * Improved cleanup of uploaded KYC documents after failed submissions.
  * Offramp withdrawals now use actual available balances and submit transactions to the API.

* **Updates**
  * Removed placeholder analytics benchmarks and mock dashboard data while live data integration is pending.
  * Added clearer KYC loading, retry, submission, and refresh error states.


apps/api/migrations/107_create_balance_audit_log.up.sql
```diff
@@ -0,0 +1,65 @@
+-- Append-only audit trail for every balance-changing vault operation
```
0xDeon: Blocking, and the only thing standing between this and a merge: the `107_` prefix collides. `dev` already has `107_add_kyc_identity_fields.{up,down}.sql`, landed via #1265/#1268 after this branch was cut.

Git merges it cleanly because the filenames differ, so this won't show as a conflict — but CI's duplicate-prefix guard in `ci.yml` fires on the merged tree, and the migration runner would either error on the duplicate version or silently skip one of them. I ran the guard's exact logic against the post-merge file list and it reports `DUPLICATE PREFIX: 107`.

Renaming both files to `108_create_balance_audit_log.{up,down}.sql` clears it. Worth rebasing on `dev` at the same time — the branch is 24 commits behind and its green CI ran against the older base.

apps/api/internal/repository/postgres/balance_audit_repository.go
```diff
@@ -0,0 +1,177 @@
+package postgres
```
0xDeon: `ListByVault` and `ListByUser` both `SELECT ... ORDER BY created_at ASC, id ASC` with no `LIMIT`/`OFFSET`, on a table the migration describes as append-only. The doc comments say they return *every* entry, so the unboundedness is intentional — but that's the pattern #1193 and #1192 were just filed and fixed for elsewhere in this codebase.

Neither method is reachable from a handler in this diff, so nothing is broken today. It's a footgun for the first caller though. Adding the page-and-limit shape now, matching what `reconciliation/comparators.go` does on `dev`, would be cheaper than retrofitting it later.

---

Separately from the requested changes — the substance of this PR is good and I want to be clear it's close. The correlation-id work threads `logpkg.WithRequestID` into the handler context with an honest fallback for cron jobs and asserts it in `worker_test.go`, which satisfies #1111 literally. The launch caps take `pg_advisory_xact_lock` on the user and global keys inside `RecordDepositWithAudit`, so the cap check isn't racy under concurrent deposits, and malformed cap values fail startup rather than silently disabling the cap. The `mock-data.ts` removal is clean — no orphaned importers.

Rename the migration and rebase, and I'll merge it.

apps/api/internal/handler/user_handler.go
```diff
@@ -223,6 +223,11 @@ func (h *UserHandler) submitKYC(w http.ResponseWriter, r *http.Request) {
 		response.WriteJSON(w, http.StatusBadRequest, response.ValidationErr("invalid user ID"))
 		return
 	}
+	// Ownership check (IDOR fix): only the authenticated user may submit KYC
```
0xDeon: The IDOR fix here is real and still needed — I checked `dev` and neither `submitKYC` nor `getKYCStatus` calls `authorizeUserAccess`, though the helper exists and is used elsewhere in the file.

One thing to re-check after rebasing: this hunk anchors on `ParseMultipartForm`'s old position, and `dev` now has a `kycStore == nil` guard inserted before it from #1265/#1268. Please confirm the ownership check still lands *before* the storage guard once rebased, rather than after it.

Same caution for `cmd/api/main.go` — #1263, #1265 and #1268 all touched that file, so the line offsets have moved under this branch's caps-wiring hunks. Git will apply them by context match, which usually works but is worth eyeballing rather than trusting.

0xDeon (changes requested):

PR #1263, opened 2026-08-28
fix: bound async goroutines, effectiveness stats, reconciliation logging, valuation pagination

### LINKED ISSUE #1193: fix(api): pending-deposit valuation silently truncates at 10,000 rows

### LINKED ISSUE #1194: fix(api): reconciliation failures can go unrecorded
## Context

The reconciliation engine discards the error from its own failure-recording call, in two places. At `apps/api/internal/reconciliation/engine.go:98` and `:115`:

```go
_ = e.repo.FailRun(ctx, run.ID, err.Error())
```

If `FailRun` itself fails — the write that exists to record that something went wrong — the run ends with no durable trace. The operator sees a run that started and never reported, which is indistinguishable from a process that was killed, and the actual error is gone.

This matters more than an ordinary swallowed error because it is the error path of the error path: it only triggers when something is already wrong, which is exactly when the evidence is most needed.

## Acceptance criteria

- [ ] A failed `FailRun` is logged with the original error it was trying to record, so the cause survives even when the write does not.
- [ ] A run that could not be marked failed is not left indistinguishable from one still in progress.
- [ ] Test drives a failing `FailRun` and asserts the original error still reaches the logs.

## Notes

Same shape as the audit-log write at `admin_handler.go:317` and the login activity write at `auth_handler.go:190` — worth a single decision about which best-effort writes are allowed to fail silently and which must leave a trace.

---

*Post-testnet backlog — found during a follow-up codebase audit.*

### LINKED ISSUE #1196: fix(api): nudge effectiveness reports perfect conversion on any query error
## Context

`GetEffectivenessStats` collapses failure and cold-start into the best possible score. At `apps/api/internal/repository/postgres/nudge_history_repository.go:66`:

```go
if err != nil || sent == 0 {
    return nudge.EffectivenessStats{ConversionRate: 1.0}, nil // Cold start safe
}
```

A database error and "no data yet" produce the same answer: a perfect 1.0 conversion rate, returned with a nil error. The ranking engine at `nudge_engine_service.go:118` then treats that nudge as the best-performing one available.

So the failure mode is inverted — a broken query does not suppress a nudge, it promotes it above every nudge with real measured performance. The query itself is also acknowledged as incomplete in the comment above it ("let's just return a stub"): the `LEFT JOIN` has no `GROUP BY` and no date bound, so the counts are not the counts the ranking assumes.

## Acceptance criteria

- [ ] A query error is returned to the caller rather than being reported as a conversion rate.
- [ ] Cold start is represented distinctly from measured performance, so ranking can treat "unknown" differently from "perfect".
- [ ] The effectiveness query is bounded by time window and grouped correctly.
- [ ] Test asserts a failing query does not cause a nudge to outrank one with real data.

## Notes

Whether cold-start nudges should be optimistically promoted is a product decision worth making explicitly — the current behaviour makes it by accident, and makes the same choice for errors.

---

*Post-testnet backlog — found during a follow-up codebase audit.*

### LINKED ISSUE #1198: fix(api): milestone notification goroutines have no timeout
## Context

`notifyMilestonesAsync` spawns a goroutine per milestone with an uncancellable context and no deadline. At `apps/api/internal/service/savings_goal_service.go:968` the call uses bare `context.Background()`.

The sibling path immediately above it, at line 954, is bounded by `streakNotifyTimeout` — so the pattern is established in the same file and this one diverges from it.

A notifier that stops responding therefore leaks a goroutine per milestone indefinitely, with no upper bound and nothing to reclaim them. It is the kind of leak that stays invisible until the day traffic makes it visible all at once.

## Acceptance criteria

- [ ] Milestone notifications run under a bounded context, consistent with the streak path.
- [ ] A hung notifier cannot accumulate goroutines without limit.
- [ ] Test asserts the goroutine exits when the notifier exceeds the timeout.

## Notes

Two related uncancellable contexts worth the same treatment: `valuation/adapters.go:139` (`PushToUser`, which also drops its error) and `service/nudge_copy.go:34` (an outbound LLM call that the caller therefore cannot abort).

---

*Post-testnet backlog — found during a follow-up codebase audit.*

davedumto:
## Summary

Four independent backend bug fixes from a post-testnet codebase audit, each in its own commit.

### #1198 — Milestone notification goroutines have no timeout

`notifyMilestonesAsync` spawned a goroutine per milestone with an uncancellable `context.Background()` and no deadline — a hung notifier could leak goroutines per milestone indefinitely. Bounded it the same way the sibling streak-milestone goroutine already was (renamed `streakNotifyTimeout` → `asyncNotifyTimeout` since it now covers both call sites).

Also fixed the two related sites the issue flagged:
- `WSNotifier.PushValuation` — was called with `context.Background()` and no bound, and dropped its error. Now bounded by a 10s timeout and logs the push failure instead of discarding it.
- `CompositeCopyGenerator`'s outbound LLM call — didn't take a `context.Context` at all, so the caller genuinely could not abort it. Threaded `ctx` through from `NudgeEngineService` (the one caller), bounded by a 10s timeout on the LLM call specifically so a slow/hung provider falls back to the template within a bounded time.

### #1196 — Nudge effectiveness reports perfect conversion on any query error

`GetEffectivenessStats` returned `ConversionRate: 1.0, nil` for both a database error and a genuine cold start (zero dispatches). The ranking engine then promoted a failing query above every nudge type with real measured performance.

`EffectivenessStats` now carries a `HasData` flag. A query error is returned to the caller instead of swallowed; cold start reports `HasData: false` (zero-value `ConversionRate`, distinct from a measured 0%). `Rank()` skips the effectiveness multiplier entirely when `HasData` is false — same treatment it already gave a missing map entry.

Also fixed the query itself: bounded to a 90-day window, and switched `COUNT(*)` to `COUNT(DISTINCT ...)` on both sides of the `LEFT JOIN` — the original was double-counting a dispatch once per matching outcome row.

### #1194 — Reconciliation failures can go unrecorded

Both `FailRun` call sites in the reconciliation engine discarded their own error (`_ = e.repo.FailRun(...)`). If `FailRun` itself failed, the run ended with no durable trace at all — indistinguishable from a killed process, with the real error gone. This is the error path of the error path: it only triggers when something's already wrong, exactly when the evidence matters most.

Added `recordFailedRun`, which logs both the original error and the `FailRun` error together when the write fails, so the cause survives even when the write doesn't.

Left the sibling swallowed-error sites the issue notes (`admin_handler.go:317`, `auth_handler.go:190`) untouched — the issue itself frames those as "worth a single decision" about which best-effort writes may fail silently, not something to decide unilaterally as part of this fix.

### #1193 — Pending-deposit valuation silently truncates at 10,000 rows

`TxPendingSource.PendingDeposits` capped at `Limit: 10000` with no pagination. Pending deposits feed a user's valuation directly, so a user past that cap saw an undercounted total with nothing indicating truncation. Now pages through every row via `ListUserTransactions`' own reported total count (`ListUserTransactions` already supports real `Limit`/`Offset` pagination with an independent `COUNT(*)`, so this was straightforward to wire up correctly).

The other two `10000`/`1000`-row constants the issue flagged (`VaultPositionSource.Positions`, `GetUserPortfolioSummary`, `GetUserAnalytics`) are bounded by **vault count**, not transaction volume — a user can only grow that number through the product UI, not through ordinary usage over time. Documented that reasoning explicitly at each site instead of rewriting them to paginate, per the issue's own "either paginate or document why the ceiling is safe" framing.

## Testing

- `go build ./...`: clean.
- `go vet ./...`: clean.
- `go test ./...`: full module passes, no failures (integration tests requiring `TEST_DATABASE_DSN` skip as expected, same as before this change).
- `gofmt -l`: clean on every file I authored or touched meaningfully. (This repo has substantial pre-existing `gofmt` drift across ~70 unrelated files, confirmed via `git show upstream/dev:<path> | gofmt -l` before touching anything — not introduced or worsened here.)
- New/expanded test coverage: `TestRank_ColdStartEntryDoesNotOutrankRealData`, `TestGetEffectivenessStats_*` (4 cases, via sqlmock matching this repo's existing `postgres` package convention), `TestEngineLogsOriginalErrorWhenFailRunItselfFails` / `TestEngineDoesNotLogWhenFailRunSucceeds`, `TestTxPendingSource_PendingDeposits_*` (5 cases covering pagination, summing, empty/single-page, and error propagation), `TestWSNotifier_PushValuation_*` (3 cases).

Closes #1198
Closes #1196
Closes #1194
Closes #1193

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved nudge ranking by preventing cold-start entries from receiving an unwarranted conversion boost.
  - Limited effectiveness calculations to recent measured activity and surfaced data errors correctly.
  - Preserved original reconciliation errors when failure recording also fails.
  - Ensured valuation data includes all pending deposits across multiple pages.

- **Reliability**
  - Added time limits and fallback behavior for generated nudge copy and milestone notifications.
  - Added timeout handling and failure logging for valuation updates.

### DIFF
--- a/apps/api/cmd/api/main.go
+++ b/apps/api/cmd/api/main.go
@@ -546,7 +546,7 @@ func run() error {
 		Goals:     valuation.NewGoalAllocationSource(postgres.NewSavingsGoalRepository(db)),
 		Oracle:    valuation.NewStaticOracle(nil),
 		Cache:     valuation.NewCache(30 * time.Second),
-		Notifier:  valuation.NewWSNotifier(wsHub),
+		Notifier:  valuation.NewWSNotifier(wsHub, baseLogger.WithGroup("valuation")),
 		Logger:    baseLogger.WithGroup("valuation"),
 	})
 	valuationHandler := handler.NewValuationHandler(valuationService)
--- a/apps/api/internal/domain/nudge/ranking.go
+++ b/apps/api/internal/domain/nudge/ranking.go
@@ -42,6 +42,12 @@ type Outcome struct {
 
 type EffectivenessStats struct {
 	ConversionRate float64 // 0 to 1
+	// HasData is false for a cold-start nudge type (no dispatches yet in the
+	// measurement window). Rank treats a cold-start entry the same as a
+	// missing one — no effectiveness boost — rather than the misleading
+	// "perfect conversion" a zero-value ConversionRate would otherwise imply
+	// (nester#1196).
+	HasData bool
 }
 
 type HistoryRepository interface {
@@ -69,7 +75,7 @@ func Rank(candidates []Candidate, segment usersignal.Segment, engagement usersig
 		score := def.BaseImpact
 
 		stats, ok := effectiveness[c.Type]
-		if ok {
+		if ok && stats.HasData {
 			score = score * (0.5 + stats.ConversionRate) // just a heuristic for the test
 		}
 
--- a/apps/api/internal/domain/nudge/ranking_test.go
+++ b/apps/api/internal/domain/nudge/ranking_test.go
@@ -51,13 +51,38 @@ func TestRank_EffectivenessBreaksTieBetweenEqualBaseImpact(t *testing.T) {
 	// Both types share BaseImpact 0.7 in the catalog; a higher historical
 	// conversion rate for one should be what breaks the tie.
 	stats := map[NudgeType]EffectivenessStats{
-		NudgeTypeMilestone:       {ConversionRate: 0.1},
-		NudgeTypeStreakMilestone: {ConversionRate: 0.9},
+		NudgeTypeMilestone:       {ConversionRate: 0.1, HasData: true},
+		NudgeTypeStreakMilestone: {ConversionRate: 0.9, HasData: true},
 	}
 
 	ranked := Rank(candidates, usersignal.SegmentActiveSaver, usersignal.TierEngaged, stats)
 
 	if ranked[0].Candidate.Type != NudgeTypeStreakMilestone {
 		t.Fatalf("expected the higher-converting nudge type to rank first, got %s", ranked[0].Candidate.Type)
+	}
+}
+
+func TestRank_ColdStartEntryDoesNotOutrankRealData(t *testing.T) {
+	userID := uuid.New()
+	candidates := []Candidate{
+		{Type: NudgeTypeMilestone, UserID: userID},
+		{Type: NudgeTypeStreakMilestone, UserID: userID},
+	}
+	// Both types share BaseImpact 0.7. A cold-start entry (HasData: false,
+	// the zero-value ConversionRate) must not be treated as a perfect 1.0
+	// conversion rate: with the old behavior a cold-start entry would score
+	// 0.7*(0.5+1.0)=1.05, beating even this strong 0.7 measured conversion
+	// rate (0.7*(0.5+0.7)=0.84). With the fix, cold start gets no boost at
+	// all and stays at the unscaled BaseImpact of 0.7 — so the type with
+	// real, strong measured performance must win (nester#1196).
+	stats := map[NudgeType]EffectivenessStats{
+		NudgeTypeMilestone:       {ConversionRate: 0.7, HasData: true},
+		NudgeTypeStreakMilestone: {HasData: false}, // cold start: zero-value ConversionRate
+	}
+
+	ranked := Rank(candidates, usersignal.SegmentActiveSaver, usersignal.TierEngaged, stats)
+
+	if ranked[0].Candidate.Type != NudgeTypeMilestone {
+		t.Fatalf("expected the nudge type with real measured data to rank first over a cold-start entry, got %s", ranked[0].Candidate.Type)
 	}
 }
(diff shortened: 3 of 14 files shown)


0xDeon (approved):

PR #1269, opened 2026-08-28
fix: post-testnet audit backlog — typed auth error, share-price tests, UUID coercion, analytics aggregation

### LINKED ISSUE #1195: fix(api): analytics report one vault's history as the whole portfolio
## Context

`GetUserAnalytics` builds a multi-vault user's entire daily series from a single vault. At `apps/api/internal/service/performance/service.go:240-247` the snapshot loop reads `userVaults[0]` with the comment "For now, we'll use the latest snapshot as a placeholder".

Two handlers serve this live: `analytics_handler.go:124` and `performance_snapshots_handler.go:41`.

A user with several vaults is shown a balance and yield history that belongs to one of them, presented as their portfolio. Nothing marks it partial. The number is not approximate — it is a different quantity wearing the right label, which is the kind of wrong a user cannot detect and will make decisions on.

Related: `vaultMonthlyYield` at lines 259-261 is declared and never populated, so that field is empty in every response.

## Acceptance criteria

- [ ] Daily snapshots aggregate across all of the user's vaults.
- [ ] `VaultMonthlyYield` is populated or removed from the response rather than always returning empty.
- [ ] Test with a two-vault user asserts the series reflects both.
- [ ] A user with no vaults gets an empty series rather than an error or a zero-filled one.

## Notes

The placeholder comment suggests this was known at the time. Worth confirming whether any UI is presenting these numbers as authoritative today.

---

*Post-testnet backlog — found during a follow-up codebase audit.*

### LINKED ISSUE #1197: fix(api): bank account IDs fall back to the zero UUID on parse failure
## Context

The bank account repository ignores UUID parse errors and uses the zero value that results. At `apps/api/internal/repository/postgres/bank_account_repository.go:97-98` and `:141`:

```go
parsedID, _ := uuid.Parse(id)
```

A malformed id does not fail the request. It becomes `00000000-0000-0000-0000-000000000000` and is used as a real identifier against the database.

The consequence depends on the query, which is what makes it worth fixing rather than reasoning about case by case: a zero UUID in a `WHERE` clause silently matches nothing, and in an `INSERT` or an ownership check it is a value an attacker can reach deterministically by sending a malformed id. Neither is behaviour anyone chose.

## Acceptance criteria

- [ ] A malformed UUID returns a validation error rather than being coerced.
- [ ] No code path uses a zero UUID as a real identifier.
- [ ] Test asserts a malformed id is rejected on every affected method rather than returning empty results.

## Notes

Worth a sweep for the same `uuid.Parse(x)` with a discarded error elsewhere in the repository layer — this is a pattern, not a one-off.

---

*Post-testnet backlog — found during a follow-up codebase audit.*

### LINKED ISSUE #1199: test(contracts): three share-price tests assert nothing

### LINKED ISSUE #1200: fix(contracts): set_allocations panics untyped on an authorization failure
## Context

`allocation_strategy::set_allocations` rejects an unauthorized caller with a bare panic. At `packages/contracts/contracts/allocation_strategy/src/lib.rs:319`:

```rust
if !AccessControl::has_role(&env, &operator, Role::Operator) {
    panic!("Caller is not an authorized operator");
}
```

Everywhere else in the crate uses `panic_with_error!` with a typed `ContractError`. A bare panic produces an untyped host error, so a caller cannot match on it — an authorization refusal becomes indistinguishable from an internal failure at the call site, and the string is not part of any stable interface.

For a privileged entrypoint this is the error most worth being able to identify: a client that cannot tell "you are not an operator" from "the contract broke" cannot retry correctly or report accurately.

## Acceptance criteria

- [ ] The check uses `panic_with_error!` with the crate's `ContractError::Unauthorized` (or equivalent).
- [ ] Test asserts an unauthorized caller receives the typed error rather than a host panic.
- [ ] A sweep confirms no other privileged entrypoint in the crate uses a bare `panic!` for an auth failure.

## Notes

Small and mechanical, but it is on an entrypoint that changes allocation — the kind of call where an operator reading logs during an incident needs the failure to be unambiguous.

---

*Post-testnet backlog — found during a follow-up codebase audit.*

Depo-dev:
Four independent findings from the post-testnet codebase audit. One commit per issue — see each commit message for the full reasoning.

- `set_allocations` refused unauthorized callers with a bare `panic!`, so a client could not tell an auth refusal from an internal failure. Now `panic_with_error!(ContractError::Unauthorized)`, with the three auth tests tightened from `.is_err()` to typed assertions. Closes #1200
- Three share-price tests were `assert!(true)` stubs, reporting the share-price arithmetic as covered when nothing exercised it. All three now run against real contract state via `NesterHarness`. Closes #1199
- The bank account repository discarded `uuid.Parse` errors, so a malformed id became the zero UUID and was used as a real identifier. Now rejected with a wrapped error on every affected method. Closes #1197
- `GetUserAnalytics` built a multi-vault user's entire daily series from `userVaults[0]`. Now aggregates across all vaults with carry-forward across un-snapshotted days, and populates the previously always-empty `VaultMonthlyYield`. Closes #1195

Verified locally: `go build ./...` and `go vet` clean; Go tests pass in performance, postgres, and handler; `cargo test` passes 52 integration + 74 allocation_strategy tests.

Out of scope, flagged for follow-up: the same discarded-`uuid.Parse` pattern exists in `savings_goal_repository.go`, `savings_schedule_repository.go`, `savings_streak_repository.go`, and `protocol_tvl_repository.go`.

## Summary by CodeRabbit

- **New Features**
  - Analytics now aggregate daily balances across all vaults, carry balances across missing snapshot days, and calculate monthly yield per vault.
  - Share-price behavior now supports accurate deposits, withdrawals, yield accrual, and minimum deposits with appropriate rounding.

- **Bug Fixes**
  - Invalid bank account and user identifiers now return clear parsing errors instead of being silently accepted.
  - Unauthorized allocation changes now return a specific authorization error.

- **Tests**
  - Added coverage for analytics aggregation, share-price scenarios, identifier validation, and authorization failures.

### DIFF
--- a/apps/api/internal/repository/postgres/bank_account_repository.go
+++ b/apps/api/internal/repository/postgres/bank_account_repository.go
@@ -94,8 +94,14 @@ func (r *BankAccountRepository) ListByUser(ctx context.Context, userID uuid.UUID
 		); err != nil {
 			return nil, err
 		}
-		parsedID, _ := uuid.Parse(id)
-		parsedUser, _ := uuid.Parse(uid)
+		parsedID, err := uuid.Parse(id)
+		if err != nil {
+			return nil, fmt.Errorf("bank account repository: parse id %q: %w", id, err)
+		}
+		parsedUser, err := uuid.Parse(uid)
+		if err != nil {
+			return nil, fmt.Errorf("bank account repository: parse user id %q: %w", uid, err)
+		}
 		var verified *time.Time
 		if verifiedAt.Valid {
 			verified = &verifiedAt.Time
@@ -138,7 +144,11 @@ func (r *BankAccountRepository) GetByID(ctx context.Context, id uuid.UUID) (bank
 	if err != nil {
 		return bankaccount.BankAccount{}, nil, "", mapBankAccountError(err)
 	}
-	parsedUser, _ := uuid.Parse(uid)
+	parsedUser, err := uuid.Parse(uid)
+	if err != nil {
+		return bankaccount.BankAccount{}, nil, "", fmt.Errorf(
+			"bank account repository: parse user id %q: %w", uid, err)
+	}
 	var verified *time.Time
 	if verifiedAt.Valid {
 		verified = &verifiedAt.Time
--- a/apps/api/internal/repository/postgres/bank_account_repository_test.go
+++ b/apps/api/internal/repository/postgres/bank_account_repository_test.go
@@ -1,0 +1,132 @@
+package postgres
+
+import (
+	"context"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/DATA-DOG/go-sqlmock"
+	"github.com/google/uuid"
+)
+
+// A malformed identifier must surface as an error rather than being coerced to
+// the zero UUID and used as a real identifier (nester#1197). Each affected
+// method is exercised separately: silently returning uuid.Nil in a domain
+// object is what let a bad value reach a WHERE clause or an ownership check.
+
+func TestBankAccountRepository_ListByUser_RejectsMalformedID(t *testing.T) {
+	db, mock, err := sqlmock.New()
+	if err != nil {
+		t.Fatalf("sqlmock.New: %v", err)
+	}
+	defer db.Close()
+
+	userID := uuid.New()
+	cols := []string{
+		"id", "user_id", "bank_name", "bank_code", "account_last4",
+		"account_name", "currency", "country", "is_default", "verified_at", "created_at",
+	}
+	mock.ExpectQuery("SELECT id, user_id, bank_name").
+		WithArgs(userID.String()).
+		WillReturnRows(sqlmock.NewRows(cols).AddRow(
+			"not-a-uuid", userID.String(), "Kuda", "090267", "1234",
+			"Ada Lovelace", "NGN", "NG", true, nil, time.Now(),
+		))
+
+	repo := NewBankAccountRepository(db)
+	accounts, err := repo.ListByUser(context.Background(), userID)
+	if err == nil {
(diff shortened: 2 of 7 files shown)


0xDeon (approved):
This looks good. Nice work

PR #1270, opened 2026-08-28
fix(api): serialise concurrent operations on the same vault position

### LINKED ISSUE #1084: fix(api): serialise concurrent operations on the same vault position
## Context

Two concurrent withdrawals on the same position can both read the pre-withdrawal balance and both pass the sufficient-funds check. Nothing in the handler takes a row lock.

## Acceptance criteria

- [ ] Money-path writes take `SELECT ... FOR UPDATE` on the position row, or use an equivalent serialisation.
- [ ] Integration test: N concurrent withdrawals of the full balance result in exactly one success.
- [ ] The lock is held for the minimum span and cannot deadlock against the deposit path — lock ordering documented.

---

*Testnet launch backlog — E1 Money-path integrity.*

Salmatcre8:
Closes #1084

## Problem

Two concurrent withdrawals on the same position could both read the pre-withdrawal balance (service-level `GetVault` is a plain `SELECT`) and both pass the sufficient-funds check. `VaultRepository.RecordWithdrawal` then decremented unconditionally — nothing on the money path took a row lock. The only thing stopping a negative position was the `check_current_balance` constraint, which surfaces as a raw SQLSTATE 23514 error on a nondeterministic subset of the racing requests.

## Fix

- `VaultRepository.RecordWithdrawal` now locks the position row with `SELECT ... FOR UPDATE` and re-runs the sufficiency check **under the lock**, returning `ErrWithdrawalExceedsPosition` to the losers (the handler already maps this to a 400, same as the pre-check).
- The service-level check stays as a fast-fail so a doomed withdrawal is never submitted on-chain; the repository check is authoritative.
- **Minimum lock span:** the transaction contains only the locked `SELECT`, the balance `UPDATE`, and the ledger `INSERT` — no chain or network I/O happens inside it.
- **Lock ordering (documented on `RecordWithdrawal`):** every money-path write locks exactly one `vaults` row first — the deposit path's single atomic `UPDATE` takes the same row lock, an equivalent serialisation — and only then inserts into `vault_transactions`. No path locks a second vaults row or takes the two in the other order, so deposit and withdrawal queue on one lock and cannot deadlock.
- The in-memory test repository mirrors the new contract.

## Acceptance criteria

- [x] Money-path writes take `SELECT ... FOR UPDATE` on the position row (withdrawal), or an equivalent serialisation (deposit's atomic `UPDATE`).
- [x] Integration test: 8 concurrent withdrawals of the full balance → exactly **1 success**, 7 × `ErrWithdrawalExceedsPosition`, final balance 0, exactly 1 ledger row (`TestVaultRepositoryIntegrationRecordWithdrawalConcurrent`).
- [x] Lock held for the minimum span; lock ordering documented; interleaved deposits/withdrawals complete without deadlock (`TestVaultRepositoryIntegrationDepositWithdrawalInterleaved`).

## Verification

- `go build ./...`, `go vet` clean.
- Full `internal/service` + `internal/repository/postgres` suites pass against a real Postgres (`TEST_DATABASE_DSN`), including the two new tests.
- Reverting only the repository change makes the new concurrency test fail (all goroutines race past the check; constraint errors instead of exactly-one-success), confirming the test catches the bug.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved withdrawal processing under concurrent activity to prevent balances from becoming negative or being withdrawn more than once.
  * Withdrawals now return a clear insufficient-balance error when funds are unavailable.
  * Interleaved deposits and withdrawals reliably preserve the correct final balance.

* **Tests**
  * Added coverage for concurrent withdrawals and mixed deposit-withdrawal activity.

### DIFF
--- a/apps/api/internal/repository/postgres/vault_repository.go
+++ b/apps/api/internal/repository/postgres/vault_repository.go
@@ -439,6 +439,22 @@ func (r *VaultRepository) UpdateHarvestFrequency(ctx context.Context, id uuid.UU
 
 // RecordWithdrawal decrements current_balance atomically and writes a ledger
 // entry. It does NOT touch total_deposited (deposits are never reversed).
+//
+// Serialisation (nester#1084): the position row is locked with
+// SELECT ... FOR UPDATE and the sufficient-funds check re-runs under that
+// lock. Two concurrent withdrawals therefore cannot both read the same
+// pre-withdrawal balance and both pass — the second waits on the row lock and
+// re-checks against the post-withdrawal balance. The service-layer check
+// stays as a fast-fail before any on-chain submit; this one is authoritative.
+//
+// Lock ordering (deadlock safety): every money-path write — RecordDeposit,
+// RecordWithdrawal, RecordHarvest, applyConfirmedBalanceChange — locks
+// exactly one vaults row first (the deposit path's single atomic UPDATE takes
+// the same row lock, an equivalent serialisation) and only then inserts into
+// vault_transactions. No money path locks a second vaults row or takes the
+// two in the other order, so deposit and withdrawal cannot deadlock; they
+// queue on the same row lock. The lock spans only the statements below — no
+// network or chain I/O happens inside the transaction.
 func (r *VaultRepository) RecordWithdrawal(ctx context.Context, id uuid.UUID, record vault.TransactionRecord) error {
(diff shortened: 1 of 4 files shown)


0xDeon (approved):
Everything looks good.Great work

PR #1271, opened 2026-08-29
feat(api): reconcile indexed vault balances against on-chain state on a schedule

### LINKED ISSUE #1082: feat(api): reconcile indexed vault balances against on-chain state on a schedule
## Context

The indexer applies events to derive balances. If it ever misses or misapplies one, the database silently diverges from the chain and nothing notices. There is no reconciliation job.

## Acceptance criteria

- [ ] A scheduled job reads authoritative balances from the vault contract and compares them to the database.
- [ ] Any divergence is recorded, alerted, and exposed as a metric — never silently auto-corrected.
- [ ] The job is safe to run against production data and has a dry-run mode.
- [ ] Test with a deliberately corrupted row proves the divergence is detected.

## Notes

This is the safety net that makes the rest of the money path trustworthy.

---

*Testnet launch backlog — E1 Money-path integrity.*

Salmatcre8:
Closes #1082

## What this does

The reconciliation engine, comparators, classifier, and audit tables (#887, migration 085) were complete but **entirely orphaned** — nothing constructed them, no scheduler ran them, no alerter existed, and no metrics were emitted. This PR wires the money path's safety net end to end. No new migration is needed; the audit schema from 085 is used as-is.

### The scheduled job (criterion 1)
- New `reconciliation.Runner` drives the existing `Engine` + `BalanceComparator` on a leader-gated loop (advisory-lock leadership, same gate as the submission reconciler): immediate first pass, then every `RECONCILE_INTERVAL` (default **5m**, matching the package's own balance cadence). `RECONCILE_ENABLED` defaults **on**.
- The authoritative read is `total_assets()` via Soroban simulation — no key material, shared `contractReader` with breaker + retry instrumentation.

### The unit contract (please review this decision)
The comparison happens in **raw stroops** via a new `ContractReader.TotalAssetsStroops` + `StroopsBalanceReader` adapter. The event indexer stores vault balances *as emitted* (migration 103 widened the columns for exactly this; `docs/event-indexer-replay.md` fixture rule 7), so exact-integer vs exact-integer is the only comparison that can't hide single-stroop bookkeeping errors. **Heads-up:** the pre-#1082 service-layer write paths store display USDC into the same column — that live incoherence is upstream of this PR, and the reconciler will surface it as real record-vs-chain divergences. The runbook gains a **First rollout** section: run `RECONCILE_DRY_RUN=true` on an existing deployment first, triage, then go live. `TotalAssets` (display) now carries a doc warning against wiring it into reconciliation.

### Recorded, alerted, metered — never corrected (criterion 2)
- Findings persist to `reconciliation_findings` through the existing repository; a decorator increments `nester_reconcile_divergences_total{kind}` **per persisted finding**, so the metric and the audit row cannot disagree. The existing `ReconciliationDivergence` page alert fires on any of them.
- `LogAlerter` logs critical/warning findings with the join-key fields (vault id, both values, difference) that the cardinality policy deliberately keeps off the metric.
- Nothing writes to `vaults` — the integration test asserts the corrupted balance is untouched.

### Liveness that can't lie
- New scrape-time pull collector `nester_reconcile_balance_last_run_age_seconds` (freshness-collector pattern): the tx poller resets the shared gauge every 15s, so sharing it would let either loop die invisibly behind the other.
- Emitted **only by the elected leader**; the anchor advances **only on successful passes** and re-anchors on leadership gain — a hanging first pass, a permanently failing loop, and a leaderless cluster all read as stalled/absent, never clean, and a failover cannot emit a follower's stale boot-age (no false page).
- New alerts **BalanceReconciliationStalled** (>1800s, page) and **BalanceReconciliationMetricsAbsent** (page — disabling the safety net is loud by design), with recording rule, promtool unit tests (healthy / stalled-while-poller-healthy / boundary / absent), dashboard line, runbook Cause E, and metrics.md docs. `ReconciliationFailing` + `MoneyPathMetricsAbsent` + `ReconciliationDivergence` texts updated for two loops, promtool expectations in lockstep.

### Production safety + dry-run (criterion 3)
- Per-vault chain-read failures (uninitialized contract, i128 past the reader's documented int64 bound) are **logged and skipped** — one bad caller-supplied contract address cannot become a standing denial of the sweep, and findings already collected survive. If **every** read fails, the run fails: an RPC outage or open breaker can never read as all-clear.
- `RECONCILE_DRY_RUN=true` runs the full read-and-compare against production data, logs each would-be finding with both values, writes nothing, counts nothing, alerts nothing; liveness still emits.
- `CompleteRun`/`FailRun` are detached from cancellation (`context.WithoutCancel`) so SIGTERM mid-sweep cannot strand run rows in status `running`; a shutdown-cancelled pass is not recorded as a failed run (no `ReconciliationFailing` tick on every deploy).

### Corrupted-row proof (criterion 4)
`TestReconciliationIntegrationDetectsCorruptedRow` (real Postgres + full migration chain) seeds a healthy vault and a vault whose DB balance is 1e7 stroops above the stubbed chain: exactly one **critical mismatch** finding with exact recorded/on-chain/difference values, run row completed with checked=2/findings=1/critical=1, divergence metric incremented once, and the corrupted balance **unchanged**. The dry-run variant proves the same detection from the logs with zero rows written.

## Verification
- `go build ./...`, `go vet` clean; full `./internal/...` suite passes against real Postgres (`TEST_DATABASE_DSN`), including new unit tests for the runner (leadership re-anchoring, success-only liveness, shutdown cancellation, ticker recurrence), comparator skip semantics, the stroops-vs-display unit contract (httptest Soroban RPC stub, incl. int64 overflow), the metrics collector, and the log alerter.
- `promtool test rules` passes for all three suites (run via `prom/prometheus` docker image, same as CI).
- Dashboards regenerated via `scripts/build_slo_dashboards.py`; only the money-path dashboard changed.

## Known scope notes
- Findings are re-recorded per run for a persisting divergence (one row per run per divergent vault) — this mirrors the tx poller re-counting stuck transactions each pass and keeps `ReconciliationDivergence` firing while disagreement persists; resolution workflow is the existing `resolution_state` column.
- Only `status='active'` vaults are swept (the comparator's pre-existing contract). Paused vaults whose balances the indexer still mutates are a possible follow-up.
- The reader's i128-to-int64 bound (~9.2e18 stroops) is documented at `TotalAssetsStroops`; a vault past it is skipped loudly each pass, never reported truncated.

## Summary by CodeRabbit

* **New Features**
  * Added scheduled vault-balance reconciliation against authoritative on-chain balances.
  * Added configurable intervals, enablement, and dry-run mode.
  * Divergences are recorded and surfaced through logs and metrics without automatic correction.
* **Bug Fixes**
  * Individual chain-read failures no longer abort otherwise successful reconciliation passes.
  * Shutdown and cancellation now preserve reconciliation status updates.
* **Observability**
  * Added dedicated liveness metrics, dashboard panels, and alerts for balance reconciliation.
* **Documentation**
  * Updated metrics references and operational runbooks with reconciliation behavior and troubleshooting guidance.

### DIFF
--- a/apps/api/.env.example
+++ b/apps/api/.env.example
@@ -235,6 +235,22 @@ TX_POLLER_INTERVAL=15s
 # Horizon time to ingest a freshly submitted transaction.
 TX_POLLER_MIN_AGE=30s
 
+# Scheduled vault-balance reconciliation (nester#1082): reads the
+# authoritative total_assets from each active vault's contract and compares
+# it to vaults.current_balance in raw stroops, recording — never correcting —
+# any divergence to the reconciliation audit tables, the log, and the
+# divergence metric. Disabling the money path's safety net is expected to
+# page (BalanceReconciliationMetricsAbsent) rather than pass silently.
+RECONCILE_ENABLED=true
+
+# How often the balance reconciler sweeps the vault book. If raised past 30m,
+# the BalanceReconciliationStalled alert threshold must move with it.
+RECONCILE_INTERVAL=5m
+
+# Dry-run rehearses a full sweep against production data: findings are logged
+# but nothing is written, counted, or alerted. Liveness still emits.
+RECONCILE_DRY_RUN=false
+
 # Payment provider keys for offramp features (bank list, account resolution).
 # At least one must be set in production or staging — the API will refuse to
 # start if both are empty in those environments.
--- a/apps/api/cmd/api/main.go
+++ b/apps/api/cmd/api/main.go
@@ -38,6 +38,7 @@ import (
 	"github.com/suncrestlabs/nester/apps/api/internal/notifications"
 	"github.com/suncrestlabs/nester/apps/api/internal/objectstorage"
 	"github.com/suncrestlabs/nester/apps/api/internal/oracle"
+	"github.com/suncrestlabs/nester/apps/api/internal/reconciliation"
 	"github.com/suncrestlabs/nester/apps/api/internal/repository"
 	"github.com/suncrestlabs/nester/apps/api/internal/repository/postgres"
 	"github.com/suncrestlabs/nester/apps/api/internal/retry"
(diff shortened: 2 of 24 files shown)


0xDeon (approved):
Lgtm.....Good work!

Issue #1085, opened 2026-08-25
feat(api): durable submission record so a lost RPC response cannot double-submit


0xDeon:
## Context

If Soroban RPC times out after accepting a transaction, the API does not know whether it landed. A naive retry double-submits; not retrying loses the operation.

## Acceptance criteria

- [ ] Every chain submission writes an intent record before submitting, with a client-generated idempotency reference.
- [ ] On timeout the record stays pending and a reconciler polls the chain for the outcome rather than resubmitting blind.
- [ ] A submission is only ever retried when the chain proves it did not land.
- [ ] Test simulates: response lost after success, response lost after failure, and RPC unavailable throughout.

## Notes

Pairs with #8.

---

*Testnet launch backlog — E2 Chain boundary.*

Lurdings:
@Lurdings has applied to work on this issue as part of the **Stellar Wave Program**'s 8th wave.



> Hi Maintainer, I’m interested in contributing to this task and can deliver a clean solution within the timeline.


ℹ️ **Repo Maintainers**: To accept this application, [review their application](https://www.drips.network/wave/maintainers/issues/a9ac5053-c854-4c94-9928-225c92ec19d0) or assign @Lurdings to this issue.

0xDeon:
Merged into `integration/chain-boundary-1085-1086-1087-1088` via #1171 rather than into `dev`, so GitHub would not close this automatically. Closing by hand.

Follow-up hardening for the gaps found during review is going out in a separate PR against `dev`.

Issue #1080, opened 2026-08-25
test(contracts): property-based invariant suite for share accounting


0xDeon:
## Context

The rounding rules in `vault/src/conversion.rs` are documented and look correct, but nothing generates adversarial sequences against them. Share-accounting bugs are exactly the class that unit tests miss and money loses.

## Acceptance criteria

- [ ] Property tests over random sequences of deposit, withdraw, yield accrual and fee events.
- [ ] Invariants asserted: sum of user assets never exceeds total assets; share price never decreases absent a loss event; round-tripping deposit then immediate withdraw never returns more than was put in; no operation panics on valid input.
- [ ] Runs in CI with a fixed seed plus a nightly randomised run.
- [ ] Any counterexample found is committed as a regression test.

## Notes

Overlaps open #1050 in spirit — check before duplicating; this one is scoped to the vault money path only.

---

*Testnet launch backlog — E1 Money-path integrity.*

Kaizer4show:
@Kaizer4show has applied to work on this issue as part of the **Stellar Wave Program**'s 8th wave.



> Hello, I would love to work on this issue


ℹ️ **Repo Maintainers**: To accept this application, [review their application](https://www.drips.network/wave/maintainers/issues/36db7266-a528-47cd-a3c2-808780404a5e) or assign @Kaizer4show to this issue.

Lurdings:
@Lurdings has applied to work on this issue as part of the **Stellar Wave Program**'s 8th wave.



> Hi Maintainer, I’m interested in contributing to this task and can deliver a clean solution within the timeline.


ℹ️ **Repo Maintainers**: To accept this application, [review their application](https://www.drips.network/wave/maintainers/issues/36db7266-a528-47cd-a3c2-808780404a5e) or assign @Lurdings to this issue.

adelekevictor12:
@adelekevictor12 has applied to work on this issue as part of the **Stellar Wave Program**'s 8th wave.



> Hello maintainer, i can fix this issue


ℹ️ **Repo Maintainers**: To accept this application, [review their application](https://www.drips.network/wave/maintainers/issues/36db7266-a528-47cd-a3c2-808780404a5e) or assign @adelekevictor12 to this issue.

Classique-seyi:
@Classique-seyi has applied to work on this issue as part of the **Stellar Wave Program**'s 8th wave.



[...]

PR #1282, opened 2026-08-30
fix: post-testnet audit — config, dead submission scaffold, vault BOLA, harvest dilution

### LINKED ISSUE #1150: security(api): add ownership checks to the vault endpoints that leak other users' balances
## Context

Three endpoints return balance data with no ownership check:

- `getSharePrice` (vault_handler.go:877-891) and `convert` (vault_handler.go:893-922) never call `auth.GetUserFromContext`, and the response includes total assets and total shares.
- `listVaults` (`GET /api/v1/vaults/all`, vault_handler.go:181-224) never reads the auth context and returns vault records whose JSON includes `user_id`, `contract_address`, `total_deposited` and `current_balance`.

Any authenticated user can page the entire platform's vault ledger, or enumerate vault UUIDs to read any individual user's balance. `getVault` in the same file does the check correctly, so this is an inconsistency rather than a design decision.

## Acceptance criteria

- [ ] Share-price and convert endpoints require ownership, or expose only non-sensitive aggregate data.
- [ ] `/vaults/all` is restricted to admins, or removed.
- [ ] Non-owner access to an existing vault returns 404, not 403.
- [ ] These routes are added to the authorization matrix test in #1101 so the gap cannot reopen.

## Notes

Found during a follow-up API audit. Same BOLA class the team already fixed in settlement_handler.

---

### LINKED ISSUE #1153: fix(api): wire or delete the unused SubmissionPipeline and ReorgSafeIndexer subsystems
## Context

Two complete, migration-backed subsystems exist and are never constructed in `main.go`:

- `SubmissionPipeline` (stellar/submission_pipeline.go) with `AllocateSequence`, `RecordSubmission`, `ResolveTimeout`, `RecoverOnStartup`, backed by `migrations/099_chain_submissions`.
- `ReorgSafeIndexer` (stellar/reorg_indexer.go) with parent-hash verification and checkpoint rollback, backed by `migrations/100_reorg_safe_indexer`.

The tables ship, so the system appears to have submission recovery, sequence reservation and reorg safety. In production none of it runs. Worse, three functions inside the pipeline are stubs: `checkOnChainStatus` returns `StatusUnknown` unconditionally, `seedSequenceFromNetwork` returns 0, and `computeTransactionHash` sha256s the base64 envelope string rather than computing a real Stellar transaction hash.

## Acceptance criteria

- [ ] A decision is made and recorded: wire these up, or delete them along with their migrations.
- [ ] If wired: the three stubs are implemented, and tests cover timeout resolution, startup recovery, sequence allocation and a reorg that replaces a seen ledger.
- [ ] If deleted: the dead code and unused tables go, so nothing implies coverage that does not exist.
- [ ] Either way the chain-boundary issues #1085 and #1089 reference the outcome.

## Notes

Found during a follow-up API audit. Wiring these may be cheaper than writing #1085 and #1089 from scratch — assess before starting either.

---

### LINKED ISSUE #1154: fix(api): drop the fiat provider requirement from staging and production config validation
## Context

`config.go:944-947` refuses to boot in staging or production unless `PAYSTACK_SECRET_KEY` or `FLUTTERWAVE_SECRET_KEY` is set. The fiat offramp feature has been abandoned, so this now blocks any staging deploy unless someone supplies a dummy credential or sets a wrong `APP_ENV` — and a wrong `APP_ENV` also disables the JWT-default guard and the CORS origin guard on lines 816 and 951.

## Acceptance criteria

- [ ] The fiat provider requirement is removed from config validation.
- [ ] No other abandoned-offramp config remains mandatory.
- [ ] A staging deploy boots with no fiat credentials configured.
- [ ] The JWT and CORS guards remain enforced in staging.

## Notes

Direct consequence of dropping the offramp phase. Small, and it unblocks the staging environment in #1114.

---

### LINKED ISSUE #1159: fix(contracts): stop harvest_vault diluting holders with the aggregate performance fee
## Context

#1078 (fixed by PR #1157) removed holder dilution from the per-user `harvest` path: the performance fee now burns fee-equivalent shares from the harvesting user, so assets and supply fall together and the exchange rate is preserved for everyone else.

The same pattern remains in the **admin-level** `harvest_vault` path.

`harvest_vault_internal` (`packages/contracts/contracts/vault/src/lib.rs`, the block ending around line 2344 on PR #1157's head) transfers `total_fee_collected` to the treasury and then does:

```rust
let total_assets = get_total_assets(&env);
let post_fee_assets = total_assets
    .checked_sub(total_fee_collected)
    .unwrap_or_else(|| panic_with_error!(&env, ContractError::ArithmeticOverflow));
set_total_assets(&env, post_fee_assets);
```

Total assets fall, total share supply does not. Every holder's share price drops by the fee amount, exactly as in the original #1078 bug. `harvest_vault` is a public entrypoint guarded only by `Role::Admin`, so this is reachable in normal operation.

The fix used in #1157 does not transfer directly. That fee is charged to a single harvesting user, so burning that user's shares is correct. `harvest_vault` collects an **aggregate** fee across all users, and there is no single holder to charge — Soroban cannot iterate positions unbounded, which is the stated reason this entrypoint exists at all. So this needs its own design decision rather than a copy of the #1157 approach.

Options worth weighing:

- Mint fee-equivalent shares to the treasury instead of reducing assets, so supply rises rather than assets falling. Dilution then lands proportionally and transparently, and the treasury holds a real claim.
- Track fees as a liability excluded from `get_net_total_assets` so share price never reflects money already owed to the treasury.
- Remove `harvest_vault` and rely solely on per-user harvest, if aggregate collection is no longer needed now that #1157 charges correctly.

## Acceptance criteria

- [ ] A test reproduces the dilution through `harvest_vault` and fails against current code.
- [ ] Calling `harvest_vault` does not reduce the share price for holders who did not harvest.
- [ ] The chosen mechanism is documented in the contract with the reasoning, in the same style as the #1157 comment.
- [ ] The treasury still receives the full fee, asserted by test.
- [ ] The property suite from #1080 covers `harvest_vault` alongside per-user `harvest`.

## Notes

Found while reviewing PR #1157. That PR is correct for the path it covers and should not be held for this — the two fees need different mechanisms.

---

*Testnet launch backlog — money-path integrity.*

davidugorji:
Four follow-up audit issues. Reasoning for each is in its commit message.

- Closes #1154 — drop the fiat provider requirement from config validation, with the staging JWT/CORS guards pinned by test.
- Closes #1153 — delete the unwired SubmissionPipeline and its tables; #1085's SubmissionStore already does this correctly. Follows the precedent migration 108 set for ReorgSafeIndexer.
- Closes #1150 — ownership checks on share-price and convert (404, not 403); `/vaults/all` removed rather than re-gated, since `/admin/vaults` already serves it behind the admin role.
- Closes #1159 — remove `harvest_vault`. Its aggregate fee was a second charge on yield per-user `harvest()` already bills, taken from holders who never harvested.

Two issues named a decision rather than a fix:

**#1153** — deleted, not wired. The capability exists and is correct; wiring the scaffold would mean reimplementing its three stubs to duplicate a live subsystem against a parallel set of tables.

**#1159** — removed, not repaired, and *not* the mint-to-treasury option the issue lists first. A new claim on a fixed asset pool has to come from somewhere: minting while the fee leaves the vault lowers the share price further than the current bug does. The per-user path already collects the full fee and holds the price exactly, so the aggregate fee was double-charging (40 taken on 100 of yield at a 20% rate). Nothing outside the contract's own tests called it.

Verification: API 53 packages green; vault contract 126 tests pass including the #1080 property suite; contracts workspace builds. The #1159 dilution test was confirmed to fail against the old code first (share price 1.5 → 1.45 for a passive holder).

## Summary by CodeRabbit

- **Security**
  - Vault share-price and conversion endpoints now restrict access to vault owners.
  - Removed the platform-wide vault listing endpoint.

- **Configuration**
  - Payment-provider keys are now optional across environments.
  - Existing staging safeguards for JWT secrets and CORS settings remain enforced.

- **Vault Behavior**
  - Removed the aggregate harvest operation.
  - Improved harvest handling so one user’s harvest does not change another holder’s value, while performance fees reach the treasury accurately.

- **Maintenance**
  - Removed unused transaction-submission tracking infrastructure and database tables.


0xDeon (approved):
Nice work. Approved!

PR #1283, opened 2026-08-30
security(api): four money-path fixes from the follow-up audit (#1148, #1149, #1151, #1152)

### LINKED ISSUE #1148: security(db): add a UNIQUE constraint on vaults.contract_address
## Context

No migration creates a unique index on `vaults.contract_address`; `migrations/002_create_vaults_table.up.sql:4` declares it as plain `TEXT NOT NULL CHECK (char_length(...) > 0)`. `createVault` accepts a client-supplied address after a format-only check, and the indexer keys its balance mutations on it with `WHERE contract_address = $2`.

An attacker registers a vault pointing at a victim's contract address. The indexer's UPDATE matches both rows, so every on-chain deposit the victim makes is credited to the attacker's vault as well — and is withdrawable.

## Acceptance criteria

- [ ] A UNIQUE constraint on contract_address, accounting for soft-deleted rows.
- [ ] Vault creation rejects an address already registered, with a clear error.
- [ ] Existing duplicates are identified and resolved before the constraint is applied.
- [ ] Migration is reversible and tested against seeded data.
- [ ] Test: registering a duplicate contract address is rejected; the indexer credits exactly one vault.

## Notes

Found during a follow-up API audit. Verified against migrations/002 and indexer.go:234.

---

### LINKED ISSUE #1149: security(api): remove or constrain X-User-Id impersonation on the service API key
## Context

`apps/api/internal/middleware/auth.go:57-65`: when the bearer token equals `NESTER_SERVICE_API_KEY`, the middleware takes the user identity from an `X-User-Id` request header and constructs a session for that user. There is no allowlist of callers, no per-caller identity, and no restriction on which routes accept it.

Anyone holding the single shared service key can act as any user on every money endpoint, including deposit and withdraw. The key is shared with the intelligence service and has no rotation story. The comparison is also a plain `==` rather than a constant-time compare, unlike the JWT path two lines above which uses `hmac.Equal`.

## Acceptance criteria

- [ ] Service auth cannot assume an arbitrary user identity on money-path routes.
- [ ] If service-to-service impersonation is genuinely required, it is restricted to an explicit route allowlist and the calling service is identified separately from the key.
- [ ] The key comparison is constant-time.
- [ ] The key is validated for length and entropy at startup, and startup fails on a weak or absent value in staging and production.
- [ ] Tests: impersonation rejected on money-path routes; weak key aborts boot.

## Notes

Found during a follow-up API audit. Verified at middleware/auth.go:57-65.

---

### LINKED ISSUE #1151: fix(api): pass share units to the contract withdraw call instead of asset stroops
## Context

`vault_service.go:598` computes `stroops := input.Amount.Mul(1e7)` — an asset amount scaled to stroops — and passes it as the second argument to `WithdrawFromVault`. That parameter is named `sharesStroops` (soroban_vault_chain_invoker.go:101-106) and is forwarded to `preview_withdraw_net` and the contract's share-denominated `withdraw`.

Whenever share price is not exactly 1.0, the user burns the wrong number of shares — withdrawing materially more or less than they asked for — while the database records the requested amount either way.

## Acceptance criteria

- [ ] The withdraw path converts the requested asset amount to shares at the current share price before invoking the contract.
- [ ] The recorded withdrawal reflects what actually happened on chain, not what was requested.
- [ ] A test with share price above and below 1.0 asserts the shares burned and the assets received are both correct.
- [ ] The parameter name and unit are documented at the call site.

## Notes

Found during a follow-up API audit. Verified at vault_service.go:598 against soroban_vault_chain_invoker.go:101.

---

### LINKED ISSUE #1152: security(api): stop deposits debiting the shared operator account
## Context

`soroban_vault_chain_invoker.go:92-97` documents that it "invokes the vault contract's deposit function with the operator as both caller and depositing user". `DepositToVault(ctx, contractAddress, amountStroops)` takes no user address, and `RecordDeposit` never passes one.

So every deposit spends platform funds on the caller's behalf. `POST /vaults/{id}/deposit` is an unauthenticated-value transfer bounded only by the operator account's balance.

## Acceptance criteria

- [ ] Deposits are funded by the depositing user's own wallet, signed client-side.
- [ ] The API never submits a value-moving transaction from the operator account on a user's behalf.
- [ ] If any operator-funded path must remain, it is explicitly allowlisted, capped, and audit-logged.
- [ ] Test asserts a deposit does not reduce the operator balance.

## Notes

Found during a follow-up API audit. Overlaps #1075: once deposits are user-signed and verified on-chain, this closes with it.

---

Johnalex-hub:
Four money-path fixes from the follow-up API audit.

Closes #1148
Closes #1149
Closes #1151
Closes #1152

## #1152 operator-funded deposits
Deposits with no user-signed `tx_hash` were submitted with the operator as both caller and depositing user, spending platform funds for anyone who asked. The wallet-signed path already existed and is unchanged; the operator-funded fallback is now off by default and fails closed on every axis independently (disabled / empty allowlist / non-positive cap each refuse alone), capped per deposit, and logged on both authorization and refusal.

**Behaviour change:** a deposit with no `tx_hash` now returns 403 where it previously succeeded by spending platform money.

## #1151 withdraw units
`WithdrawFromVault`'s second parameter is share stroops, but asset stroops were passed. At share price 1.25 a request for 100 assets took 125; at 0.8 it took 32 instead of 40 — and the DB recorded the requested figure either way. Now converts assets to shares at the share price, refined against the contract's own `preview_withdraw`, and records the shares actually burned. Every `VaultDepositInvoker` amount parameter is now documented with its unit, since both are `int64` and the compiler cannot tell them apart.

## #1149 X-User-Id impersonation
Any holder of the shared service key could act as any user on any route. Money-path routes now refuse service credentials outright; the comparison is constant-time; the calling service is identified separately via `X-Service-Name`; and a configured key is validated at startup for length, entropy, and not being the JWT secret.

**Deviation:** startup does *not* require the key to be present. An empty value already disables service auth entirely (the safe state), and mandating presence broke six existing production-config tests while forcing deployments that do not use service auth to invent an unused secret. A key that *is* set is validated in every environment.

## #1148 duplicate contract address
The UNIQUE index already exists — migration 104 creates `uq_vaults_contract_address_live`, scoped to live rows so a soft-deleted vault does not burn its address. What was missing was app-layer handling: `mapRepositoryError` had no branch for that constraint, so a duplicate registration returned an opaque 500. Now returns 409 `CONTRACT_ADDRESS_REGISTERED`. No redundant migration added.

## Verification
Full `apps/api` suite passes uncached, `go vet` clean, gofmt clean on all touched files. Each security fix was mutation-tested by reverting its guard and confirming the test fails — #1151 reproduces the exact figures above, and #1152's test shows the operator balance being debited without the fix.

## Summary by CodeRabbit

* **New Features**
  * Added configurable operator-funded deposits with vault allowlists and per-deposit spending caps.
  * Withdrawal processing now accurately converts asset amounts to vault shares and records shares actually redeemed.

* **Bug Fixes**
  * Duplicate live contract registrations now return a clear conflict error.
  * Invalid share prices and refused operator-funded deposits receive clearer error responses.

* **Security**
  * Service credentials are blocked from money-related actions and require stronger configuration safeguards.

### DIFF
--- a/apps/api/cmd/api/main.go
+++ b/apps/api/cmd/api/main.go
@@ -413,6 +413,34 @@ func run() error {
 		baseLogger.Info("no signing configured: chain write operations are unavailable")
 	}
 
+	// Operator-funded deposits (nester#1152).
+	//
+	// A deposit with no user-signed tx_hash is submitted with the operator as
+	// both caller and depositing user, so it spends platform funds on the
+	// caller's behalf. Disabled unless explicitly configured, and even then
+	// only for allowlisted vaults under a per-deposit cap, with every use
+	// logged. A nil policy would refuse everything, but it is always
+	// installed so the refusals are logged rather than silent.
+	operatorFundedVaults, err := service.ParseOperatorFundedVaultIDs(cfg.Stellar().OperatorFundedDepositVaults())
+	if err != nil {
+		return fmt.Errorf("parse operator-funded deposit allowlist: %w", err)
+	}
+	operatorFundedCap, err := decimal.NewFromString(cfg.Stellar().OperatorFundedDepositMaxAmount())
+	if err != nil {
+		return fmt.Errorf("parse operator-funded deposit cap: %w", err)
+	}
+	vaultService.SetOperatorFundedDepositPolicy(service.NewOperatorFundedDepositPolicy(
+		cfg.Stellar().OperatorFundedDepositsEnabled(),
+		operatorFundedVaults,
+		operatorFundedCap,
+		baseLogger.WithGroup("operator-funded-deposits"),
+	))
+	if cfg.Stellar().OperatorFundedDepositsEnabled() {
+		baseLogger.Warn("operator-funded deposits are ENABLED: the API can spend platform funds on a user's behalf",
+			"allowlisted_vaults", len(operatorFundedVaults),
+			"per_deposit_cap", operatorFundedCap.String())
+	}
+
 	if cfg.Stellar().RPCURL() != "" {
 		vaultService.SetChainEventVerifier(service.NewStellarChainEventVerifier(cfg.Stellar().RPCURL()))
 	}
--- a/apps/api/internal/auth/context.go
+++ b/apps/api/internal/auth/context.go
@@ -12,6 +12,15 @@ type User struct {
 	Scopes        []string
 	Roles         []string
 	SessionID     string
+	// ServiceName identifies the calling service on requests authenticated
+	// with the shared service API key. Empty for ordinary user sessions.
+	//
+	// The key carries no identity of its own, so without this an audit log
+	// can say a request was made by "some holder of the service key" and
+	// nothing more (nester#1149). It is self-asserted, not authenticated:
+	// good enough to tell well-behaved callers apart in logs, not an
+	// authorization signal.
+	ServiceName string
 }
 
 // HasScope reports whether the user holds the given scope.
--- a/apps/api/internal/config/config.go
+++ b/apps/api/internal/config/config.go
@@ -994,6 +994,30 @@ func (c *Config) validate(loader *envLoader) {
 		loader.addError("AUTH_JWT_SECRET has insufficient entropy: use at least 8 distinct characters")
 	}
 
+	// NESTER_SERVICE_API_KEY authenticates service-to-service callers and is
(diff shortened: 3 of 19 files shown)


0xDeon (approved):
Lgtm.... Great work!

Issue #782, opened 2026-07-16
feat(api): add configurable rate-limiting middleware for auth, settlement, and write endpoints


0xDeon:
## Context
The Go API (`apps/api`, Chi router) currently exposes auth, settlement, vault and profile endpoints with **no rate limiting**. The only limiter in the stack is `slowapi` inside the Python intelligence service — it does not protect the Go API. Unthrottled `POST /api/v1/auth/challenge` / `verify` and `POST /api/v1/settlements` are abusable for credential-stuffing, challenge-store flooding, and offramp spam.

## What we need
Add configurable rate-limiting middleware to the Go API and apply it globally plus tighter buckets on sensitive routes.

## Scope
- Add `internal/middleware/ratelimit.go` implementing a token-bucket limiter. Prefer a **Redis-backed** limiter (the Redis client already exists) so limits hold across replicas; fall back to in-process `golang.org/x/time/rate` when Redis is absent.
- Key by client IP (respecting `X-Forwarded-For` only from trusted proxies) for anonymous routes, and by authenticated user ID for protected routes.
- Return `429 Too Many Requests` with `Retry-After` and a JSON error envelope consistent with `internal/api/response.go`.
- Config knobs in `internal/config/config.go` + `.env.example`: global RPS/burst, and a stricter override for `auth/*` and `settlements`.
- Wire it in the router; exclude `/health*`, `/readyz`, `/metrics`.

## Acceptance criteria
- [ ] Exceeding the configured rate on `POST /auth/challenge` returns `429` with `Retry-After`; under the limit is unaffected.
- [ ] Limits are enforced per-user on authenticated routes and per-IP on anonymous ones.
- [ ] Limits are configurable via env; defaults documented in `.env.example`.
- [ ] Health/readiness/metrics endpoints are never rate-limited.
- [ ] Table-driven tests cover: under-limit pass, over-limit `429`, window reset, and per-key isolation.
- [ ] No regression to existing middleware ordering (auth still runs, CORS intact).

## Files to check
- `apps/api/internal/middleware/` (existing `auth.go` for the pattern)
- `apps/api/internal/config/config.go`, root `.env.example`
- API server/router setup in `apps/api/cmd/` and `internal/api/response.go`
- Redis client wiring under `apps/api/internal/`

## Out of scope
WAF/edge rate limiting, DDoS mitigation at infra level.

---
### Contribution requirements (read before you start)
[...]

PR #1284, opened 2026-08-30
security(api): money-path integrity and auth handshake hardening

### LINKED ISSUE #1104: security(api): rate limit and lock out repeated auth challenge failures
## Context

The challenge and verify endpoints can be hammered to flood the challenge store and to brute-force signatures. Open issue #782 covers general rate limiting; this is the auth-specific hardening on top.

## Acceptance criteria

- [ ] Per-wallet and per-IP limits on challenge and verify.
- [ ] Progressive backoff and temporary lockout after repeated failures.
- [ ] Challenges are single-use and expire quickly.
- [ ] Lockouts are observable as a metric.
- [ ] Load test proves the challenge store cannot be flooded.

## Notes

Coordinate with #782 rather than duplicating it.

---

*Testnet launch backlog — E4 Auth and access.*

### LINKED ISSUE #1145: security(api): verify the amount and destination of a confirmed transaction, not just that it succeeded
## Context

`lookupHorizonTransaction` (apps/api/internal/service/transaction_service.go:179-223) decodes only `{Successful, CreatedAt, ResultXdr}` and returns `StatusCompleted` on `payload.Successful` alone. `applyConfirmedBalance` then credits `model.Amount` — a value the client supplied verbatim in the request. Nothing compares the credited amount against the actual operation, the destination, or the asset.

A user can post ANY real successful testnet transaction hash with `type: "deposit"` and an arbitrary amount, and the poller will confirm it against Horizon and credit the full amount to their vault. This is unlimited free balance through an authenticated, ownership-checked endpoint.

## Acceptance criteria

- [ ] The Horizon lookup decodes the operations of the transaction, not only its success flag.
- [ ] Confirmation requires that the transaction actually moved the claimed asset, in the claimed amount, to the vault's contract address.
- [ ] The credited amount is taken from the on-chain operation, never from the request body.
- [ ] A hash whose operations do not match the claim is rejected and the transaction is marked failed with a typed reason.
- [ ] Tests: unrelated-but-successful hash rejected; correct hash with inflated amount rejected; wrong asset rejected; wrong destination rejected; matching transaction credits exactly the on-chain amount.

## Notes

Found during a follow-up API audit. Verified at transaction_service.go:210-223. This is the single most severe finding in the backlog — more direct than #1075, because it needs no forged amount, only a real hash.

---

### LINKED ISSUE #1146: fix(api): convert stroop event amounts to asset units before writing vault balances
## Context

`applyIndexedEvent` (apps/api/internal/stellar/indexer.go:223-256) writes the raw extracted event amount straight into `total_deposited` and `current_balance`. `extractEventAmount` returns the value verbatim and its own comment notes Soroban amounts are stroops. Meanwhile the API path stores plain asset units and multiplies by 1e7 only for the chain call (`vault_service.go:349`).

So the same two columns are written in two different units. Every indexed on-chain deposit credits 10,000,000x the real amount.

## Acceptance criteria

- [ ] The indexer converts stroops to asset units using the asset's decimals before any balance write.
- [ ] A single documented helper performs the conversion; no call site does it inline.
- [ ] A test asserts a 1 USDC on-chain deposit results in a balance of 1, not 10000000.
- [ ] A round-trip test proves the API path and the indexer path produce identical balances for the same deposit.

## Notes

Found during a follow-up API audit. Verified at indexer.go:228-235 against vault_service.go:349.

---

### LINKED ISSUE #1147: fix(api): stop the indexer and the API write path double-counting the same deposit
## Context

Two independent writers increment the same balance columns with no shared dedupe key. `RecordDeposit` (vault_repository.go:305-312) increments on vault id and is passed `TxHash: ""` by the handler, so it has no dedupe key at all. The indexer (`stellar/indexer.go:228-235`) increments on contract address, deduped only by `processed_events.event_id`.

A deposit made through the API increments the balance immediately, and the indexer increments it again when the resulting on-chain event arrives. Every deposit is counted twice.

## Acceptance criteria

- [ ] A single authoritative writer for on-chain-derived balances, or a shared idempotency key that both paths honour.
- [ ] A deposit made through the API and then observed by the indexer results in exactly one credit.
- [ ] Integration test drives an API deposit followed by the corresponding indexed event and asserts the balance moved once.
- [ ] The chosen ownership model is documented in the indexer package.

## Notes

Found during a follow-up API audit. Depends on the on-chain verification in #1075; resolve them together.

---

Chidimj:
Testnet launch backlog: the three money-path integrity findings plus the auth-specific hardening. Four branches, one PR — #1145 and #1147 touch the same confirmation path and #1146 and #1147 the same indexer branches, so they are easier to review and land together than separately.

Closes #1104
Closes #1145
Closes #1146
Closes #1147

## What each commit does

**#1145 — verify amount and destination, not just success.** The Horizon lookup decoded only the success flag and credited `model.Amount` from the request body, so any real successful hash posted as a deposit credited an arbitrary amount. Confirmation now decodes the transaction's operations and requires a value-transferring operation that paid the vault's own contract address, in the vault's currency, for the claimed amount. The credited figure comes from the on-chain operation. Mismatches are marked failed with a typed reason; transient failures stay pending rather than falling through to a credit.

**#1146 — stroop conversion.** The indexer wrote raw stroop event amounts into balance columns that store asset units, inflating every indexed deposit by 1e7. Conversion now happens at the extraction boundary through one shared helper, which the existing chain-event verifier also routes through.

**#1147 — double counting.** The indexer and the API write path incremented the same columns with no shared dedupe key, so an API deposit was credited again when its event arrived. Both writers now claim `vault_transactions.transaction_hash` before moving a balance. The ownership model and why single-writer was rejected are documented in `internal/stellar/balance_ownership.go`.

**#1104 — auth lockout.** Adds per-wallet limiting and a progressive, capped failure lockout (per wallet and per IP) on top of the per-IP limiter from #782, with lockout metrics. Challenges were already single-use and short-lived, so that criterion needed no change.

## Verification

`go build ./...` and `go vet ./...` clean. Unit suites green for middleware, metrics, config, handler, service, stellar. The #1147 integration tests run against real PostgreSQL with the full migration chain.

Each fix was mutation-tested — reverted, and the new tests confirmed to fail. #1147's headline test fails with `total_deposited = 200, want 100`; 8 of 10 #1145 tests fail against the pre-fix behaviour.

## Notes for review

- `#1145` scopes the claim check to deposits: withdrawals move value out of the vault rather than into its contract address, and settlements move no vault balance.
- `#1147` deliberately still applies events carrying no transaction hash — those can only be direct on-chain movements nothing else will credit, and blocking them would silently drop real deposits.
- `#1104` keeps wallet addresses and IPs out of metric label values on purpose; they are attacker-controlled and unbounded.
- `apps/api/internal/stellar/submission_pipeline.go` is unformatted on `dev` already and was left alone.

## Summary by CodeRabbit

- **New Features**
  - Added progressive authentication lockouts after repeated failures, with separate wallet and IP protection.
  - Added authentication lockout monitoring and retry guidance.

- **Bug Fixes**
  - Deposits and withdrawals are now protected against duplicate balance credits or debits.
  - Deposit confirmations verify the on-chain destination, asset, and amount before crediting funds.
  - Improved handling of precise Stellar asset amounts to prevent conversion errors.
  - Invalid or mismatched confirmations no longer incorrectly update balances.

### DIFF
--- a/apps/api/.env.example
+++ b/apps/api/.env.example
@@ -132,6 +132,18 @@ RATELIMIT_WALLET_WINDOW=1m
 RATELIMIT_AUTH_LIMIT=10
 RATELIMIT_AUTH_WINDOW=1m
 
+# Auth-failure lockout (nester#1104). Distinct from RATELIMIT_AUTH_* above,
+# which bounds how FAST the auth endpoints may be called; these bound how many
+# times a caller may FAIL, tracked per wallet address and per client IP
+# independently. The backoff doubles per failure past the threshold, starting
+# at AUTH_LOCKOUT_BASE and capped at AUTH_LOCKOUT_MAX. The cap is deliberate:
+# without it anyone could deny service to a wallet they do not control by
+# failing verify against it forever. A successful login clears the history.
+AUTH_FAILURE_THRESHOLD=5
+AUTH_FAILURE_WINDOW=15m
+AUTH_LOCKOUT_BASE=30s
+AUTH_LOCKOUT_MAX=15m
+
 # Strict per-user rate limit for settlement creation (POST /api/v1/settlements).
 # Keyed by user ID so a single account cannot spam settlements across IPs.
 RATELIMIT_SETTLEMENT_LIMIT=5
--- a/apps/api/cmd/api/main.go
+++ b/apps/api/cmd/api/main.go
@@ -1379,6 +1379,28 @@ func run() error {
 		},
 		"authentication rate limit exceeded",
 	)
+	// authGuard hardens the same handshake beyond request rate (nester#1104):
+	// a per-wallet limit (the limiter above keys only on IP, so a distributed
+	// client could flood the challenge store for one wallet without tripping
+	// it), and a progressive lockout on repeated FAILURES tracked per wallet
+	// and per IP. Slowing down does not evade the lockout, because the backoff
+	// escalates with the failure count rather than resetting with time.
+	authLockoutCfg := middleware.AuthLockoutConfig{
+		Threshold: cfg.RateLimit().AuthFailureThreshold(),
+		Window:    cfg.RateLimit().AuthFailureWindow(),
+		Base:      cfg.RateLimit().AuthLockoutBase(),
+		Max:       cfg.RateLimit().AuthLockoutMax(),
+	}
+	authGuard := middleware.NewAuthGuard(
+		middleware.NewLimiter(redisClient, "authwallet", cfg.RateLimit().AuthLimit(), cfg.RateLimit().AuthWindow()),
+		middleware.NewAuthLockout(redisClient, "wallet", authLockoutCfg),
+		middleware.NewAuthLockout(redisClient, "ip", authLockoutCfg),
+		appMetrics,
+		[]middleware.AuthGuardStage{
+			{Stage: metrics.AuthStageChallenge, Route: middleware.RouteMatch{Method: http.MethodPost, Path: "/api/v1/auth/challenge"}},
+			{Stage: metrics.AuthStageVerify, Route: middleware.RouteMatch{Method: http.MethodPost, Path: "/api/v1/auth/verify"}},
+		},
+	).Middleware()
 	// settlementLimiter applies a strict per-user limit to settlement creation to
 	// prevent settlement spam. Placed after authentication so it keys by user ID.
 	settlementLimiter := middleware.SensitiveUserRouteLimiter(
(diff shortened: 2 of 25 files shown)


0xDeon:
@Chidimj FIx failing CI

Chidimj:
> @Chidimj FIx failing CI
Okay, just a minute

0xDeon (approved):
LGTM!!!

